PR #8, opened 2015-02-24
Added User Refresh Token support

### LINKED ISSUE #1: ADC Support for User Refresh Tokens
The initial implementation of Application Default Credentials is missing one of the 3 credential types: User Refresh Tokens. This is basically the end result of a 3LO flow, and is an important part of the tool integration, as the Cloud SDK writes out this form of credential in the well-known-file location.

The file format to support is this:

```javascript
{  
  "type": "authorized_user",     
  "client_id": "kflc91.apps.googleusercontent.com",     
  "client_secret": "1/olgReg3YaBQqxm6T",
  "refresh_token": "2/fFAGRNJru1FTz70BzhT3Zg"     
}
```

The "type" value is to differentiate it from a service account. The other 3 values are needed to get refresh tokens back. This is a common OAuth2 scenario, so there is likely to be an existing object that already supports getting access tokens from these values. This file could be in either the well-known file  location or the environment variable location.

For reference, the Java implementation of these features is here:
https://github.com/google/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
https://github.com/google/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/UserCredentials.java

This should be about 3 days of work.

murgatroid99:
This adds support for loading a User Refresh Token instead of JWT from a file. It also moves the implementation of `fromStream` into `googleauth.js` to avoid duplicating the implementation.

This should close #1.


lib/auth/googleauth.js
```diff
@@ -294,12 +309,25 @@ GoogleAuth.prototype.fromJSON = function(json, opt_callback) {
...
 GoogleAuth.prototype.fromStream = function(stream, opt_callback) {
   var that = this;
-  var jwtClient = new that.JWTClient();
-  jwtClient.fromStream(stream, function(err) {
-    if (err) {
+  if (!stream) {
+    process.nextTick(function() {
+      callback(
+        opt_callback,
+        new Error('Must pass in a stream containing the service account auth settings.'));
```
anthmgoogle: I may be misreading, but this function looks like it should support both user and service account files but the error message implies that it only supports service accounts.

lib/auth/refreshclient.js
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * User Refresh Token credentials.
+ *
+ * @param {string} clientId The authentication client ID.
+ * @param {string} clientSecret The authentication client secret.
+ * @param {string} refreshToken The authentication refresh token.
+ * @constructor
+ */
+function RefreshClient(clientId, clientSecret, refreshToken) {
```
anthmgoogle: This is a very implementation-oriented name. Perhaps UserRefreshClient would still be consistent with other types but add a hint about what you would use it for.

lib/auth/googleauth.js
```diff
@@ -294,12 +309,25 @@ GoogleAuth.prototype.fromJSON = function(json, opt_callback) {
...
+    });
+    return;
+  }
+  var s = '';
```
jasonall: I believe this stream handling code exists in three places now: googleauth, jwtclient, and refreshclient. Probably this code should be pushed down into authclient (or oauthclient) and jwtclient and refreshclient should call down to the base class. Also googleauth should delegate the fromStream call down to either jwtclient or refreshclient, similar to what happens in fromJSON.

test/test.googleauth.js
```diff
@@ -224,6 +278,28 @@ describe('googleAuth', function() {
         done();
       });
     });
+
+    it('should read another stream and create a RefreshClient', function(done) {
```
jasonall: Add a similar test which reads a stream and produces a jwtclient.

test/test.refresh.js
```diff
@@ -0,0 +1,155 @@
...
+  };
+}
+
+describe('Refresh Token auth client', function() {
+
+});
+
+describe('.fromJson', function () {
+
+  it('should error on null json', function () {
```
jasonall: For this and all other negative tests, use the flavor which provides a 'done' handler, and call done from within the callback. Currently, if the callback is never executed, the tests will pass.

test/test.refresh.js
```diff
@@ -0,0 +1,155 @@
...
+    delete json.refresh_token;
+
+    var auth = new googleAuth();
+    var refresh = new auth.RefreshClient();
+    refresh.fromJSON(json, function (err) {
+      assert.equal(true, err instanceof Error);
+    });
+  });
+
+  it('should create RefreshClient with clientId_', function() {
```
jasonall: For this and other positive tests, use a 'done' handler and call done from within the callback. Otherwise the test will pass if the callback is never executed.

test/test.googleauth.js
murgatroid99: See line 204.

lib/auth/googleauth.js
murgatroid99: It doesn't make sense to put it in `authclient` or `oauthclient` because `OAuth2Client` does not have anything reasonable to do with a stream, and neither does `computeclient` (another subclass of `Auth2Client`.

murgatroid99: Also, I can't have `fromStream` delegate to `jwtclient` or `refreshclient` directly because it needs to read the JSON content to determine which to delegate to, and I can't do that before reading the stream and parsing it into JSON, at which point I might as well just delegate in `fromJSON` anyway.

murgatroid99:
I am closing this pull request and moving to #14 because when this repository became public, it disassociated with my fork.

PR #16, opened 2015-03-09
Merging in changes from the google-api-nodejs-library repo

jasonall:
There have been some recent changes to the utils.js file in the google-api-nodejs-client repo. However, that file has been moved to this repo, and will soon be deleted from the original repo.

I am merging the exact same changes into this repo.

### DIFF
--- a/test/test.utils.js
+++ b/test/test.utils.js
@@ -1,0 +1,45 @@
+/**
+ * Copyright 2014 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


test/test.utils.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+'use strict';
+
+var assert = require('assert');
+var utils = require('../lib/utils.js');
+
+describe('Utils', function() {
+
+  it('should not extend funtions on the prototype', function() {
```
tbetbetbe: s/funtions/functions/

jasonall:
Actually that typo is from the previous commit to the google-api-nodejs-client repo, which I merged here. I will fix the typo here. Thanks!

PR #19, opened 2015-03-23
Add url parameter as request method.

fossamagna:
First argument of request method is First argument of request method
must be url. And number of argument must be three.


tbetbetbe:
@fossamagna - thank you for this contribution.
This change breaks a number of tests and the Travis build no longer succeeds - we can't accept the PR as is.
Can you add the necessary updates to the tests to the PR as well please ?

fossamagna:
@tbetbetbe I updated to the test code to the PR. Please check.

PR #22, opened 2015-03-30
Added usage example to readme

jasonall:
Added an example similar to that found in the google-auth-library-ruby readme, also added the same Application Default Credentials section found there. A couple of other small cleanups.


anthmgoogle:
LGTM.

PR #26, opened 2015-04-20
Create a meaningful error from response.error.errors array

gfxmonk:
Certain API responses include an array of errors, rather than just one - e.g. https://cloud.google.com/compute/docs/reference/latest/regionOperations . Currently this throws a fairly useless error with an `undefined` message and an error code of `200`.

When `error.errors` is an array, this patch includes an error text with newline-joined messages from each individual error, the `code` from the first error, and an `errors` field with the underlying array of error objects.


tbetbetbe:
Thanks, this looks useful.  Could you please add unit test covering the added behaviour ?

PR #29, opened 2015-04-22
Nodejs auth: add jwt header creds

tbetbetbe:


README.md
```diff
@@ -45,14 +45,16 @@ google.auth.getApplicationDefault(function(err, authClient) {
...
-    
+
     // Fetch the access token
-    authClient.getAccessToken(function(err, token)) {
+    var optionalUri = null;  // optionally specify the URI being authorized
+    authClient.getRequestMetadata(optionalUri, function(err, tokens)) {
       if (err === null) {
-        // Use the access token
+        // Use the auth tokens
+        var access_token = tokens.access_token;
```
anthmgoogle: This is a little different from the Java and C++ implementations, which construct the bearer token and put it in the authorization header, so that the dictionary can be copied into a request directly. If we want to have something that returns an access token, perhaps it should still be called getAccessToken, and if we have getRequestMetadata it should just return a dictionary of headers.

lib/auth/jwtheader.js
```diff
@@ -0,0 +1,176 @@
...
+  this.email = email;
+  this.key = key;
+}
+
+/**
+ * Executes the given callback if it is not null.
+ *
+ * Is used to simplify invocation of optional callbacks.
+ */
+function maybeCall_(c, err, res) {
```
anthmgoogle: This seems like a general purpose helper. Perhaps make it internal or move to a utility class.

lib/auth/jwtheader.js
```diff
@@ -0,0 +1,176 @@
...
+ * JWTHeader service account credentials.
+ *
+ * Create a new access token by using the credential to create a new JWT token
+ * that's recognized as the access token.
+ *
+ * @param {string=} email the service account email address.
+ * @param {string=} key the private key that will be used to sign the token.
+ * @constructor
+ */
+function JWTHeader(email, key) {
```
anthmgoogle: Should this be called JWTAccess instead? We discussed what to name this and rejected JWTHeader in favor of JWTAccess which describes that the JWT in the header grants direct access.

anthmgoogle:
Did a review pass. The implementation seems consistent with other lanagues. The concept name is different and the abstraction level of getRequestMetadata is very different. In particular, we wanted getRequestMetadata to support the IAM credential, which has different 2 header values and no access token.

anthmgoogle:
After reviewing the IAM changes, those are more what I was expecting, so I'm suggesting a small change to getRequestMetadata to return the authorization header instead of the access token.

lib/auth/jwtheader.js
tbetbetbe: Fixed

lib/auth/jwtheader.js
tbetbetbe: I removed this and replaced it's behaviour with the idiom described in #33.

README.md
tbetbetbe: Fixed

PR #31, opened 2015-04-23
Adds an implementation of IAM credentials

### LINKED ISSUE #25: Add a new credential type, IAMCredential
It is constructed with and holds two fields
  - iam-token
  - iam-authority-selector
IAMCredential applies these values to requests as a pair HTTP headers (or an equivalents) keys
- "x-goog-iam-authorization-token"
- "x-goog-iam-authority-selector"
respectively

N.B, there is no requirement that an IAMCredential be returned by Application Default Credentials.

tbetbetbe:
Fixes #25

### DIFF
--- a/lib/auth/iam.js
+++ b/lib/auth/iam.js
@@ -1,0 +1,59 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


lib/auth/iam.js
```diff
@@ -0,0 +1,59 @@
...
+'use strict';
+
+/**
+ * IAM credentials.
+ *
+ * @param {string=} selector the iam authority selector
+ * @param {string=} token the token
+ * @constructor
+ */
+function IAM(selector, token) {
```
anthmgoogle: This name seems a little non specific since IAM is a technology set. Perhaps IAMCredentials or IAMAuth?

anthmgoogle:
LGTM although it may be worth discussing the name.

lib/auth/iam.js
tbetbetbe: Done

PR #38, opened 2015-05-18
Fixed error code not being parsed correctly.

### LINKED ISSUE #37: Error code no longer passed through in response
[This is cross-posted from https://github.com/google/google-api-nodejs-client/issues/424. I'm not sure what project it should be in.]

I'm using v2.0.3 of this library in the following way:

```
      client.users.drafts.update({
        userId: 'me',
        id: gmailDraftId,
        resource: {
          message: {
            raw: require('base64-url').encode(job.data.emailDraft)
          }
        }
      }, function(err, res, req) {
          // Detect 404 errors
         if (err.code === 404) console.log('does not exist');
      }));
```

In v1.x of this library, `err.code` was populated and correctly set to `404` when the draft didn't exist. However, since I upgraded to 2.x, `err.code` is now undefined. However, the response from the api hasn't changed; `req.body` is still: ` { error: { errors: [Object], code: 404, message: 'Not Found' } } }`.

Did something break in 2.x causing the error code to not get passed through from the response?

fiznool:
Fixes #37

### DIFF
--- a/lib/transporters.js
+++ b/lib/transporters.js
@@ -82,7 +82,7 @@ DefaultTransporter.prototype.wrapCallback_ = function(opt_callback) {
         err = new Error(body.error.errors.map(
                          function(err) { return err.message; }
                        ).join('\n'));
-        err.code = body.error.errors[0].code;
+        err.code = body.error.code;
         err.errors = body.error.errors;
 
       } else {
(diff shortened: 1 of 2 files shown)


fiznool:
CLA signed

PR #61, opened 2015-09-04
Use metadata.google.internal

qix:
Using the full hostname rather than requiring the resolv.conf domain configuration is required for our setup.


qix:
I signed it!

tbetbetbe:
#69 replaces this

PR #63, opened 2015-10-22
Accept the new issuer value for Google ID tokens.

mcduan:
Google used to use "accounts.google.com" as the ID token issuer and now is using "https://accounts.google.com" in its OpenId Connect spec compliant endpoints. Per https://developers.google.com/identity/protocols/OpenIDConnect?hl=en#sendauthrequest, both "accounts.google.com" and "https://accounts.google.com" should be accepted.


ryanseys:
LGTM. @tbetbetbe what do you think?

tbetbetbe:
LGTM

PR #69, opened 2015-11-12
Use full compute metadata DNS name

jonparrott:
Use the `metadata.google.internal` instead of just `metadata`. This enables Compute Engine auth to work on Kubernetes, Container Engine,  and Container-Optimized VMs and setups with custom DNS servers. Otherwise, the DNS lookup for `metadata` will fail.


jonparrott:
I just noticed that #61 was created in Sept, however, it seems to be out of date now and the tests are failing.

jonparrott:
Thanks. I'll keep an eye on the next release.

PR #76, opened 2016-03-24
Don't depend on a library to get the noop function

murgatroid99:
We don't need to depend on a library just to get `function() {}`.

### DIFF
--- a/lib/auth/jwtaccess.js
+++ b/lib/auth/jwtaccess.js
@@ -17,7 +17,9 @@
 'use strict';
 
 var jws = require('jws');
-var noop = require('lodash.noop');
+
+function noop() {
+}
 
 /**
  * JWTAccess service account credentials.
--- a/lib/auth/jwtclient.js
+++ b/lib/auth/jwtclient.js
@@ -19,9 +19,10 @@
 var Auth2Client = require('./oauth2client.js');
 var gToken = require('gtoken');
 var JWTAccess = require('./jwtaccess.js');
-var noop = require('lodash.noop');
 var util = require('util');
 
+function noop() {
+}
 
 /**
  * JWT service account credentials.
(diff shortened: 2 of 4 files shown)


tbetbetbe:
Here's a different view.

Part of the way the npm ecosystem works is that even single functions are available as modules.
It is normal for node packages to make use of them; there are many more npm than _.empty.

Is this PR addressing any actual issue ?  If it's not, then I'd prefer to close it

murgatroid99:
Taking a dependency does not have zero cost. We saw that this week with the left-pad fiasco. And the `lodash.noop` module *literally* does nothing.

tbetbetbe:
- I don't think  _.noop in any danger of being removed as leftpad was
- I'm curious to understand what the costs are, could you please outline them ?  Like I said, if the costs amount to an issue that needs to be fixed, and re-open the PR.

nicolasnoble:
The costs are tied to actively using an internet connection to download a package that literally does nothing. That in my book validates the usefulness of such a PR.

tbetbetbe:
I'm not really sure that counts as an issue ;)

We are not going to agree on this because neither does the internet:
- For this PR: http://www.haneycodes.net/npm-left-pad-have-we-forgotten-how-to-program
- the counterargument: https://github.com/sindresorhus/ama/issues/10#issuecomment-117766328

There are 22 real open issues in this project, I'd prefer to spend time addressing them.

murgatroid99:
There's a difference between importing a single-line function that does something slightly non-obvious, and importing a function that literally does nothing. Nobody is going to implement noop incorrectly. Nobody needs to take time to think about how to implement it. There are no language gotchas or environment-specific concerns. Using `lodash.noop` just means that a hundred thousand people a month download it as a transitive dependency for absolutely no benefit.

Issue #67, opened 2015-11-12
automatic token refresh is not happening in getrequestmetadata


ssuda:
incase of 401 or 403, automatic token refresh used happen before.
But it is not happening now.  Is it removed or bug?

L422Y:
I still seem to be having this issue, my application will work fine and then eventually only get `{ error: 'unauthorized_client' }` responses from the server until I manually go through getting new keys.

JustinBeckwith:
Greetings!  We have a fair bit of code and tests in place to make sure this works.  We've also made a few recent changes (like #242) that should make it even better.  If you're still having issues here, providing a code snippet to reproduce the issue would be *super* helpful. 

Thanks!

PR #77, opened 2016-03-25
Make Oauth2Client#request retry on auth errors as documented

### LINKED ISSUE #42: Buggy comment in oauth2client.js regarding replay on 401/403
The docstring on `OAuth2Client.prototype.request` says:

```
 * Provides a request implementation with OAuth 2.0 flow.
 * If credentials have a refresh_token, in cases of HTTP
 * 401 and 403 responses, it automatically asks for a new
 * access token and replays the unsuccessful request.
```

But I don't see any logic that would retry on 401/403, nor does that seem plausible given that it wouldn't work with stream bodies.  AFAIK, the token is only refreshed *before* making the request.  (Sorry if I missed something.)

murgatroid99:
Fixes #42 and (I think) #67.

### DIFF
--- a/test/test.jwt.js
+++ b/test/test.jwt.js
@@ -268,12 +268,12 @@ describe('JWT auth client', function() {
     });
 
     it('should refresh token if the server returns 403', function(done) {
-      var scope = nock('http://example.com')
+      nock('http://example.com')
           .log(console.log)
           .get('/access')
           .reply(403);
 
-      var auth = new googleAuth();
+      var auth = new GoogleAuth();
       var jwt = new auth.JWT(
           'foo@serviceaccount.com',
           '/path/to/key.pem',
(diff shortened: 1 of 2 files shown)


tbetbetbe:
Please add unit tests that verify the fixes.

murgatroid99:
I've added a test for that

murgatroid99:
Ping

test/test.jwt.js
```diff
@@ -267,6 +267,39 @@ describe('JWT auth client', function() {
       });
     });
 
+    it('should refresh token if the server returns 403', function(done) {
```
tbetbetbe: optional: You can test both 401 and 403 by calling the spec in a loop

tbetbetbe:
LGTM; please merge once you've merged/rebased and the travis CI has passed

PR #79, opened 2016-04-01
Capitalize constructor names everywhere to satisfy linter

### LINKED ISSUE #78: Fix lint errors
All of the travis builds are failing with dozens of lint errors, which makes it hard to tell if the tests are actually passing. This should be fixed.

murgatroid99:
This fixes #78. Note that all of these changes except the one in `googleauth.js` are in test files, and none of these changes has any effect on the API surface


tbetbetbe:
LGTM

PR #80, opened 2016-04-01
Check that pem exists before verifying it.

### LINKED ISSUE #46: TypeError: Not a Buffer
When using the `verifyIdToken` to verify a really old token (So far I only encoutered with these kind of tokens), instead of getting an expired error the app throws this error.

I've digged a bit and found that the main issues resides within the line 495 in `oauth2client.js`:

```
var pem = certs[envelope.kid];
```

That resolves to undefined sometimes. `envelope.kid` holds a value, but not one that matches the certs generated by the `getFederatedSignonCerts` function in `oauth2client.js`, thus causing the `pemVerifier.verify` throw.

murgatroid99:
Should fix #46.

### DIFF
--- a/lib/auth/oauth2client.js
+++ b/lib/auth/oauth2client.js
@@ -492,6 +492,10 @@ OAuth2Client.prototype.verifySignedJwtWithCerts =
       throw new Error('Can\'t parse token payload: ' + segments[1]);
     }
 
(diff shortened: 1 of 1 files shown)


murgatroid99:
ping

tbetbetbe:
LGTM

brunolazzaro:
Good to see this fixed! 👍

PR #84, opened 2016-05-03
Fix User-Agent header

### LINKED ISSUE #83: User Agent header grows unboundedly
Every call made using this module appears to add another `"google-api-nodejs-client/0.9.7"` to the `User-Agent` header. This causes the header to grow in size unbounded.

Example: http://pastebin.com/dHFGRkB8

Culprit: https://github.com/google/google-auth-library-nodejs/blob/master/lib/transporters.js#L42

matthewloring:
Prevent User-Agent header from growing with each request.

Fixes #83

### DIFF
--- a/lib/transporters.js
+++ b/lib/transporters.js
@@ -39,8 +39,11 @@ DefaultTransporter.prototype.USER_AGENT =
 DefaultTransporter.prototype.configure = function(opts) {
   // set transporter user agent
   opts.headers = opts.headers || {};
-  opts.headers['User-Agent'] = opts.headers['User-Agent'] ?
-    opts.headers['User-Agent'] + ' ' + this.USER_AGENT : this.USER_AGENT;
+  if (!opts.headers['User-Agent']) {
+    opts.headers['User-Agent'] = this.USER_AGENT;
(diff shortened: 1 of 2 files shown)


lib/transporters.js
```diff
@@ -39,8 +39,11 @@ DefaultTransporter.prototype.USER_AGENT =
 DefaultTransporter.prototype.configure = function(opts) {
   // set transporter user agent
   opts.headers = opts.headers || {};
-  opts.headers['User-Agent'] = opts.headers['User-Agent'] ?
-    opts.headers['User-Agent'] + ' ' + this.USER_AGENT : this.USER_AGENT;
+  if (!opts.headers['User-Agent']) {
+    opts.headers['User-Agent'] = this.USER_AGENT;
+  } else if (!opts.headers['User-Agent'].endsWith(this.USER_AGENT)) {
```
stephenplusplus: I think we should find another way to check this, since this is only compatible with Node 4 and up.

matthewloring: Done.

tbetbetbe:
LGTM

PR #91, opened 2016-08-04
Handle symlinks to default application credentials

JonathanPorta:
When using Kubernetes 1.3 secrets with Node.js the secret volume can be
a symbolic link to the credentials file. Previously, this would throw an
error due to the assertion of lstat.isFile(). This patch adds real path
resolution prior to asserting that filePath points to valid file.


JonathanPorta:
I signed it!

tbetbetbe:
LGTM - thanks for the contribution

PR #96, opened 2016-09-29
Prepare for 0.9.9 release

bjwatson:
@murgatroid99 @jayantkolhe I think this is what we need to be ready to roll as soon as we have push access to npmjs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.9.9 (10/14/2016)
+
+### Changes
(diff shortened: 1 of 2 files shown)


bjwatson:
@murgatroid99 FYI, I think this contains the release notes you need, although there might be more since I did this.

murgatroid99:
Can you update this to account for #98.

bjwatson:
@murgatroid99 Done.

PR #98, opened 2016-10-01
Add support for detecting project ID. Fixes #97

### LINKED ISSUE #97: Support detection of project ID
RE: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1653

It would be nice if the authClient returned from this module also exported its best guess at the project_id. In the case of service account JSON files, they have the `project_id` set, so exporting that should be simple. For ADC, it would involve spawning a command:

```sh
$ gcloud -q config list core/project --format=json
{
  "core": {
    "project": "grape-spaceship-123"
  }
}
```

cc @tmatsuo @jmdobry

jmdobry:
Fixes google/google-auth-library-nodejs#97

- [x] - Add support for detecting default project ID
  - [x] - Detect from environment variables
  - [x] - Detect from `project_id` field in GOOGLE_APPLICATION_CREDENTIALS key file
  - [x] - Detect from Google Cloud SDK default config `core/project` setting
  - [x] - Detect Compute Engine project ID
- [x] - Write tests

### Before

```js
auth.getApplicationDefault(err, authClient) {
  // ... 
});
```

### After

```js
auth.getApplicationDefault(err, authClient, projectId) {
  // ... 
});
```

### DIFF
--- a/lib/auth/googleauth.js
+++ b/lib/auth/googleauth.js
@@ -18,6 +18,7 @@
 
 var JWTClient = require('./jwtclient.js');
 var ComputeClient = require('./computeclient.js');
+var exec = require('child_process').exec;
 var fs = require('fs');
 var os = require('os');
 var path = require('path');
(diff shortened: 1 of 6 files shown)


stephenplusplus:
What do you think about putting the project ID as a property on the authClient / a getter? Since users of the library are likely caching the authClient, it should be easier for them to just grab the projectId off of that object as opposed to additionally caching the projectId returned when they create the client.

jmdobry:
They are related but still separate concerns. One is a set of credentials (a json  key file), the other an identifier for a project. The `auth` object itself actually will have both of them cached. Today you can already access the cached credentials. My PR caches the project ID too. e.g.

```js
// Get just the project ID
auth.getDefaultProjectId(function (err, projectId) {
  // The project ID has now been cached
  console.log(auth._cachedProjectId === projectId); // true
});
```

```js
// Retrieve credentials and the project id
auth.getApplicationDefault(function (err, credentials, projectId) {
  // Both are cached
  console.log(auth._cachedCredential === credentials); // true
  console.log(auth._cachedProjectId === projectId); // true
});
```

This is how python is implementing it:

```python
import google.auth

credentials, project_id = google.auth.default()
```

stephenplusplus:
I think the functionality of auto-detection of a project ID is quite helpful, and I think users will appreciate it, too. My thought is we should make it as easy as possible for them to get by putting it on the authClient instance so they don't have to "catch it before it's gone" (in the callback of the auth client fetcher).

But either way, `google-auto-auth` wraps this library, so I'll put it on the `authClient` that library returns, which should help us in our client libs.

jmdobry:
> catch it before it's gone

It won't be gone, it's cached on the `auth` object, just like the credentials. Once that callback is called they can access the cached project id on the `auth` instance they used to call `getApplicationDefault`, e.g. `auth._cachedProjectId`. Also, if the `credentials`  object is actually a user's Service Account, it will already have a `project_id` field. I don't see anywhere in `google-auth-library-nodejs` where it actually modifies the default application credentials. It just loads them and caches them. I'm not sure one would want to attach the default project id to the loaded credentials, hence I cached them separate from each other.

> But either way, `google-auto-auth` wraps this library, so I'll put it on the authClient that library returns, which should help us in our client libs.

👍, though personally I would have `google-auto-auth` do as I've done here and cache the project id on the google-auto-auth `Auth` instance, not the `authClient` value (which is actually the parsed key file). e.g. (from [`google-auto-auth`](https://github.com/stephenplusplus/google-auto-auth/blob/master/index.js#L56)):

```js
function addScope(err, authClient, projectId) {
  if (err) {
    callback(err);
    return;
  }

  if (authClient.createScopedRequired && authClient.createScopedRequired()) {
    // ...
  }

  self.authClient = authClient;
  self.projectId = projectId;
  callback(null, authClient, projectId);
}
```

Then `google-cloud-node` or `gax-nodejs` can cache the credentials and projectId separately as they see fit. The two values are used for separate things. One for making an authorization header to pass with requests, the second to be a path parameter in requests (especially in the case of `gax-nodejs`.)

stephenplusplus:
> it's cached on the auth object, just like the credentials.

I guess all I'm really saying is it shouldn't be "private". In JS, the leading `_` tells developers not to rely on this value. And it tells contributors to this library that they can change the purpose of that property or remove it without being concerned it would break users' code.

>```js
  self.authClient = authClient;
  self.projectId = projectId;
```

Yep, that's exactly how I would do it :)

jmdobry:
So perhaps we should also add synchronous getters to google-auth-library-nodejs for the credentials and projectId

stephenplusplus:
Can we also read the project_id from a service account keyfile?

jmdobry:
Yep: https://github.com/google/google-auth-library-nodejs/pull/98/files#diff-74a29c6f6e51bc579dce9754d2d4eeb0R185

stephenplusplus:
Is that in the case I don't have `gcloud` SDK installed at all, and I just provide a JSON key file?

jmdobry:
Yeah, newer Service Account key files have a `project_id` field. This PR looks for a projectId in following places (in order of precedence):

* `GCLOUD_PROJECT` or `GOOGLE_CLOUD_PROJECT` env var
* `project_id` field of Service Account key file pointed to by `GOOGLE_APPLICATION_CREDENTIALS`
* `core/project` field of default config file of Google Cloud SDK
* AppEngine Identity (TODO still)
* GCE metadata server (TODO still)

stephenplusplus:
Does this library accept a key file path provided directly as a string, though? [This](https://github.com/stephenplusplus/google-auto-auth/blob/5115e7cd8331f79d498ab307c12161985d4a370e/index.js#L45) is how I do it in google-auto-auth, but I'm not sure if I'm doing it in a way that this library intended to support.

jmdobry:
Passing `keyFile` works as a consequence of `google-auth-library-nodejs` using `gtoken` under the hood. `google-auth-library-nodejs` itself does not actually do anything with `keyFile`.

`getApplicationDefault` and `getDefaultProjectId` don't look at `keyFile` at all.

stephenplusplus:
@jmdobry what's the process for getting this shipped?

stephenplusplus:
// @tbetbetbe @jasonall

jmdobry:
I'm not sure. Wishing I had write access.

PR #101, opened 2016-10-27
Fix getApplicationDefault to also return projectId.

jmdobry:


jmdobry:
@murgatroid99 PTAL

jmdobry:
ping? @murgatroid99 @bjwatson

bjwatson:
I'm not familiar enough with NodeJS to review this. @jmuk do you have a couple minutes to take a look?

lib/auth/googleauth.js
```diff
@@ -37,9 +37,9 @@ function GoogleAuth() {
 }
 
 // Executes the given callback if it is not null.
-function callback(c, err, res) {
+function callback(c, err, arg1, arg2) {
   if (c) {
-    c(err, res);
+    c(err, arg1, arg2);
```
jmuk: +1, although I'm not seeing usage of >2 args right now.

PlasmaPower: Yeah, it's just future proofing of course.

jmdobry: Are we sure we want to do this? Won't it wipeout the user's context if they bound their callback to something? e.g.

```js
function callback(c) {
  if (c) {
    c.apply(null, Array.prototype.slice.call(arguments, 1));
  }
}
```

PlasmaPower: No, a bind will still work even if apply is called on the result. Think of a this-only bind like this:

```js
function bind(fn, self) {
  return function() {
    return fn.call(self, arguments);
  }
}
```

Example of this working:


```js
var fn = (function () { return this.a }).bind({a: "hello"});
console.log(fn.call(null, []));
```

(outputs hello)

jmdobry: Did you mean to use `apply` in your examples?

jmdobry: Fixed

bjwatson:
No I can't. I'm not a member either.

@jmdobry can you make both @jmuk and I members of this repo? @jmuk should probably also be an owner.

jmdobry:
I can't, as I am not an admin, though @justinbeckwith and I have wanted to be for a while.

bjwatson:
@murgatroid99 @ofrobots, are you admins?

stephenplusplus:
Looks like I can merge. Should I try to release also? :)

PR #109, opened 2016-12-05
Fix oauth2 token refresh

ianmetcalf:
It looks like #77 attempted to get this working, but didn't cover the case i've been running into. 

Here is the documentation for the error response from google drive: https://developers.google.com/drive/v3/web/handle-errors#401_invalid_credentials

If the failed 401 (or 403) request includes the error in the response body, it gets parsed by the transporter and passed back as an error to the callback. This prevents the token from being refreshed based on the current conditional logic.

I've fixed this by adding an exception for these errors. I've also added tests for `OAuth2Client#request()` based on the ones for `JWT#request()` and made sure to cover this case with a failing test before adding the fix.

Let me know if you need anything else.

### DIFF
--- a/lib/auth/oauth2client.js
+++ b/lib/auth/oauth2client.js
@@ -344,10 +344,12 @@ OAuth2Client.prototype.request = function(opts, callback) {
 
   // Hook the callback routine to call the _postRequest method.
   var postRequestCb = function(err, body, resp) {
+    var statusCode = resp && resp.statusCode;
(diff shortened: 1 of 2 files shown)


0xmtn:
Hey @ianmetcalf,

Would this PR solve my issue on https://github.com/google/google-auth-library-nodejs/issues/111? Can you have a look? What do you think?

jmdobry:
Thanks @ianmetcalf

PR #110, opened 2016-12-08
Specify minimum request@2.79

SimenB:
This uses the uuid module, as node-uuid is deprecated

Support is dropped for node@0.10, but that's EOL anyway (and already in your semver range)

https://github.com/request/request/blob/master/CHANGELOG.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "gtoken": "^1.1.0",
     "lodash.noop": "^3.0.0",
     "jws": "^3.1.0",
-    "request": "^2.74.0",
+    "request": "^2.79.0",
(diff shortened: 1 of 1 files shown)


jmdobry:
I'm not sure if we're ready for this quite yet. `request` actually dropped support for 0.12.x [about a month ago](https://github.com/request/request/commit/4c4557faabf415a01ce2d8a7c3473311c126f69b). The most recent version of `request` that is compatible with Node v0.12 is `2.76.0`.

Some discussion here: GoogleCloudPlatform/google-cloud-node#1855

JustinBeckwith:
This is already at 2.81.0, so this isn't needed.

PR #112, opened 2016-12-29
Remove unused dependencies and upgrade remaining dependencies.

jmdobry:


package.json
```diff
@@ -33,24 +33,19 @@
...
-    "async": "^2.0.0",
-    "gtoken": "^1.1.0",
-    "lodash.noop": "^3.0.0",
-    "jws": "^3.1.0",
-    "request": "^2.74.0",
-    "string-template": "^1.0.0"
+    "gtoken": "^1.2.1",
+    "lodash.noop": "^3.0.1",
+    "jws": "^3.1.4",
+    "request": "^2.79.0"
```
stephenplusplus: request@2.79.0 has `engines >= 4` in package.json, so `npm install google-auth-library` will at best log a warning, and at worst fail to install due to either the user's own configuration or at runtime if request makes use of a non <4 feature. Do we care?

jmdobry: We've a couple options:

* I change the `request` version back to `^2.74.0` right away
* We leave it at `^2.79.0` for now, and trust our 0.10.0 tests to catch any future incompatibilities that might require us to downgrade or lock to a specific version
* We drop support for Node.js 0.10 and 0.12, making this upgrade to `^2.79.0` okay

stephenplusplus: We probably want someone else to decide this, so until we drop 0.12, my vote to play it safe and lock to 2.74.0.

jmdobry: I went ahead and reverted to `"request": "^2.74.0"` so that decisions about `request v2.79.0` and Node.js 0.10/0.12 don't have to block the 0.10.0 release.

PR #116, opened 2016-12-29
Prepare for 0.10.0 release.

jmdobry:


ofrobots:
#109 is listed as a bug fix in the changelog which may lead people to believe that it is semver-patch. I don't think it is semver-patch, correct?

jmdobry:
It is a bug fix, but I believe the nature of the change could be breaking for people who were coding around the problem before, so I bumped the minor version, which indicates a breaking change since we're pre-1.0.0. 

I rephrased that heading to `Breaking bug fixes`

CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+## 0.10.0 (12/29/2016)
+
+### Breaking bug fixes
+
+* Fix oauth2 token refresh (#109) ([@ianmetcalf][])
+
+### Backwards compatible changes
+
+* Remove unused dependencies and upgrade remaining dependencies. (#112) ([@jmdobry][])
```
ofrobots: This includes a bump to `request@2.79.0`. As per https://github.com/google/google-auth-library-nodejs/pull/110#issuecomment-266346971, I think we should understand/agree on the implications before releasing this change.

jmdobry: re: https://github.com/google/google-auth-library-nodejs/pull/112#discussion_r94173660

jmdobry: I went ahead and reverted to `"request": "^2.74.0"` so that decisions about `request v2.79.0` and Node.js 0.10/0.12 don't have to block this release.

ofrobots:
LGTM.

PR #121, opened 2017-01-10
Replace all instances of process.nextTick

cristiancavalli:
- Replace all instances of `process.nextTick` with
    `setImmediate`.


cristiancavalli:
@google/google-node-team @stephenplusplus @jmdobry for your collective considerations

ofrobots:
+1 !

I think this should be considered semver-major though in case anyone is dependent on the precise scheduling.

PR #126, opened 2017-02-27
Remove lodash.noop lib dependency

SkeLLLa:
Lodash noop lib is useless and it's better to use just:
```js
var noop = Function;
```

Because of following reasons:
* It's shorter
* It's faster
* It doesn't take space in node_modules
* It will not fail in case of deletion of lodash.noop
* It will not take space in package.json and yarn.lock
* Even lodash devs removed lodash.noop from their code.

### DIFF
--- a/lib/auth/jwtaccess.js
+++ b/lib/auth/jwtaccess.js
@@ -17,7 +17,7 @@
 'use strict';
 
 var jws = require('jws');
-var noop = require('lodash.noop');
+var noop = Function.prototype;
 
 /**
  * JWTAccess service account credentials.
--- a/lib/auth/jwtclient.js
+++ b/lib/auth/jwtclient.js
@@ -19,7 +19,7 @@
 var Auth2Client = require('./oauth2client.js');
 var gToken = require('gtoken');
 var JWTAccess = require('./jwtaccess.js');
-var noop = require('lodash.noop');
+var noop = Function.prototype;
 var isString = require('lodash.isstring');
 var util = require('util');
 
--- a/lib/auth/oauth2client.js
+++ b/lib/auth/oauth2client.js
@@ -18,7 +18,7 @@
 
 var AuthClient = require('./authclient.js');
 var LoginTicket = require('./loginticket.js');
-var noop = require('lodash.noop');
+var noop = Function.prototype;
 var PemVerifier = require('./../pemverifier.js');
 var querystring = require('querystring');
 var util = require('util');
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     "jws": "^3.1.4",
     "lodash.isstring": "^4.0.1",
     "lodash.merge": "^4.6.0",
-    "lodash.noop": "^3.0.1",
     "request": "^2.74.0"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -756,10 +756,6 @@ lodash.keys@^3.0.0:
     lodash.isarguments "^3.0.0"
     lodash.isarray "^3.0.0"
 
-lodash.noop@^3.0.1:
-  version "3.0.1"
-  resolved "https://registry.yarnpkg.com/lodash.noop/-/lodash.noop-3.0.1.tgz#38188f4d650a3a474258439b96ec45b32617133c"
-
 lodash@3.7.x:
   version "3.7.0"
   resolved "https://registry.yarnpkg.com/lodash/-/lodash-3.7.0.tgz#3678bd8ab995057c07ade836ed2ef087da811d45"


SkeLLLa:
I signed it!

murgatroid99:
Why `Function` instead of `function() {}`? Strictly speaking, `Function` isn't a noop; It's a function that returns a new function based on the input.

jmuk:
`noop` is not equivalent to `Function` itself. The latter is a constructor of a function. The equivalent to `noop` is `function() {}`

jmdobry:
From a `noop` perspective, the following should all work:

```js
var noop = function () {};
var noop = new Function();
var noop = Function();
var noop = Function;
var noop = Function.prototype;
```

cristiancavalli:
Why not create and use a named, static noOp in `utils.js` and just import it? In the current form you would be needlessly creating a new instance of the same function signature in each file and allocating 4 unnecessary function instances.

jmdobry:
With `var noop = Function.prototype;` we can avoid 4 extra function instances _and_ avoid modifying `utils.js`.

Ultimately though, I don't think this lib should be using noop at all, the callback should be mandatory (breaking change) where the lib is currently falling back to the noop.

SkeLLLa:
@cristiancavalli usually in my libs I prefer to set once `global.noop = Function`. In case of public lib it can do some harm to others, so it's probably not an option. But in any case creating 1 or 4 anonymous functions will not have significant affect on performance.

Also I think `var noop = Function.prototype` will be better, because it will not create additional instances, just like with `var noop = Function`

cristiancavalli:
@jmdobry What is the problem making noOp in utils? It's a util right?
@SkeLLLa Thanks, but the performance impact is not the point - its the idea that each top-level has a different localized assignment and implementation for what is ostensibly a globally-shared entity thereby creating an anti-pattern

SkeLLLa:
@cristiancavalli as I've wrote using global.noop is not an option for a public library. Also using `lodash.noop` as 3rd party module is also kind of antipattern. 

As for implementing in some module a simple noop function will require to add `require('utils.js')` in all files where you need `noop`. Even if not utils are required in all files that not a guarantee that it will be the same in future. So I would agree with @jmdobry that's `var noop = Function.prototype;` will be good enough.

cristiancavalli:
also @jmdobry Every single file in this PR **already** requires util.js except for `jwtaccess.js`. Why would we create a localized implementation in each file of a function which is a util and only needs to be implemented once.

cristiancavalli:
@SkeLLLa All the files (save 1) in this PR **already** requires util

cristiancavalli:
Also @SkeLLLa not sure how using a third-party module is an anti-pattern? Even for something like noOp. Where as this copy-paste solutions seems like a classic case of Abstraction inversion to me on the class level

SkeLLLa:
@cristiancavalli as for me it's definite antipattern to require module that doesn't contain any logic and have only one line of code that could be easilly replaced by one line in simple javascript code. It's like [adding two numbers in js with help of jquery](https://twitter.com/addyosmani/status/453181760868597760?lang=en)

cristiancavalli:
@SkeLLLa I understand your complaint - I've had it many times myself and I think subbing out lodash.noOp is better than keeping it in but I don't what you're citing is not an anti-pattern. An inefficiency? Totally. Violating any rules of abstraction/inheritance/relation? I can't see any, but I may not be seeing it

jmdobry:
I'm fine with either option in this case, but @cristiancavalli makes a good point that the files are already importing `utils.js` anyway, so I'm cool with moving the `noop` definition to `utils.js`.

SkeLLLa:
@jmdobry file https://github.com/google/google-auth-library-nodejs/blob/master/lib/auth/jwtaccess.js isn't importing utils

cristiancavalli:
@SkeLLLa Just by LOC, even with the change to jwtaccess.js for utils, you'd still come in under the current change footprint. (just sayin')

jmdobry:
Alright folks, just took a closer look at the code, and `utils.js` is not even used by the library, it's only used in one place in the tests. Turns out we confused `var util = require('util');` with `var utils = require('../utils');`. So, to move the `noop` to `utils.js` is now a bigger change than we thought.

cristiancavalli:
@jmdobry Nice catch. Given that, I'm fine with @SkeLLLa 's current solution.

cristiancavalli:
Thanks @SkeLLLa !!

SkeLLLa:
@cristiancavalli as for antipatterns - each module you require contains "risk". Even such little module as noop. Also noop module contains only one function that's already exist in javascript out-of-the-box. It makes no sense to write a separate module that will export something like
```js
module.exports = Array.prototype.map;
```
The same with noop. But it's my point of view.

PR #130, opened 2017-03-25
Fix project ID caching

alexbr:
There was a typo in googleauth.js which prevented project ID caching
from working. Cache lookups were against `_cachedProjectID` but the
project ID was set to `_cahchedprojectID` when found.


alexbr:
I signed it.

cristiancavalli:
@jmdobry I'm thinking it might be beneficial to have a follow-on commit somewhere for this class which mocks the caching lifecycle and then specifically checks to see that the property gets set on the instance, wdyt?

cristiancavalli:
@jmdobry should I merge this in?

PR #131, opened 2017-03-29
added input validation to idToken in verifyIdToken

StefanGussner:
Gives a meaningful error message when the ID Token is not a string. In the old version there was just an exception in Line 519: jwt.split is not a function.

### DIFF
--- a/lib/auth/oauth2client.js
+++ b/lib/auth/oauth2client.js
@@ -437,6 +437,10 @@ OAuth2Client.prototype.verifyIdToken = function(idToken, audience, callback) {
       'an ID Token and a callback method');
   }
 
+  if((typeof idToken) !=='string'){
+    throw new Error('The ID Token has to be a string');
+  }
+
   this.getFederatedSignonCerts(function(err, certs) {
     if (err) {
       callback(err, null);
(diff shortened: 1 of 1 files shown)


StefanGussner:
I signed it!

lib/auth/oauth2client.js
```diff
@@ -437,6 +437,10 @@ OAuth2Client.prototype.verifyIdToken = function(idToken, audience, callback) {
       'an ID Token and a callback method');
   }
 
+  if((typeof idToken) !=='string'){
```
cristiancavalli: @StefanGussner Thanks for this addition! A quick nit:

* Since `lodash.isstring` is already included as a dependency to this lib -- to better conform with existing code would you mind changing this to look something like:

```JS
var isString = require('lodash.isstring');
// ...
if (!isString(idToken)) {
  throw new Error('...');
}
```

cristiancavalli:
@jmdobry All clear to merge in?

StefanGussner:
yes :)

cristiancavalli:
@jmdobry ping -- can we merge this in

lukesneeringer:
@bantini Needs conversion to TypeScript.

JustinBeckwith:
Addressed by #184.  Thanks!

PR #139, opened 2017-05-03
Update `request` dependency to v2.81.0

kohlmannj:
This prevents the `google-auth-library`, by way of a dependency on an older version of `request`, from emitting a `node-uuid is deprecated` message during installation.


kohlmannj:
Note: I've reviewed the Travis build failure for [#244.3](https://travis-ci.org/google/google-auth-library-nodejs/jobs/228368310), and I'm not convinced I know enough about the project to determine why it failed.

ofrobots:
Sorry for the late response. This has already been addressed in another PR. Closing.

PR #141, opened 2017-05-05
Move code from lib/ to src/

ofrobots:
To be transpiled source shouldn't exist in `lib`.


JustinBeckwith:
In general, LGTM.  It is however, a breaking change.  How are we communicating those to users?

ofrobots:
In a follow-on PR, I can address the concern about the directory path changing being a breaking change for external users who are deep requiring files directly. I will modify the gulp configuration so that generated files do end up in `lib/`.

ofrobots:
Added a commit here that modifies the gulpfile to make sure the output js files are produced in lib/. This commit also enables the generation of .d.ts files.

ofrobots:
Avoiding a semver major is slightly more complicated than originally thought. Tests depend upon relative paths to the source files which don't look the same if the directory structure changes. e.g.

```
import GoogleAuth from '../src/auth/googleauth';
```

ofrobots:
Superceded by #144

PR #142, opened 2017-05-05
Add official tslint config

ofrobots:
'variable-name' had to be disabled as this module has some public
members that don't follow the camelCase naming convention.

'no-duplicate-variable', 'no-unreachable' and 'label-undefined' are no
longer necessary as the compiler obviates these now.

This is dependent on #141.

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -29,7 +29,8 @@
     "radix": true,
     "semicolon": [true, "always"],
     "switch-default": true,
-    "triple-equals": [true, "allow-null-check"]
+    "triple-equals": [true, "allow-null-check"],
 //    "variable-name": [true, "check-format", "ban-keywords"]
+    "variable-name": [true, "ban-keywords"]
   }
 }
--- a/tslint.json
+++ b/tslint.json
@@ -1,37 +1,35 @@
 {
-  "lintOptions": {
-    "typeCheck": true
-  },
-  "extends": "tslint:latest",
   "rules": {
-    "ban": [true,
-      ["describe", "only"],
-      ["it", "only"],
-      ["Object", "assign"]
-    ],
-
-    "callable-types": true,
-    "interface-name": [false],
-    "interface-over-type-literal": true,
-    "file-header": [true,
-      "Copyright "
-    ],
-    "max-classes-per-file": [false],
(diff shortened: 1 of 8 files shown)


cristiancavalli:
whoops sorry, wrong window focus :) ✋

tslint.json
```diff
@@ -1,37 +1,35 @@
...
+    "no-unused-expression": true,
+    "no-unused-variable": true,
+    "no-use-before-declare": true,
+    "no-var-keyword": true,
+    "quotemark": [true, "single"],
+    "radix": true,
+    "semicolon": [true, "always"],
+    "switch-default": true,
+    "triple-equals": [true, "allow-null-check"]
+//    "variable-name": [true, "check-format", "ban-keywords"]
```
kjin: Is it alright that we have lines that are commented out? Also, I noticed that the value was changed from `[false]` -- would it be the same to only have this set to `[true, "ban-keywords"]`?

src/auth/oauth2client.ts
```diff
@@ -339,6 +338,7 @@ export default class OAuth2Client extends AuthClient {
         retry = false;
         // Force token refresh
         this.refreshAccessToken(() => {
+          // tslint:disable-next-line:no-use-before-declare
```
matthewloring: Would it be better to hoist the definition of `authCb`?

ofrobots: There is a circular dependency.

matthewloring: Sorry, I meant declaration, not definition. Something like:

```javascript
let authCb;
const postRequestCb = ... authCb ...
authCb = ... postRequestCb ...
```

tslint.json
ofrobots: tslint doesn't seem to mind comments in JSON files. I did want to leave behind some indication of how we are diverging from the google style guide. While these aren't strictly JSON anymore, I think we're fine in practice.

The second suggestion of `[true, "ban-keywords"]` was good. Done.

src/auth/oauth2client.ts
ofrobots: Done.

PR #144, opened 2017-05-09
move typescript files into separate ts/ dir

ofrobots:
It seems that tslint wasn't being run on the test files. This commit
also fixes that.

This supercedes #141. This ensures that the output js files remain in their original location as before this PR avoiding an unnecessary semver major.


src/test/test.googleauth.ts
```diff
@@ -357,11 +357,11 @@ describe('GoogleAuth', () => {
 
     it('should read the stream and create a jwt', (done) => {
       // Read the contents of the file into a json object.
-      const fileContents = fs.readFileSync('./test/fixtures/private.json', 'utf-8');
+      const fileContents = fs.readFileSync('./src/test/fixtures/private.json', 'utf-8');
```
JustinBeckwith: Do tests really belong in src?

ofrobots:
I think it is worse to have input and output files commingled. @JustinBeckwith would you be happier if I called it `ts/` instead of `src/`?

ofrobots:
Preemptively moved stuff to ts/. @JustinBeckwith PTAL.

ofrobots:
To elaborate on the challenge/constraints here:

* The output JavaScript files must be in lib/ because external modules depend upon internal files.
* The directory structure of input and output directory must be comparable. That is, tests should be able to do require('../lib/file').
* We do not want input and output files to be commingled. We cannot put output files in a toplevel build/ directory because of the previous bullet. Specifically, going forward I want to start using a package.json `files` section rather than a .npmignore (i.e. more tightly control what actually gets shipped in the tarball).

DominicKramer:
Another possibility would be to have a `lib.ts` directory that is transpiled to `lib` and a `test.ts` directory that is transpiled to `test`.

ofrobots:
@DominicKramer I am not sure that would work. Example:

```ts
import GoogleAuth from '../lib/auth/googleauth';
```

This will have to be changed to `../lib.ts/auth/googleauth` to make it compile, but the javascript transpiled version will no longer work.

Perhaps there is way to work around this with `tsconfig.json` config, but I don't know it.

DominicKramer:
@ofrobots Ok.  That makes sense.

PR #143, opened 2017-05-05
Apply official clang format style

ofrobots:
This is dependent on #141 and #142.

PR #145, opened 2017-05-11
Generate and export type definitions

ofrobots:
We now generate type definitions that can ship as part of the module.

This is indirectly dependent on #143, and those commits will show up here until that lands.

### DIFF
--- a/ts/lib/auth/refreshclient.ts
+++ b/ts/lib/auth/refreshclient.ts
@@ -36,21 +36,19 @@ export default class UserRefreshClient extends Auth2Client {
   }
 
   // Executes the given callback if it is not null.
-  private callback =
-      (c, err?, res?) => {
-        if (c) {
-          c(err, res);
-        }
-      }
+  private callback(c, err?, res?) {
+    if (c) {
+      c(err, res);
+    }
+  }
 
-                         /**
-                          * Refreshes the access token.
-                          * @param {object=} ignored_
-                          * @param {function=} opt_callback Optional callback.
-                          * @private
-                          */
-                         protected refreshToken(ignored_, opt_callback):
-                             request.Request {
+  /**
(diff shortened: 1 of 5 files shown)


ts/lib/auth/refreshclient.ts
```diff
@@ -41,13 +43,14 @@ export default class UserRefreshClient extends Auth2Client {
...
       }
 
-  /**
-   * Refreshes the access token.
-   * @param {object=} ignored_
-   * @param {function=} opt_callback Optional callback.
-   * @private
-   */
-  protected refreshToken(ignored_, opt_callback) {
+                         /**
```
JustinBeckwith: What's up with this?

tsconfig.json
```diff
@@ -5,7 +5,9 @@
         "noImplicitAny": false,
         "sourceMap": true,
         "alwaysStrict": true,
-        "outDir": "./"
+        "outDir": "./",
```
JustinBeckwith: Should we add the d.ts to .gitignore, since the build produces it?

ts/lib/auth/refreshclient.ts
ofrobots: This commit is from #143. Let's move the conversation there.

ofrobots: apparently it is not! sorry for the confusion.

tsconfig.json
ofrobots: `types/` is already in `.gitignore` and I would prefer to use paths rather than specific file extensions.

ts/lib/auth/refreshclient.ts
ofrobots: Fixed.

ofrobots:
This is rebased and ready for another look.

PR #147, opened 2017-05-12
use files section in package.json

ofrobots:
This provides better control over what gets included in the package.

I am not sure why npm is still including .js.map files in the pack. The .npmignore should still be referenced. Any one have any ideas?


ofrobots:
Dealt with the `.js.map` file issue. Another look would be great.

PR #149, opened 2017-06-11
Treat 4xx and 5xx responses as errors

saicheems:
At the moment, 4xx and 5xx responses are returned in
google-api-nodejs-client as regular responses, not errors. From the user
perspective, this is an odd pattern.

This commit forces 4xx and 5xx responses to be returned as errors, and
not be treated as a successful requests.

### DIFF
--- a/ts/lib/transporters.ts
+++ b/ts/lib/transporters.ts
@@ -103,8 +103,8 @@ export class DefaultTransporter {
           (err as RequestError).code = body.error.code || res.statusCode;
         }
         body = null;
-      } else if (res.statusCode >= 500) {
-        // Consider all '500 responses' errors.
+      } else if (res.statusCode >= 400) {
+        // Consider all 4xx and 5xx responses errors.
         err = new RequestError(body);
         (err as RequestError).code = res.statusCode;
         body = null;
--- a/ts/lib/auth/oauth2client.ts
+++ b/ts/lib/auth/oauth2client.ts
@@ -297,7 +297,10 @@ export default class OAuth2Client extends AuthClient {
 
     return this.refreshToken(
         thisCreds.refresh_token, (err, tokens, response) => {
-          if (err) {
+          // If the error code is 403 or 404, fall-through to replace the error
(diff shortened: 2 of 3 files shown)


saicheems:
See also:

https://github.com/google/google-api-nodejs-client/issues/424

saicheems:
fyi @tcoffee-google @vchudnov-g

saicheems:
+jmdobry

saicheems:
+ @jmdobry 
+ @JustinBeckwith

saicheems:
+ @ofrobots

stephenplusplus:
Should we special-case 404, or possibly let any non-2xx trigger an Error?

ts/lib/transporters.ts
```diff
@@ -95,9 +95,8 @@ export class DefaultTransporter {
         }
 
         body = null;
-
-      } else if (res.statusCode >= 500) {
-        // Consider all '500 responses' errors.
+      } else if (res.statusCode === 404 || res.statusCode >= 500) {
```
JustinBeckwith: Technically wouldn't any 4xx or 5xx be an error?

JustinBeckwith:
Hah, I had similar feedback - at least any 4xx or 5xx.

ts/lib/transporters.ts
saicheems: Changed to anything >= 400

saicheems:
OK, looks like there was some conflict with a just merged PR - I resolved by casting all errors to `RequestError`, but I don't know if that's correct.

Otherwise I've changed it to consider all response status codes >= 400 to be errors.

saicheems:
Note, also had to add the exception in `oauth2client.ts` for the 403 case, which also has a special error message.

saicheems:
Ping, don't have permission to merge this myself

lukesneeringer:
I do not have permission either.

ofrobots:
Landed. Also, I have added @lukesneeringer as an owner of this repo.

PR #151, opened 2017-06-25
enable noImplicitAny

JustinBeckwith:
Trying to clean up the typing support here by enabling noImplicitAny on the compiler rules.

### DIFF
--- a/ts/lib/auth/computeclient.ts
+++ b/ts/lib/auth/computeclient.ts
@@ -16,17 +16,16 @@
 
 import * as request from 'request';
 
-import {RequestCallback, RequestError} from './../transporters';
+import {BodyResponseCallback, RequestError} from './../transporters';
 import Auth2Client from './oauth2client';
 
-export interface Tokens {
+export interface Token {
   expires_in: number;
   expiry_date: number;
 }
 
-export interface RefreshTokenCallback {
-  (err: Error, tokens: Tokens, response: request.RequestResponse): void;
-}
+export declare type RefreshTokenCallback =
+    (err: Error, token: Token, response: request.RequestResponse) => void;
 
 export default class Compute extends Auth2Client {
   /**
@@ -63,7 +62,7 @@ export default class Compute extends Auth2Client {
   /**
    * Refreshes the access token.
    * @param {object=} ignored_
-   * @param {function=} opt_callback Optional callback.
+   * @param {function=} callback Optional callback.
    */
   protected refreshToken(ignored: any, callback?: RefreshTokenCallback):
       request.Request {
@@ -71,14 +70,14 @@ export default class Compute extends Auth2Client {
     // request for new token
     return this.transporter.request(
         {method: 'GET', uri: uri, json: true}, (err, body, response) => {
-          const tokens = body as Tokens;
-          if (!err && tokens && tokens.expires_in) {
-            tokens.expiry_date =
-                ((new Date()).getTime() + (tokens.expires_in * 1000));
-            delete tokens.expires_in;
+          const token = body as Token;
+          if (!err && token && token.expires_in) {
+            token.expiry_date =
+                ((new Date()).getTime() + (token.expires_in * 1000));
+            delete token.expires_in;
           }
           if (callback) {
-            callback(err, tokens, response);
+            callback(err, token, response);
           }
         });
   }
@@ -93,7 +92,7 @@ export default class Compute extends Auth2Client {
    */
   protected postRequest(
       err: Error, result: any, response: request.RequestResponse,
-      callback: RequestCallback) {
+      callback: BodyResponseCallback) {
     if (response && response.statusCode) {
       let helpfulMessage = null;
       if (response.statusCode === 403) {
--- a/ts/lib/auth/computeclient.ts
+++ b/ts/lib/auth/computeclient.ts
@@ -16,7 +16,17 @@
 
 import * as request from 'request';
 
+import {RequestCallback, RequestError} from './../transporters';
 import Auth2Client from './oauth2client';
+
+export interface Tokens {
+  expires_in: number;
+  expiry_date: number;
+}
+
+export interface RefreshTokenCallback {
+  (err: Error, tokens: Tokens, response: request.RequestResponse): void;
+}
 
 export default class Compute extends Auth2Client {
   /**
@@ -55,11 +65,13 @@ export default class Compute extends Auth2Client {
    * @param {object=} ignored_
    * @param {function=} opt_callback Optional callback.
    */
-  protected refreshToken(ignored, callback?): request.Request {
+  protected refreshToken(ignored: any, callback?: RefreshTokenCallback):
+      request.Request {
     const uri = this._opts.tokenUrl || Compute._GOOGLE_OAUTH2_TOKEN_URL;
     // request for new token
     return this.transporter.request(
-        {method: 'GET', uri: uri, json: true}, (err, tokens, response) => {
+        {method: 'GET', uri: uri, json: true}, (err, body, response) => {
+          const tokens = body as Tokens;
           if (!err && tokens && tokens.expires_in) {
             tokens.expiry_date =
                 ((new Date()).getTime() + (tokens.expires_in * 1000));
@@ -79,7 +91,9 @@ export default class Compute extends Auth2Client {
    * @param {object} response The HTTP response.
    * @param {Function} callback The callback.
    */
-  protected postRequest(err, result, response, callback) {
+  protected postRequest(
+      err: Error, result: any, response: request.RequestResponse,
+      callback: RequestCallback) {
     if (response && response.statusCode) {
       let helpfulMessage = null;
       if (response.statusCode === 403) {
@@ -102,7 +116,7 @@ export default class Compute extends Auth2Client {
           err.message = helpfulMessage;
         } else {
           err = new Error(helpfulMessage);
-          err.code = response.statusCode;
+          (err as RequestError).code = response.statusCode;
         }
       }
     }
--- a/ts/lib/auth/googleauth.ts
+++ b/ts/lib/auth/googleauth.ts
@@ -23,7 +23,7 @@ import * as util from 'util';
 
 import {DefaultTransporter, Transporter} from '../transporters';
 
-import {RequestCallback} from './../transporters';
+import {BodyResponseCallback} from './../transporters';
 import Compute from './computeclient';
 import IAMAuth from './iam';
 import JWTAccess from './jwtaccess';
@@ -95,7 +95,7 @@ export default class GoogleAuth {
 
   /**
    * Obtains the default project ID for the application..
-   * @param {function=} opt_callback Optional callback.
+   * @param {function=} callback Optional callback.
    */
   public getDefaultProjectId(callback: ProjectIdCallback) {
     // In implicit case, supports three environments. In order of precedence,
@@ -161,15 +161,15 @@ export default class GoogleAuth {
 
   /**
    * Obtains the default service-level credentials for the application..
-   * @param {function=} opt_callback Optional callback.
+   * @param {function=} callback Optional callback.
    */
   public getApplicationDefault(
-      opt_callback: (err: Error, credential: any, projectId: string) => void) {
+      callback?: (err: Error, credential: any, projectId: string) => void) {
     // If we've already got a cached credential, just return it.
     if (this.cachedCredential) {
       setImmediate(() => {
         this.callback(
-            opt_callback, null, this.cachedCredential, this._cachedProjectId);
+            callback, null, this.cachedCredential, this._cachedProjectId);
       });
     } else {
       // Inject our own callback routine, which will cache the credential once
@@ -181,12 +181,12 @@ export default class GoogleAuth {
           this.getDefaultProjectId((err2, projectId) => {
             setImmediate(() => {
               // Ignore default project error
-              this.callback(opt_callback, null, result, projectId);
+              this.callback(callback, null, result, projectId);
             });
           });
         } else {
           setImmediate(() => {
-            this.callback(opt_callback, err, result);
+            this.callback(callback, err, result);
           });
         }
       };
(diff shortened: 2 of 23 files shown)


ts/lib/auth/computeclient.ts
```diff
@@ -16,8 +16,18 @@
...
 
+import {RequestCallback, RequestError} from './../transporters';
 import Auth2Client from './oauth2client';
 
+export interface Tokens {
+  expires_in: number;
+  expiry_date: number;
+}
+
+export interface RefreshTokenCallback {
```
DominicKramer: Why did you choose to use an interface here?  If you wanted to, I think you could also use something like the following.
```
export declare type RefreshTokenCallback = (err: Error, tokens: Tokens, response: request.RequestReponse) => void;
```

ts/lib/auth/computeclient.ts
```diff
@@ -102,7 +116,7 @@ export default class Compute extends Auth2Client {
           err.message = helpfulMessage;
         } else {
           err = new Error(helpfulMessage);
```
DominicKramer: Is it possible to directly create a `RequestError` here to avoid the cast later?

ts/lib/auth/googleauth.ts
```diff
@@ -353,8 +362,8 @@ export default class GoogleAuth {
    * @param {object=} json The input object.
    * @param {function=} opt_callback Optional callback.
    */
-  public fromJSON(json, opt_callback) {
-    let client = null;
+  public fromJSON(json: any, opt_callback: (err: Error, client?: any) => void) {
```
DominicKramer: Do you know if Typescript supplies a `json` type?

ts/lib/auth/googleauth.ts
```diff
@@ -465,36 +478,36 @@ export default class GoogleAuth {
...
    * @api private
    */
-  private _mockWellKnownFilePath(filePath) {
+  private _mockWellKnownFilePath(filePath: string) {
     return filePath;
   }
 
   // Executes the given callback if it is not null.
-  private callback(c, ...args) {
+  private callback(c: Function, ...args: any[]) {
```
DominicKramer: Would it be possible to get rid of `any` here by using:
```
private callback<T>(c: Function, ...args: T[]) {
...
}
```
or does the usage of this function use args of multiple types?

ts/lib/auth/jwtclient.ts
```diff
@@ -107,7 +109,7 @@ export default class JWT extends Auth2Client {
    * Get the initial access token using gToken.
    * @param {function=} opt_callback Optional callback.
    */
-  public authorize(opt_callback?) {
+  public authorize(opt_callback?: Function) {
```
DominicKramer: It looks like `opt_callback` could have the type:
```
(err: Error, result: any) => void
```

ts/lib/auth/jwtclient.ts
```diff
@@ -148,7 +152,7 @@ export default class JWT extends Auth2Client {
    * @param {object=} json The input object.
    * @param {function=} opt_callback Optional callback.
    */
-  public fromJSON(json, opt_callback) {
+  public fromJSON(json: any, opt_callback: (err?: Error) => void) {
```
DominicKramer: Based on its name, it looks like `opt_callback` should be of an optional type.

ts/lib/auth/oauth2client.ts
```diff
@@ -18,17 +18,26 @@ import * as querystring from 'querystring';
...
 import PemVerifier from './../pemverifier';
+import {RequestCallback, RequestError} from './../transporters';
 import AuthClient from './authclient';
+import Credentials from './credentials';
 import LoginTicket from './loginticket';
 
 const merge = require('lodash.merge');
 
+export interface GenerateAuthUrlOpts {
+  response_type?: string;
```
DominicKramer: Could the value of `response_type` be restricted to a set of strings so that its type could be `string1|string2|...`?

ts/lib/transporters.ts
```diff
@@ -19,7 +19,18 @@ import * as request from 'request';
 // tslint:disable-next-line
 const pkg = require('../package.json');
 
-export interface Transporter { request(opts, opt_callback): any; }
+export interface Transporter {
+  request(opts: any, opt_callback: RequestCallback): any;
```
DominicKramer: Should `opt_callback` be of an optional type?

ts/test/test.googleauth.ts
```diff
@@ -497,7 +500,7 @@ describe('GoogleAuth', () => {
 
     it('should error on non-string file path', (done) => {
       const auth = new GoogleAuth();
-      auth._getApplicationCredentialsFromFilePath(2, (err) => {
+      auth._getApplicationCredentialsFromFilePath(2 as any, (err: Error) => {
```
DominicKramer: Just curious.  Why is the cast to `any` needed here?

JustinBeckwith: This is kind of funny.  The test was verifying the type of the first argument.  I added typing to the method, and now it won't compile if you try to pass a Number. The test is in TypeScript as well, so it's impossible to pass the wrong type unless you cast to <any>.

ts/lib/auth/computeclient.ts
JustinBeckwith: Nice!  Switched this one over.

ts/lib/auth/googleauth.ts
ofrobots: At the least it should be a `string`.

ts/lib/auth/googleauth.ts
```diff
@@ -411,7 +423,8 @@ export default class GoogleAuth {
    * @param {string} - The API key string
    * @param {function=} - Optional callback function
    */
-  public fromAPIKey(apiKey, opt_callback) {
+  public fromAPIKey(
+      apiKey: string, opt_callback: (err: Error, client?: JWTClient) => void) {
```
ofrobots: nit: `opt_callback` seems wrong as-is. The `opt` implies that this should be `opt_callback?`, but I would recommend changing it to `callback?`.

ts/lib/auth/googleauth.ts
ofrobots: The args can all be different types.

In a follow-on it would be better to get rid of this function altogether.

ts/lib/auth/iam.ts
```diff
@@ -45,7 +50,9 @@ export default class IAMAuth {
    * @param {function} metadataFn a callback invoked with object
    *                   containing request metadata.
    */
-  public getRequestMetadata(unused_uri: string, metadataFn) {
+  public getRequestMetadata(
+      unused_uri: string,
+      metadataFn: (err: Error, metadata: RequestMetadata) => void) {
```
ofrobots: nit: `metadata?`.

ts/lib/auth/jwtaccess.ts
```diff
@@ -55,7 +57,8 @@ export default class JWTAccess {
    * @param {function} metadataCb a callback invoked with the jwt
    *                   request metadata.
    */
-  public getRequestMetadata(authURI: string, metadataCb) {
+  public getRequestMetadata(
+      authURI: string, metadataCb: (err: Error, headers: any) => void) {
```
ofrobots: `headers?`

Also: can we avoid `any` here?

ts/lib/auth/jwtaccess.ts
```diff
@@ -85,7 +88,7 @@ export default class JWTAccess {
    * @param {object=} json The input object.
    * @param {function=} opt_callback Optional callback.
    */
-  public fromJSON(json, opt_callback) {
+  public fromJSON(json: any, opt_callback: (err: Error) => void) {
```
ofrobots: `json` can be `string`.

ts/lib/auth/jwtaccess.ts
```diff
@@ -114,7 +117,8 @@ export default class JWTAccess {
    * @param {object=} stream The input stream.
    * @param {function=} opt_callback Optional callback.
    */
-  public fromStream(stream, opt_callback) {
+  public fromStream(
+      stream: stream.Readable, opt_callback: (err: Error) => void) {
```
ofrobots: `opt_callback` => `callback?`.

ts/lib/auth/jwtclient.ts
```diff
@@ -74,7 +75,8 @@ export default class JWT extends Auth2Client {
    * @param {string} opt_uri the URI being authorized.
    * @param {function} metadataCb
    */
-  public getRequestMetadata(opt_uri: string, metadataCb) {
+  public getRequestMetadata(
+      opt_uri: string, metadataCb: (err: Error, result?: any) => void) {
```
ofrobots: I would have recommended `opt_uri` => `uri?`, but I don't think those are the actual semantics here. I think this is supposed to be a nullable string? not really optional...

ts/test/test.googleauth.ts
DominicKramer: Strange.

ts/lib/transporters.ts
```diff
@@ -19,7 +19,18 @@ import * as request from 'request';
...
+}
+
+export interface RequestCallback {
```
ofrobots: Multiple people have lost multiple hours of debugging time because of this callback. Can we call this something that people don't easily mistake for `request.RequestCallback`?

Perhaps `BodyResponseCallback` or something like that?

ts/lib/auth/computeclient.ts
JustinBeckwith: Eugh, sadly it gets super hairy.  These errors are threaded through tons of callbacks, some of which originate outside of here (in request).  This seemed like the safest way to pin the code on the error.

ts/lib/auth/googleauth.ts
JustinBeckwith: Not that I know of - I think it's pretty much a string or a dynamically typed object :)

ts/lib/auth/googleauth.ts
JustinBeckwith: I was able to add an `err?: Error` in the middle there, but this method is used with a varying number of varying types after that.  There are `n` arguments, so it's hard to figure out what the types would be.

ts/lib/auth/jwtclient.ts
JustinBeckwith: Nice catch!  Actually was able to type the result to `Credential` as well.

ts/lib/auth/jwtclient.ts
JustinBeckwith: fixed

ts/lib/auth/computeclient.ts
DominicKramer: If it is used to annotate an exported member or function then it would need to be exported.

ts/lib/auth/computeclient.ts
DominicKramer: Ok sounds good.

ts/lib/auth/googleauth.ts
DominicKramer: Something like:
```
interface Json {
    [key: string]: string | number | boolean | Json;
}
```
isn't perfect, but using that or `string` could help distinguish if the type should be the string representation of the JSON or the JSON string decoded as an object.

ts/lib/auth/googleauth.ts
DominicKramer: Yeah, that is what I suspected.

ts/lib/auth/jwtclient.ts
DominicKramer: Awesome.

ts/lib/auth/oauth2client.ts
JustinBeckwith: Yeah, we can use an enum here I think.  Do you know the valid values?

ts/lib/transporters.ts
JustinBeckwith: Good call

ts/lib/auth/computeclient.ts
```diff
@@ -16,8 +16,18 @@
 
 import * as request from 'request';
 
+import {RequestCallback, RequestError} from './../transporters';
 import Auth2Client from './oauth2client';
 
+export interface Tokens {
```
JustinBeckwith: singularized

ts/lib/auth/googleauth.ts
```diff
@@ -268,7 +275,8 @@ export default class GoogleAuth {
    * @return {boolean} Returns true if the callback has been executed; false otherwise.
    * @api private
    */
-  public _tryGetApplicationCredentialsFromWellKnownFile(opt_callback?) {
+  public _tryGetApplicationCredentialsFromWellKnownFile(
+      opt_callback?: (err: Error, result: any) => void) {
```
JustinBeckwith: Added the ? to the result.  On dumping the any(s) here - most of the places where I left an any, it was because the data was pulled from a JSON.parse(...) where I'm not 100% sure about the contents.  I want to create a follow up issue to lock down the types even more if I can :)

ts/lib/auth/googleauth.ts
JustinBeckwith: I ensured the opt_callbacks were all actually optional, and renamed to just callback

ts/lib/auth/iam.ts
JustinBeckwith: fixed

ts/lib/auth/jwtaccess.ts
JustinBeckwith: I don't think it can.  This is an object returned from `JSON.parse(...)`

ts/lib/auth/jwtaccess.ts
JustinBeckwith: Made headers optional.  I'd like to take these `any`s and address them after we get the low hanging fruit here.

ts/lib/auth/jwtaccess.ts
JustinBeckwith: fixed and renamed

ts/lib/auth/jwtclient.ts
JustinBeckwith: I don't think `uri?` would be appropriate, since the metadataCb is actually required as written.  The `?` only really works on trailing properties, right?

ts/lib/transporters.ts
JustinBeckwith: Good call - updated!

PR #157, opened 2017-07-31
state is a valid option of GenerateAuthUrlOpts

stevemao:
See https://auth0.com/docs/protocols/oauth2/oauth-state


lukesneeringer:
@bantini Please check the spec, ensure validity, and add a test and submit this PR if so.

stevemao:
This is a ts type signature change which involves no functionality changes. I don't know if we need to write a test against a type signature.

JustinBeckwith:
*shrugs* if it's just a typing fix, I don't think we need a test.

JustinBeckwith:
Addressed by #184.  Thanks!

PR #159, opened 2017-08-14
refactor: keep CommonJS compat for module

ofrobots:
GoogleAuth is being exported as an ES6 class. This breaks compatibility
with existing JS users of this library for the main module export. This
change adds an index.js that re-exports GoogleAuth as a function.

Note that we are not using this hack for every single file in this
module. Anyone reaching into the internals of this module will need to
update their code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 node_modules/*
-lib/
-test/
+/lib/
+/test/
 types/
 .idea
 .vscode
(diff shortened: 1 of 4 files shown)


JustinBeckwith:
That was probably my mistake.  Thanks for catching.  Do we have tests that would capture this moving forward?

LGTM.

ofrobots:
Added test.

PR #162, opened 2017-08-15
feature: export DefaultTransport

ofrobots:
Properly export DefaultTransport - some dependent libraries are using
this via loading an internal file.


JustinBeckwith:
If customers are relying on this, I would love to have some kind of test we can use to verify we don't break it in the future?

PR #164, opened 2017-08-17
fix: checkIsGCE: report unexpected errors

ofrobots:


ofrobots:
The main reason I put this together is that I am seeing reports of 'Unable to acquire ADC' on GCF environments. The expectation is that these are socket timeouts, but it would be good to know the precise error. The existing code was unable to distinguish expected and unexpected error in DNS resolution.

PR #173, opened 2017-11-04
Switch from gulp to gts, TypeScript 2.6

JustinBeckwith:
Moving from gulp to gts.  Hoping to do this without any API breaking changes.


src/auth/credentials.ts
```diff
@@ -17,6 +17,6 @@
 export interface Credentials {
   refresh_token?: string;
   expiry_date?: number;
-  access_token?: string;
+  access_token?: string|null;
```
ofrobots: So it can be `string` or `null` or `undefined`. Is there a way to avoid the `null`?

src/auth/jwtclient.ts
```diff
@@ -25,11 +25,11 @@ const isString = require('lodash.isstring');
 const noop = Function.prototype;
 
 export class JWT extends OAuth2Client {
-  public email: string;
-  public keyFile: string;
-  public key: string;
-  public scopes: string|string[];
-  public subject: string;
+  public email?: string;
+  public keyFile?: string|null;
```
ofrobots: nit: likewise, it would be good if we could eliminate one of null vs. undefined if possible. Here and below.

src/auth/oauth2client.ts
```diff
@@ -35,12 +35,17 @@ export interface GenerateAuthUrlOpts {
...
 
 export class OAuth2Client extends AuthClient {
-  private _redirectUri: string;
-  private _certificateCache: any = null;
-  private _certificateExpiry: Date = null;
-  protected _opts: any;
-  public _clientId: string;
-  public _clientSecret: string;
+  private redirectUri?: string;
+  private certificateCache: any = null;
```
ofrobots: likewise

src/auth/oauth2client.ts
```diff
@@ -49,17 +54,17 @@ export class OAuth2Client extends AuthClient {
...
    * @param {string=} clientSecret The authentication client secret.
    * @param {string=} redirectUri The URI to redirect to after completing the auth request.
-   * @param {Object=} opt_opts optional options for overriding the given parameters.
+   * @param {Object=} optOpts optional options for overriding the given parameters.
    * @constructor
    */
   constructor(
       clientId?: string, clientSecret?: string, redirectUri?: string,
-      opt_opts?: any) {
+      optOpts?: any) {
```
ofrobots: just drop the `opt` prefix now. he function type is sufficient.

src/auth/refreshclient.ts
```diff
@@ -21,7 +21,9 @@ import {BodyResponseCallback} from './../transporters';
 import {OAuth2Client} from './oauth2client';
 
 export class UserRefreshClient extends OAuth2Client {
-  public _refreshToken: string;
+  // TODO: refactor tests to make this private
+  // tslint:disable-next-line
```
ofrobots: let be specific about the lint rule you want to disable. BTW, note that gts 0.5.0 will no longer complain about this.

package.json
```diff
@@ -4,10 +4,10 @@
   "author": "Google Inc.",
   "description": "Google APIs Authentication Client Library for Node.js",
   "engines": {
-    "node": ">=0.10"
+    "node": ">=4"
```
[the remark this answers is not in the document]
JustinBeckwith: We weren't testing on on 0.10 ... so it's hard to say what was real here.  What do you think we should do?  
- [ ] Keep this at 0.10 and add 0.10 back to travis config
- [ ] Move this to 4, and bump semver major
- [ ] Move this to 4, and don't bump semver major since technically 0.10 has been out of support a loooooooong time

I'm cool with any of these :)

ace-n:
My vote's for moving to 4 and doing a semver-major bump. IIRC 0.10 is no
longer officially supported, but unless we know nobody's using it with this
repo there is still a potential for breaking changes.

src/auth/credentials.ts
JustinBeckwith: The only place using null was the test, so I just changed it to undefined 🤷‍♂️

src/auth/googleauth.ts
```diff
@@ -41,27 +41,24 @@ export class GoogleAuth {
...
 
   // Note:  this properly is only public to satisify unit tests.
   // https://github.com/Microsoft/TypeScript/issues/5228
   get isGCE() {
-    return this._isGCE;
+    return this.checkIsGCE;
   }
 
-  private _cachedProjectId: string;
+  private intCachedProjectId: string;
```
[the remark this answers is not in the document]
JustinBeckwith: I had used this since `gts fix` was complaining about _propertyNames.  Now that I know 0.5.1 lifts that restriction, I'll just use a _ here.

src/auth/googleauth.ts
```diff
@@ -85,6 +82,7 @@ export class GoogleAuth {
   /**
    * Convenience field mapping in the OAuth2 credential type.
    */
+  // tslint:disable-next-line
```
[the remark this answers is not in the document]
JustinBeckwith: Added!

src/auth/jwtclient.ts
JustinBeckwith: I hear you.  There are a few reasons I don't want to do this here.
1. For now - I just want this PR to represent the truth as we see it with the supported types.  We can come back around in follow up PRs and make it be what we want, but as of it now this is mostly documenting the types that are used.
2. These are both passed as constructor parameters, with tests that try passing null values.  So technically null is a valid value here.
3. There are places in the API where the JWT() is instantiated with no constructor parameters.  So technically undefined is valid as well.

I know this code is gross, but I'm scared to try touching too much in this already large PR.

src/auth/refreshclient.ts
JustinBeckwith: fixed!

src/auth/oauth2client.ts

package.json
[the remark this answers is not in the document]
JustinBeckwith: Ok, so I tested 0.10 to see if it even worked.  Result: 
image

There's no way anyone was actually using this - it fails on npm install, with about 100 peer dependency failures.  I vote we keep supported at 4, and don't bother with the semver major bump.  I don't know how it's possible that this is breaking someone who wasn't already broken.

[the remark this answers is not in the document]
JustinBeckwith: Doubling back.  As part of the PR, we're now changing the way the module is imported/exported, so it's a semver major bump anyway :)

src/auth/oauth2client.ts
```diff
@@ -500,21 +503,26 @@ export class OAuth2Client extends AuthClient {
                 null, response);
             return;
           }
-          const cacheControl = response.headers['cache-control'];
+
+          let cacheControl = response ? response.headers['cache-control'] : '';
+          if (cacheControl instanceof Array) {
+            cacheControl = cacheControl.length > 0 ? cacheControl[0] : '';
```
ofrobots: I'm not too familiar with the `cache-control` control header. Under what cases can it be an array, and why is it okay to pick the first one? Does this need a code comment?

src/auth/oauth2client.ts
```diff
@@ -500,21 +503,26 @@ export class OAuth2Client extends AuthClient {
...
+          }
+
           let cacheAge = -1;
           if (cacheControl) {
             const pattern = new RegExp('max-age=([0-9]*)');
             const regexResult = pattern.exec(cacheControl);
-            if (regexResult.length === 2) {
+            if (regexResult && regexResult.length === 2) {
```
ofrobots: Not sure if you'd prefer this, but I find the following more readable for regexp.exec:

```js
const [match, maxAge] = pattern.exec(cacheControl) || [];
if (match) cacheAge = maxAge;
```

README.md
```diff
@@ -98,6 +98,23 @@ authFactory.getApplicationDefault(function(err, authClient) {
 });
 ```
 
+## Changelog
+
+### 0.11.0 -> 1.0.0
```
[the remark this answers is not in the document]
JustinBeckwith: ack

README.md
```diff
@@ -98,6 +98,23 @@ authFactory.getApplicationDefault(function(err, authClient) {
...
+
+- Support for Node.js 0.10 and 0.12 has officially been dropped.  This library is tested against Node.js version 4 and above.
+- The API has been optimized for ES Modules.  The `GoogleAuth` and `DefaultTransporter` objects are now properties of the module.  For example:
+
+```js
+// es6/TypeScript
+import {GoogleAuth, DefaultTransporter} from 'google-auth-library';
+
+// es5
+const GoogleAuth = require('google-auth-library').GoogleAuth;
```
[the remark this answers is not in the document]
JustinBeckwith: Realistically, why not.  Doubt there are many people with v4 projects out there adopting this or the first time.

src/auth/oauth2client.ts
JustinBeckwith: I switched this up because the NetIncomingHeader type in the http says it's string|string[].  Realistically, this was working before so I just returned it to original form and did a cast.

JustinBeckwith:
@ofrobots I updated this to work with `gts@0.5.1`.  Apologies for the massive set of changes.  

Right now the only breaking changes are the way we export GoogleAuth from `index.ts`, and the change to drop 0.10 (which wasn't really a break).  I know we chatted about a bunch of other changes - but I'm starting to think it just makes sense to get this out to 1.0 with the real types finally in place.  If we want to go do axios or something like that, lets plan it another iteration.  Thoughts?

ofrobots:
It would be good to drop the dependency on `request`, and specially the callback with reversed arguments fixed before 1.0. Once we reach 1.0, I would expect the frequency of semver majors to be quite low.

JustinBeckwith:
That works :)  Do you want these changes in master, or a 2.0 branch? I can do the axios upgrade pretty quickly after this lands.

src/transporters.ts
```diff
@@ -17,33 +17,44 @@
...
 export interface Transporter {
-  request(opts: any, callback?: BodyResponseCallback): any;
+  request(opts: request.Options, callback?: BodyResponseCallback):
+      request.Request;
 }
 
 export interface BodyResponseCallback {
-  (err: Error, body: any, res?: request.RequestResponse): void;
+  // The `body` object is a truly dynamic type.  It must be `any`.
+  // tslint:disable-next-line
```
ofrobots: specific the rule you want to disable. `tslint:disable-next-line no-any`

test/test.googleauth.ts
```diff
@@ -160,7 +162,9 @@ describe('GoogleAuth', () => {
 
     it('should error on null json', (done) => {
       const auth = new GoogleAuth();
-      auth.fromJSON(null, (err) => {
+      // Test verifies invalid parameter tests, which requires cast to any.
+      // tslint:disable-next-line
```
ofrobots: specify which rule to skip

test/test.googleauth.ts
```diff
@@ -174,8 +178,10 @@ describe('GoogleAuth', () => {
         before(() => {
           auth = new GoogleAuth();
         });
-        it('Should error given an invalid api key', (done) => {
-          auth.fromAPIKey(null, (err) => {
+        it('Should error given an invalid api key', done => {
+          // Test verifies invalid parameter tests, which requires cast to any.
+          // tslint:disable-next-line
```
ofrobots: ditto

test/test.googleauth.ts
```diff
@@ -408,47 +435,51 @@ describe('GoogleAuth', () => {
...
-        // Ensure that the correct bits were pulled from the stream.
-        assert.equal(json.private_key, result.key);
-        assert.equal(json.client_email, result.email);
-        assert.equal(null, result.keyFile);
-        assert.equal(null, result.subject);
-        assert.equal(null, result.scope);
-
+        assert(result);
+        const jwt = result as JWT;
+        if (result) {
```
ofrobots: nit: don't intermingle the two aliases jwt and result. Use jwt here onwards. Or move the cast to inside the if.

test/test.googleauth.ts
```diff
@@ -399,7 +424,9 @@ describe('GoogleAuth', () => {
 
     it('should error on null stream', (done) => {
       const auth = new GoogleAuth();
-      auth.fromStream(null, (err) => {
+      // Test verifies invalid parameter tests, which requires cast to any.
+      // tslint:disable-next-line
```
ofrobots: ditto

test/test.googleauth.ts
```diff
@@ -477,18 +508,22 @@ describe('GoogleAuth', () => {
...
 
     it('should error on null file path', (done) => {
       const auth = new GoogleAuth();
-      auth._getApplicationCredentialsFromFilePath(null, (err) => {
-        assert.equal(true, err instanceof Error);
-        done();
-      });
+      (auth as
+           // Test verifies invalid parameter tests, which requires cast to any.
+           // tslint:disable-next-line
```
ofrobots: ditto.
I will stop pointing these out from here on.

src/auth/oauth2client.ts
```diff
@@ -587,8 +612,8 @@ export class OAuth2Client extends AuthClient {
           'No expiration time in token: ' + JSON.stringify(payload));
     }
 
-    const iat = parseInt(payload.iat, 10);
-    const exp = parseInt(payload.exp, 10);
+    const iat = payload.iat as number;
```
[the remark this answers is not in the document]
JustinBeckwith: Nice tip on the isNaN check.

PR #182, opened 2017-11-18
chore: switch to axios

JustinBeckwith:
Switching out request for axios, and making the API support callback and async styles.


JustinBeckwith:
@ofrobots all the tests are passing, this is ready for a pass.  Apologies for the size of the review, couldn't really find a way to split it up.

package.json
```diff
@@ -20,19 +20,17 @@
...
+    "axios": "^0.17.1",
     "gtoken": "^1.2.3",
     "jws": "^3.1.4",
-    "lodash.isstring": "^4.0.1",
-    "lodash.merge": "^4.6.0",
-    "request": "^2.81.0"
+    "lodash.isstring": "^4.0.1"
   },
   "devDependencies": {
+    "@types/jws": "^3.1.0",
```
ofrobots: qq: I don't see an import of `@types/axios`. Intentional?

src/auth/authclient.ts
```diff
@@ -28,5 +34,5 @@ export abstract class AuthClient {
    * Provides an alternative request
    * implementations with auth credentials.
    */
-  abstract request(opts: request.Options): request.Request|void;
+  abstract request(opts: AxiosRequestConfig): Promise<BodyResponse>;
```
ofrobots: This is a breaking change, but keeping the function name the same would create a landmine for users. Some alternatives:

* Does this need to be public API at all?
* Could we provide an API compatible implementation that provides `request.Result` on top of axios (i.e. avoid this being a semver major).
* If semver major is unavoidable and we need to expose this API, can we rename it to make the migration for dependents easier?

package.json
JustinBeckwith: Yup - they kindly package their types in the npm module :)

src/auth/authclient.ts
JustinBeckwith: I spot checked a few dependents, and it looks like folks aren't directly using this method (though it's always hard to say).   A few thoughts:
- We could try making this protected, and see what happens :)
- I don't think wrapping this with request style APIs would be successful.  There are way too many differences between Axios and Request, and I don't want to manage a janky compat shim layer if possible.
- Another option here is to check for common params that could come in with a request Options object that aren't present on an Axios config, and throw if I see them with a humane warning.

I would prefer either making this protected, and/or more humane defensive parameter checking with appropriate exceptions. 

Eh?

package.json
ofrobots: I am curious how.. I don't see a `types` property in their [`package.json`](https://github.com/axios/axios/blob/master/package.json)

jinwoo: It has `typings` and `types` and `typings` are synonymous: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

src/transporters.ts
```diff
@@ -74,11 +65,25 @@ export class DefaultTransporter {
    * @param {Function=} callback Optional callback.
    * @return {Request} Request object
    */
-  request(opts: request.Options, callback?: BodyResponseCallback) {
+  request(opts: AxiosRequestConfig): AxiosPromise;
+  request(opts: AxiosRequestConfig, callback?: BodyResponseCallback): void;
+  request(opts: AxiosRequestConfig, callback?: BodyResponseCallback):
```
ofrobots: Can this be an async function?

src/auth/authclient.ts
```diff
@@ -14,12 +14,18 @@
...
-import * as request from 'request';
+import axios, {AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios';
 
 import {DefaultTransporter} from '../transporters';
 
 import {Credentials} from './credentials';
 
+export interface BodyResponse {
+  // tslint:disable-next-line no-any
+  body: any;
```
ofrobots: Is this the same value as `res.data`? In that case let the users dig it out. You're making a semver major change regardless, and abbreviating `res.data` to `body` isn't that big a win. That gets rid of this whole interface.

Secondly, instead of using `any`, you can use generics similarly to how `axios` itself is [doing](https://github.com/axios/axios/blob/d7f021b8d4cc50bfa0653011bc02452d234d1255/index.d.ts#L51).

ofrobots: Okay, so I think I understand your motivation. This way, the new API continues working without requiring changes from users, *in most cases*. The tests themselves are easy to migrate.

This is very dangerous. Unless you are 100% compatible you should not pretend to be compatible. The users need to migrate; by making their code work in most cases will set them up for major sadness in the obscure cases where it happens to not work.

If this is not a duck, it should not walk like a duck, nor talk like a duck.

ofrobots: I don't think this particular API is in heavy use by users in any case, so we should not try to optimize for 'ease for migration'.

src/transporters.ts
```diff
@@ -87,48 +92,28 @@ export class DefaultTransporter {
...
-        return callback && callback(err, body, res);
-      }
-      // Only and only application/json responses should
-      // be decoded back to JSON, but there are cases API back-ends
-      // responds without proper content-type.
-      try {
-        body = JSON.parse(body as string);
-      } catch (err) {
-        /* no op */
+  private processError(e: AxiosError): RequestError {
```
ofrobots: jsdoc is no longer correct.

src/transporters.ts
JustinBeckwith: Sorta not-really yes.  `AxiosPromise` is a Promise object, so callers of the API can just `await` it like you would a normal async function.  The reason it's not marked as `async` is because you can't have some overloaded method signatures async, and others non-async.  I want to support the mostly backward compatibility here, so that's why I opted for just returning a Promise.  To end users - it should make no difference (I think).

src/auth/authclient.ts
JustinBeckwith: If we can switch out request for axios, AND add support to async APIs, all while maintaining 90% compatibility - I'd really like to do it.  For most users, they're not going to need to make any changes.  For the users that are really deeply using the API, the only real differences will be:
- Inputs to the various request methods (we talked about throwing when they look like non-axios supported objects)
- The response objects in the callback.  In most cases from what I can tell, people just grab the body and work with that.  In the event that they actually need the response object, they're going to figure out pretty quickly that it's different - all the property names are different, even statusCode.  

I know we're in the uncanny valley of compatibility, but I think we can *really* reduce the pain of this upgrade for > 90% of dependents.

ofrobots: > I know we're in the uncanny valley of compatibility, but I think we can really reduce the pain of this upgrade for > 90% of dependents.

You're changing behavior in a very subtle way. 90% of users will probably not see any pain, but the pain to the remaining 10% could be substantial. For example, `request` provides only network errors via `err`, whereas axios considers 5xx responses to be errors too.

The evidence that would be required to convince me is a sampling of dependent modules that demonstrates that a "no-thinking-required migration" in fact produces the behavior those modules expected.

JustinBeckwith: Nah, you convinced me :) I'll try it out.

JustinBeckwith:
@ofrobots I think I covered all the feedback so far.  Ready for the next wave :)

src/auth/googleauth.ts
```diff
@@ -108,9 +123,22 @@ export class GoogleAuth {
...
   /**
    * Obtains the default project ID for the application..
-   * @param {function=} callback Optional callback.
+   * @param callback Optional callback
+   * @returns Promise that resolves with project Id (if used without callback)
    */
-  getDefaultProjectId(callback: ProjectIdCallback) {
+  getDefaultProjectId(): Promise<string>;
+  getDefaultProjectId(callback: ProjectIdCallback): void;
+  getDefaultProjectId(callback?: ProjectIdCallback): Promise<string|null>|void {
```
ofrobots: nit: I'm glad that you're expanding the API here to better support promises, but in such large refactor PRs it is better to have a pure refactor CL that is easier to review; and follow-up with new functionality.

src/auth/googleauth.ts
```diff
@@ -123,182 +151,182 @@ export class GoogleAuth {
...
+    let projectId: string|null|undefined;
 
-      // json file
-      this.getFileProjectId((err?: Error|null, projectId?: string) => {
-        if (err || projectId) {
-          myCallback(err, projectId);
-          return;
-        }
+    // environment variable
+    projectId = this.getProductionProjectId();
```
ofrobots: This code here will be simplified a lot if you used the following pattern:

```ts
projectId = this.getProductionProjectId() || await this.getFileProjectId() | await this.getDefaultServiceProjectId() || ...
if (projectId)
```

src/auth/googleauth.ts
```diff
@@ -123,182 +151,182 @@ export class GoogleAuth {
...
   /**
    * Run the Google Cloud SDK command that prints the default project ID
-   * @param {function} callback Callback.
-   * @api private
    */
-  _getSDKDefaultProjectId(
-      callback:
-          (error: Error|null, stdout: string, stderr: string|null) => void) {
-    exec('gcloud -q config list core/project --format=json', callback);
+  _getSDKDefaultProjectId():
```
ofrobots: I think this is private, but expose for testing? If so could this be an async function so that you don't have to manual promising below (nit).

src/auth/computeclient.ts
```diff
@@ -62,64 +57,58 @@ export class Compute extends OAuth2Client {
...
-                ((new Date()).getTime() + (token.expires_in * 1000));
-            delete token.expires_in;
-          }
-          if (callback) {
-            callback(err, token, response);
-          }
-        });
+  protected async refreshToken(refreshToken?: string|
+                               null): Promise<GetTokenResponse> {
+    try {
```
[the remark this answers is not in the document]
JustinBeckwith: fixed

src/auth/computeclient.ts
```diff
@@ -62,64 +57,58 @@ export class Compute extends OAuth2Client {
...
+      const url = this.opts.tokenUrl || Compute._GOOGLE_OAUTH2_TOKEN_URL;
+      // request for new token
+      const res = await this.transporter.request({url});
```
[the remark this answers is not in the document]
JustinBeckwith: yes :)

src/auth/credentials.ts
```diff
@@ -19,6 +19,7 @@ export interface Credentials {
   expiry_date?: number;
   access_token?: string;
   token_type?: string;
+  expires_in?: number;
```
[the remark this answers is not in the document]
JustinBeckwith: Yeah, nice call.  I separated the types and did some gymnastics to get this right.

src/auth/googleauth.ts
JustinBeckwith: I'm usually better about this, apologies.  I tried to bite off too much with this one.

src/auth/googleauth.ts
JustinBeckwith: not sure what I was thinking there lol

src/auth/googleauth.ts
JustinBeckwith: It is private, but exposed for testing. If I make this an async, I need to promisify exec.  I think I avoided doing that so I wouldn't need to pull in pify just for this one case.  What do you think?  Is it worth adding another dependency to clean this up?

src/auth/googleauth.ts
```diff
@@ -453,15 +486,23 @@ export class GoogleAuth {
    * @param {function=} - Optional callback function
    */
   fromAPIKey(
-      apiKey: string, callback?: (err?: Error|null, client?: JWT) => void) {
+      apiKey: string,
```
ofrobots: Same. This is a sync function; it would be good to drop the callback parameter as part of the 1.0 milestone. Don't fix here, open a bug.

src/auth/jwtaccess.ts
```diff
@@ -57,112 +54,116 @@ export class JWTAccess {
...
+   * @param {function} metadataCb a callback invoked with the jwt request metadata.
+   * @returns a Promise that resolves with the request metadata response
    */
-  getRequestMetadata(
-      authURI: string,
-      metadataCb:
-          (err: Error|null, headers?: http.IncomingHttpHeaders|null) => void) {
-    const iat = Math.floor(new Date().getTime() / 1000);
-    const exp = iat + 3600;  // 3600 seconds = 1 hour
+  getRequestMetadata(authURI: string): RequestMetadataResponse;
```
ofrobots: Likewise. This method seems synchronous to me. Let's drop the callback version of the API. Open bug (or attach to the running bug for this).

src/auth/jwtaccess.ts
```diff
@@ -57,112 +54,116 @@ export class JWTAccess {
...
+  getRequestMetadata(authURI: string, callback: RequestMetadataCallback): void;
+  getRequestMetadata(authURI: string, callback?: RequestMetadataCallback):
+      void|RequestMetadataResponse {
+    try {
```
ofrobots: This try block is over broad. I think you only intent to catch errors during `jws.sign`, so target it to that specifically.

src/auth/jwtaccess.ts
```diff
@@ -57,112 +54,116 @@ export class JWTAccess {
...
-    if (!json.client_email) {
-      done(new Error(
-          'The incoming JSON object does not contain a client_email field'));
-      return;
-    }
-    if (!json.private_key) {
-      done(new Error(
-          'The incoming JSON object does not contain a private_key field'));
-      return;
+  fromJSON(json: JWTInput, callback?: (err: Error|null) => void) {
```
ofrobots: Same. This looks sync.

src/auth/jwtclient.ts
```diff
@@ -107,46 +103,53 @@ export class JWT extends OAuth2Client {
...
+    } else {
+      return this.authorizeAsync();
+    }
+  }
+
+  private async authorizeAsync(): Promise<Credentials> {
+    const result = await this.refreshToken();
+    if (!result) {
+      throw new Error('No result returned');
+    } else {
```
ofrobots: nit: drop the `else`

src/auth/jwtclient.ts
```diff
@@ -107,46 +103,53 @@ export class JWT extends OAuth2Client {
...
-      ignored: string|null,
-      callback?: (err: Error, credentials?: Credentials) => void) {
-    const done = callback || noop;
-    return this.createGToken((err, newGToken) => {
-      if (err) {
-        return done(err);
-      } else {
-        return newGToken.getToken((err2?: Error|null, token?: string|null) => {
-          return done(err2, {
+  async refreshToken(refreshToken?: string|null) {
```
ofrobots: nit: drop the `async` unless you have `await`.

src/auth/jwtclient.ts
```diff
@@ -107,46 +103,53 @@ export class JWT extends OAuth2Client {
...
+          return reject(err);
+        } else {
+          const creds = {
             access_token: token,
             token_type: 'Bearer',
             expiry_date: newGToken.expiresAt
-          });
-        });
-      }
+          } as Credentials;
```
ofrobots: nit: can the cast be avoided?

src/auth/jwtclient.ts
```diff
@@ -155,57 +158,72 @@ export class JWT extends OAuth2Client {
...
-    if (!json.client_email) {
-      done(new Error(
-          'The incoming JSON object does not contain a client_email field'));
-      return;
-    }
-    if (!json.private_key) {
-      done(new Error(
-          'The incoming JSON object does not contain a private_key field'));
-      return;
+  fromJSON(json: JWTInput, callback?: (err?: Error) => void): void {
```
ofrobots: Same. This looks sync.

src/auth/jwtclient.ts
```diff
@@ -155,57 +158,72 @@ export class JWT extends OAuth2Client {
...
-    });
-    inputStream.on('end', () => {
-      try {
-        const data = JSON.parse(s);
-        this.fromJSON(data, callback);
-      } catch (err) {
-        done(err);
+  }
+
+  private async fromStreamAsync(inputStream: stream.Readable) {
```
ofrobots: nit: drop `async` unless you're awaiting.

src/auth/jwtclient.ts
```diff
@@ -215,37 +233,36 @@ export class JWT extends OAuth2Client {
    * @param {function=} callback - Optional callback to be invoked after
    *  initialization.
    */
-  fromAPIKey(apiKey: string, callback?: (err: Error) => void) {
-    const done = callback || noop;
+  fromAPIKey(apiKey: string, callback?: (err?: Error) => void): void {
```
ofrobots: Same. This looks sync.

src/auth/googleauth.ts
```diff
@@ -123,182 +151,182 @@ export class GoogleAuth {
...
-              'Could not load the default credentials. Browse to ' +
-              'https://developers.google.com/accounts/docs/application-default-credentials for ' +
-              'more information.'));
-        }
-      });
+      const gce = await this._checkIsGCE();
+      if (gce) {
+        // For GCE, just return a default ComputeClient. It will take care of
+        // the rest.
+        return {projectId: null, credential: new Compute()};
```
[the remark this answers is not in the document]
JustinBeckwith: added!

src/auth/googleauth.ts
```diff
@@ -324,126 +352,131 @@ export class GoogleAuth {
...
    * @param {object=} json The input object.
    * @param {function=} callback Optional callback.
+   * @returns Promise that resolves with the OAuth2Client (if no callback is
+   * passed)
    */
-  fromJSON(
-      json: JWTInput,
-      callback?: (err: Error|null, client?: UserRefreshClient|JWT) => void) {
+  fromJSON(json: JWTInput): JWT|UserRefreshClient;
+  fromJSON(json: JWTInput, callback: CredentialCallback): void;
```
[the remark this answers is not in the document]
JustinBeckwith: Created #188

src/auth/googleauth.ts
```diff
@@ -324,126 +352,131 @@ export class GoogleAuth {
...
+  fromJSON(json: JWTInput, callback?: CredentialCallback): JWT|UserRefreshClient
+      |void {
     let client: UserRefreshClient|JWT;
-    if (!json) {
+    try {
+      if (!json) {
```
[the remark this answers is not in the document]
JustinBeckwith: Agree that we can drop this if we make the API entirely sync.  There are a few other places where this happens, I can clean them up there too.

src/auth/jwtaccess.ts
JustinBeckwith: added to running issue

src/auth/googleauth.ts
JustinBeckwith: added to #188

src/auth/jwtaccess.ts
JustinBeckwith: Right now it's capturing all errors so they can be passed to the callback.  Will scope this down as part of #188

src/auth/jwtaccess.ts
JustinBeckwith: Added to #188

src/auth/jwtclient.ts
JustinBeckwith: fixed

src/auth/jwtclient.ts
JustinBeckwith: Forgot to update this function after updating gtoken.  Now using await.

src/auth/jwtclient.ts
JustinBeckwith: fixed

src/auth/jwtclient.ts
JustinBeckwith: in #188

src/auth/jwtclient.ts
JustinBeckwith: fixed

src/auth/jwtclient.ts
JustinBeckwith: In #188

src/auth/googleauth.ts
ofrobots: Add a TODO.

JustinBeckwith: Done

test/test.googleauth.ts
```diff
@@ -1029,642 +936,576 @@ describe('GoogleAuth', () => {
...
+    // back, but the execution count has not rev'd again, indicating that we
+    // got the cached values this time.
+    const isGCE2 = await auth._checkIsGCE();
+    assert.equal(true, auth.isGCE);
+    assert.equal(1, (auth.transporter as MockTransporter).executionCount);
+  });
 
-    it('Does not execute the second time when not running on GCE', (done) => {
-      const auth = new GoogleAuth();
+  it('Does not execute the second time when not running on GCE', async () => {
```
ofrobots: TIL that does mocha checks the return value to see if it is a promise.

test/test.googleauth.ts
```diff
@@ -1029,642 +936,576 @@ describe('GoogleAuth', () => {
...
+      createIsGCENock();
+      await auth._checkIsGCE();
+      // Assert that the flags are set.
+      assert.equal(true, auth.isGCE);
+
+      nock('http://metadata.google.internal')
+          .get('/computeMetadata/v1/instance/service-accounts/?recursive=true')
+          .reply(404);
+
+      try {
```
ofrobots: assert.throws

test/test.googleauth.ts
```diff
@@ -1029,642 +936,576 @@ describe('GoogleAuth', () => {
...
-          auth._tryGetApplicationCredentialsFromEnvironmentVariable();
-      assert.equal(false, credentialFlag);
-      auth.getCredentials((_err, body) => {
-        assert.equal(true, _err instanceof Error);
-        done();
-      });
+      const client =
+          await auth._tryGetApplicationCredentialsFromEnvironmentVariable();
+      assert.equal(null, client);
+      try {
```
ofrobots: assert.throws

test/test.jwt.ts
```diff
@@ -643,7 +612,7 @@ describe('.fromAPIKey', () => {
     it('should set the .apiKey property on the instance', (done) => {
       jwt.fromAPIKey(KEY, (err) => {
         assert.strictEqual(jwt.apiKey, KEY);
-        assert.strictEqual(err, null);
+        assert.equal(err, null);
```
ofrobots: why was this change needed?

test/test.oauth2.ts
```diff
@@ -855,7 +854,7 @@ describe('OAuth2 client', () => {
     const auth = new GoogleAuth();
     const oauth2client =
         new auth.OAuth2(CLIENT_ID, CLIENT_SECRET, REDIRECT_URI);
-    oauth2client.request(({} as request.OptionsWithUri), (err, result) => {
+    oauth2client.request(({} as AxiosRequestConfig), (err, result) => {
```
ofrobots: All properties in `AxiosRequestConfig` are optional, so the cast should not be needed.

test/test.oauth2.ts
```diff
@@ -996,7 +995,10 @@ describe('OAuth2 client', () => {
       oauth2client.credentials = {access_token: 'abc', refresh_token: 'abc'};
       oauth2client.revokeCredentials((err, result) => {
         assert.equal(err, null);
-        assert.equal(result.success, true);
+        assert(result);
```
ofrobots:
```ts
assert(result && result.data);
assert.equal(result!.data!.success, true);
```

test/test.oauth2.ts
```diff
@@ -1034,8 +1036,11 @@ describe('OAuth2 client', () => {
       const oauth2client =
           new auth.OAuth2(CLIENT_ID, CLIENT_SECRET, REDIRECT_URI);
       oauth2client.getToken('code here', (err, tokens) => {
-        assert(tokens.expiry_date >= now + (10 * 1000));
-        assert(tokens.expiry_date <= now + (15 * 1000));
+        assert(tokens);
+        if (tokens && tokens.expiry_date) {
```
ofrobots: likewise.

test/test.transporters.ts
```diff
@@ -25,12 +26,11 @@ const version = require('../../package.json').version;
 nock.disableNetConnect();
 
 describe('Transporters', () => {
-
   const defaultUserAgentRE = 'google-api-nodejs-client/\\d+.\\d+.\\d+';
   const transporter = new DefaultTransporter();
 
   it('should set default client user agent if none is set', () => {
-    const opts = transporter.configure(({} as request.OptionsWithUrl));
+    const opts = transporter.configure(({} as AxiosRequestConfig));
```
ofrobots: cast should be unnecessary

src/auth/oauth2client.ts
```diff
@@ -176,30 +234,26 @@ export class OAuth2Client extends AuthClient {
...
-                ((new Date()).getTime() + (tokens.expires_in * 1000));
-            delete tokens.expires_in;
-          }
-          const done = callback || noop;
-          done(err, tokens, response);
-        });
+    const res = await this.transporter.request<CredentialRequest>(
+        {method: 'POST', url, data});
+    const tokens = res.data as Credentials;
+    if (res.data && res.data.expires_in) {
```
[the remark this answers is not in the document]
JustinBeckwith: done

src/auth/oauth2client.ts
```diff
@@ -208,65 +262,72 @@ export class OAuth2Client extends AuthClient {
...
-        return callback(null, tokens.access_token, response);
-      });
+      const r = await this.refreshAccessToken();
+      if (!r.credentials || (r.credentials && !r.credentials.access_token)) {
+        throw new Error('Could not refresh access token.');
+      }
+      return {token: r.credentials.access_token, res: r.res};
     } else {
-      return callback(null, this.credentials.access_token, null);
+      return {token: this.credentials.access_token, res: undefined};
```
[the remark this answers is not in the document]

src/auth/oauth2client.ts
```diff
@@ -307,64 +376,89 @@ export class OAuth2Client extends AuthClient {
...
+    }
+
+    const credentials = this.credentials;
+    credentials.token_type = credentials.token_type || 'Bearer';
+    tokens.refresh_token = credentials.refresh_token;
+    this.credentials = tokens;
+    const headers = {
+      Authorization: credentials.token_type + ' ' + tokens.access_token
+    };
+    return {headers, res: r ? r.res : null};
```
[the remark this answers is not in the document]
JustinBeckwith: How did I not know that was possible in TypeScript....  fixed

src/auth/refreshclient.ts
```diff
@@ -58,45 +55,38 @@ export class UserRefreshClient extends OAuth2Client {
    * @param {function=} callback Optional callback.
    */
   fromJSON(json: JWTInput, callback?: (err?: Error) => void) {
-    if (!json) {
-      if (callback) {
-        callback(new Error(
-            'Must pass in a JSON object containing the user refresh token'));
+    try {
```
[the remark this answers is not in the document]
JustinBeckwith: added to #188

src/options.ts
```diff
@@ -0,0 +1,33 @@
+import {AxiosRequestConfig} from 'axios';
```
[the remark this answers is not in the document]
JustinBeckwith: Hah, vscode auto imports things for you sometimes.  This shouldn't be here at all. Fixed

src/options.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// tslint:disable-next-line no-any
+export function validate(options: any) {
```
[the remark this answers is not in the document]
JustinBeckwith: fixed

src/transporters.ts
```diff
@@ -51,12 +44,12 @@ export class DefaultTransporter {
...
-   * @param {object} opts Options to configure.
-   * @return {object} Configured options.
+   * @param opts AxiosRequestConfig options.
+   * @return Configured options.
    */
-  configure(opts: request.Options): request.Options {
+  configure(opts: AxiosRequestConfig = {}): AxiosRequestConfig {
     // set transporter user agent
-    opts.headers = opts.headers || {};
+    if (!opts.headers) opts.headers = {};
```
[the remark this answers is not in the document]
JustinBeckwith: No reason *shrugs* I put it back

src/transporters.ts
```diff
@@ -69,66 +62,71 @@ export class DefaultTransporter {
...
+      } else {
+        throw e;
+      }
+    }
+
+    if (callback) {
+      axios(opts)
+          .then(r => {
+            callback(null, r);
+            return;
```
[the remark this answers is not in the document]
JustinBeckwith: fixed

test/test.googleauth.ts
```diff
@@ -15,20 +15,36 @@
...
+});
+
+function createIsGCENock(isGCE = true) {
+  nock('http://metadata.google.internal').get('/').reply(200, null, {
+    'metadata-flavor': 'Google'
+  });
+}
+
+function createGetProjectIdNock(projectId: string) {
+  nock('http://169.254.169.254')
```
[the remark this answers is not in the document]
JustinBeckwith: I opened #190 to track.  There was some interesting context in the code on why it's using the IP:

` Uses 169.254.169.254 for the metadata server to avoid request latency from DNS lookup. See https://cloud.google.com/compute/docs/metadata#metadataserver for information about this IP address. (This IP is also used for Amazon EC2 instances, so the metadata flavor is crucial.)`

test/test.googleauth.ts
```diff
@@ -485,225 +485,197 @@ describe('GoogleAuth', () => {
...
 
-    it('should error on invalid symlink', (done) => {
+    it('should error on invalid symlink', async () => {
       const auth = new GoogleAuth();
-      auth._getApplicationCredentialsFromFilePath(
-          './test/fixtures/badlink', (err) => {
-            assert.equal(true, err instanceof Error);
-            done();
-          });
+      try {
```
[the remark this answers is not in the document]
JustinBeckwith: It doesn't work with async calls :(  You always get a ` AssertionError [ERR_ASSERTION]: Missing expected exception.`.  I could use something like this if you think it's better:
https://github.com/akdor1154/node-assert-throws-async

test/test.googleauth.ts
```diff
@@ -485,225 +485,197 @@ describe('GoogleAuth', () => {
...
+      }
+      assert.fail('failed to throw');
     });
 
-    it('should handle errors thrown from createReadStream', (done) => {
+    it('should handle errors thrown from createReadStream', async () => {
       // Set up a mock to throw from the createReadStream method.
       const auth = new GoogleAuth();
       auth._createReadStream = () => {
         throw new Error('Hans and Chewbacca');
```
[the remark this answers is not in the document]
JustinBeckwith: Ew, you're right.  Fixed.

test/test.googleauth.ts
```diff
@@ -776,147 +740,117 @@ describe('GoogleAuth', () => {
...
-            assert.equal('hello', (result as JWT).email);
-          }
-          step();
-        });
-    assert.equal(true, handled);
-    step();
+    const result = await auth._tryGetApplicationCredentialsFromWellKnownFile();
+    assert(result);
+    if (result) {
+      assert.equal('hello', (result as JWT).email);
```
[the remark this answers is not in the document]
JustinBeckwith: No, but that does help get rid of the `if (result)` part.  The issue here is that the method returns `JWT|UserRefreshClient`, and `UserRefreshClient` doesn't have an `email` property.

test/test.jwt.ts
JustinBeckwith: In the `fromAPIKey` callback I had it set up so that the callback could just be invoked with no parameters, making `err` undefined instead of null.  Not sure what the best practice is here, or if it matters.  I could make err `err: Error|null` instead of `err?: Error` if you think that's better :)

test/test.oauth2.ts
JustinBeckwith: fixed here and a few other places

test/test.oauth2.ts
JustinBeckwith: eugh, that's so much nicer

test/test.oauth2.ts
JustinBeckwith: Fixed here and many other places

test/test.transporters.ts
JustinBeckwith: fixed

test/test.googleauth.ts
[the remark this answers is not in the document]
ofrobots: I know about that one. We are not consistent internally though.

I am doubtful that the optimization was worthwhile though; and I suspect it might be environment specific (k8s, is my guess). A better fix would have been to try resolving `metadata.google.internal.` with the trailing dot. Regardless, let's address it in #190.

test/test.googleauth.ts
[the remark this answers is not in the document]
ofrobots: Ack. This repeated enough times that it would be worth writing a wrapper for this at the least.

PR #183, opened 2017-11-20
chore: upgrade to typed versions of jws and gtoken

JustinBeckwith:


src/auth/jwtclient.ts
```diff
@@ -14,48 +14,24 @@
...
-}
-
 export class JWT extends OAuth2Client {
   email?: string;
   keyFile?: string|null;
   key?: string|null;
   scopes?: string|string[]|null;
   scope?: string|null;
   subject?: string;
-  gToken: GoogleToken;
```
ofrobots: semver major?

test/test.jwt.ts
```diff
@@ -319,7 +286,6 @@ describe('JWT auth client', () => {
       jwt.request({uri: 'http://bar'}, () => {
         assert.equal('initial-access-token', jwt.credentials.access_token);
         assert.equal(false, scope.isDone());
-        nock.cleanAll();
```
ofrobots: thanks for these

src/auth/jwtclient.ts
JustinBeckwith: So this was super weird.  It was exposing a reference to the GoogleToken.getToken function, which was only public so it could be mocked in tests.  While technically yes this is a breaking change, I can't imagine anyone is using it this way.  

That having been said - yes, lets lump this in with the other breaking changes we have for the 1.0 release.

test/test.jwt.ts
JustinBeckwith: 👍  self preservation.  If I don't do it in a afterEach, tests fail in super confusing ways.

Issue #170, opened 2017-10-18
JWT client callback function has no error and no tokens


bantini:
Originally raised by @0xc00010ff in the nodejs library. [#812](https://github.com/google/google-api-nodejs-client/issues/812)

The JWT authorization API callback does not return any tokens as suggested by the docs. Rather, it just updated the google.auth.JWT client object to have the access token _somewhere in there_.

My suggestion is either to send back a valid `tokens` parameter or remove it all together and publish the requirement to reach into the jwtClient and grab the bearer token. The `token` param in the docs and example is a big red herring. 

```javascript
var key = require('/path/to/key.json');
var jwtClient = new google.auth.JWT(
  key.client_email,
  null,
  key.private_key,
  ['https://www.googleapis.com/auth/plus.me', 'https://www.googleapis.com/auth/calendar'], // an array of auth scopes
  null
);

jwtClient.authorize(function (err, tokens) {

  console.log(err, tokens); // undefined, null
  console.log(jwtClient) // [object Object] with bearer token snarled in the object

  if (err) {
    console.log(err);
    return;
  }

  token = tokens[0]; // null!

  // Make a request to the Play Store to validate a receipt. Use the JWT token
  Request(`https://play.store.url/with/product/id?access_token=${token}`)
});
```

JustinBeckwith:
I can't seem to reproduce this one.  I added a few asserts in the existing test in #195 just to make sure... but it doesn't seem to be an issue.  If someone can share a solid repo, please let me know!

JustinBeckwith:
If someone can share a repro here, feel free to re-open.

PR #195, opened 2017-11-26
test: add asserts on authorize

JustinBeckwith:
Just a few more asserts to validate that #170 doesn't pop up.

### DIFF
--- a/test/test.jwt.ts
+++ b/test/test.jwt.ts
@@ -81,12 +81,17 @@ describe('JWT auth client', () => {
           ['http://bar', 'http://foo'], 'bar@subjectaccount.com');
 
       createGTokenMock();
-      jwt.authorize(() => {
+      jwt.authorize((err, creds) => {
+        assert.equal(err, null);
+        assert.notEqual(creds, null);
         assert.equal('foo@serviceaccount.com', jwt.gtoken.iss);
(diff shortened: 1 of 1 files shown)


test/test.jwt.ts
```diff
@@ -81,12 +81,17 @@ describe('JWT auth client', () => {
...
-      jwt.authorize(() => {
+      jwt.authorize((err, creds) => {
+        assert.equal(err, null);
+        assert.notEqual(creds, null);
         assert.equal('foo@serviceaccount.com', jwt.gtoken.iss);
         assert.equal(PEM_PATH, jwt.gtoken.keyFile);
         assert.equal(['http://bar', 'http://foo'].join(' '), jwt.gtoken.scope);
         assert.equal('bar@subjectaccount.com', jwt.gtoken.sub);
         assert.equal('initial-access-token', jwt.credentials.access_token);
+        assert.equal(creds!.access_token, jwt.credentials.access_token);
```
kjin: Wow!! I've never seen this before

JustinBeckwith: I hadn't either!  @ofrobots to the rescue.

JustinBeckwith: I mean csharp has it like `creds?.access_token` so I guess I shouldn't be surprised, but still nice.

PR #199, opened 2017-11-29
Cherry pick changes from master into next

JustinBeckwith:


src/auth/oauth2client.ts
```diff
@@ -17,18 +17,20 @@
...
 import * as http from 'http';
 import * as querystring from 'querystring';
-
 import {PemVerifier} from './../pemverifier';
 import {BodyResponseCallback} from './../transporters';
 import {AuthClient} from './authclient';
 import {CredentialRequest, Credentials} from './credentials';
 import {LoginTicket} from './loginticket';
 
+const isString = require('lodash.isstring');
```
gaofanmichael: Can we use import signature on this? `@sindresorhus/is` can be an alternative as well.

ofrobots: Let's open this as a bug. A cherry-pick between branches shouldn't edit.

ofrobots:
@JustinBeckwith an alternative would be to rebase the next branch onto master. WDYT? If that sounds reasonable, go ahead and do that, no need to PR.

src/auth/oauth2client.ts
JustinBeckwith: FWIW I did try to do that :) Couldn't get it to work and decided to come back later.

ofrobots:
I have a rebased `n2` branch on my fork. I can replace the `next` branch once the pending PRs land. Closing this PR.

Issue #196, opened 2017-11-28
Update docs in the next branch


JustinBeckwith:
The docs are currently incomplete, and the ones in the `next` branch are just wrong.  Lets update these before the 1.0 release.

JustinBeckwith:
Fixed in the next branch by #201

PR #201, opened 2017-12-01
docs: add docs and samples

JustinBeckwith:
This includes a few things:
- Adds a few examples of using the API
- Improves the README
- ~~Update to a working version of gtoken~~ (handled in #203) 
- ~~Fixes a few bugs with Content-Type~~ (handled in #203) 
- ~~Exposes objects as exports instead of relying on public properties on the GoogleAuth object~~ (handled in #202)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,12 +11,14 @@
 This is Google's officially supported [node.js][node] client library for using OAuth 2.0 authorization and authentication with Google APIs.
 
 ## Installation
-
 This library is distributed on `npm`. To add it as a dependency, run the following command:
 
 ``` sh
 $ npm install google-auth-library
 ```
+
+## Upgrading to 1.x
+The `1.x` release includes a variety of bug fixes, new features, and breaking changes. Please take care, and see [the release notes](https://github.com/google/google-auth-library-nodejs/releases/tag/1.0.0-alpha.1) for a list of breaking changes, and the upgrade guide.
 
 ## Application Default Credentials
 This library provides an implementation of [Application Default Credentials][] for Node.js. The [Application Default Credentials][] provide a simple way to get authorization credentials for use in calling Google APIs.
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ async function main() {
  * Create a new OAuth2Client, and go through the OAuth2 content
  * workflow.  Return the full client to the callback.
  */
-async function getAuthenticatedClient() {
+function getAuthenticatedClient() {
   return new Promise((resolve, reject) => {
     // create an oAuth client to authorize the API call.  Secrets are kept in a `keys.json` file,
     // which should be downloaded from the Google Developers Console.
--- a/README.md
+++ b/README.md
@@ -1,102 +1,247 @@
-# Google APIs Node.js Client
+<img src="https://avatars0.githubusercontent.com/u/1342004?v=3&s=96" alt="Google Inc. logo" title="Google" align="right" height="96" width="96"/>
 
+# Google Auth Library
+
+[![npm version][npmimg]][npm]
 [![Build Status][travisimg]][travis]
 [![Code Coverage][coverallsimg]][coveralls]
 [![Dependencies][david-dm-img]][david-dm]
+[![Known Vulnerabilities][snyk-image]][snyk-url]
 
-This is Google's officially supported [node.js][node] client library for using
-OAuth 2.0 authorization and authentication with Google APIs.
+This is Google's officially supported [node.js][node] client library for using OAuth 2.0 authorization and authentication with Google APIs.
+
+## Installation
+
+This library is distributed on `npm`. To add it as a dependency, run the following command:
+
+``` sh
+$ npm install google-auth-library
+```
+
+## Application Default Credentials
+This library provides an implementation of [Application Default Credentials][] for Node.js. The [Application Default Credentials][] provide a simple way to get authorization credentials for use in calling Google APIs.
+
+They are best suited for cases when the call needs to have the same identity and authorization level for the application independent of the user. This is the recommended approach to authorize calls to Cloud APIs, particularly when you're building an application that uses Google Cloud Platform.
+
+#### Download your Service Account Credentials JSON file
+
+To use `Application Default Credentials`, You first need to download a set of JSON credentials for your project. Go to **APIs & Auth** > **Credentials** in the [Google Developers Console][devconsole] and select **Service account** from the **Add credentials** dropdown.
+
+> This file is your *only copy* of these credentials. It should never be
+> committed with your source code, and should be stored securely.
+
+Once downloaded, store the path to this file in the `GOOGLE_APPLICATION_CREDENTIALS` environment variable.
+
+#### Enable the API you want to use
+
+Before making your API call, you must be sure the API you're calling has been enabled. Go to **APIs & Auth** > **APIs** in the [Google Developers Console][devconsole] and enable the APIs you'd like to call. For the example below, you must enable the `DNS API`.
+
+
+#### Choosing the correct credential type automatically
+
+Rather than manually creating an OAuth2 client, JWT client, or Compute client, the auth library can create the correct credential type for you, depending upon the environment your code is running under.
+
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -1,135 +1,285 @@
...
+    console.error(e);
   }
+  process.exit();
+}
+
+/**
+ * Create a new OAuth2Client, and go through the OAuth2 content
+ * workflow.  Return the full client to the callback.
+ */
+async function getAuthenticatedClient() {
```
ofrobots: style nit: use `async` only when you need to use an `await`.

examples/adc.js
```diff
@@ -0,0 +1,69 @@
...
+    const scopes = ['https://www.googleapis.com/auth/cloud-platform'];
+    client = client.createScoped(scopes);
+  }
+  return {
+    client: client,
+    projectId: res.projectId
+  }
+}
+
+main();
```
ofrobots: nit

examples/jwt.js
```diff
@@ -0,0 +1,51 @@
...
+    keys.client_email,
+    null,
+    keys.private_key,
+    ['https://www.googleapis.com/auth/cloud-platform'],
+  );
+  await client.authorize();
+  return client;
+}
+
+main();
```
ofrobots: nit

examples/oauth2.js
```diff
@@ -0,0 +1,86 @@
...
+        resolve(oAuth2Client);
+      }
+    }).listen(3000, () => {
+      // open the browser to the authorize url to start the workflow
+      opn(authorizeUrl);
+    });
+  });
+}
+
+main();
```
ofrobots: nit

README.md
JustinBeckwith: It's using await in that method to call `getToken()`

README.md
```diff
@@ -1,135 +1,285 @@
...
-
-The [Application Default Credentials][] provide a simple way to get authorization credentials for use
-in calling Google APIs.
+This library provides an implementation of [Application Default Credentials][] for Node.js. The [Application Default Credentials][] provide a simple way to get authorization credentials for use in calling Google APIs.
 
-They are best suited for cases when the call needs to have the same identity and authorization
-level for the application independent of the user. This is the recommended approach to authorize
-calls to Cloud APIs, particularly when you're building an application that uses Google Compute
-Engine.
+They are best suited for cases when the call needs to have the same identity and authorization level for the application independent of the user. This is the recommended approach to authorize calls to Cloud APIs, particularly when you're building an application that uses Google Compute Engine.
```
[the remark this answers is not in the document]
JustinBeckwith: fixed

README.md
ofrobots: The function that calls `getToken` is already async. This one doesn't need to be.

examples/adc.js
JustinBeckwith: fixed

examples/jwt.js
JustinBeckwith: fixed

examples/oauth2.js
JustinBeckwith: fixed

README.md
```diff
@@ -1,135 +1,285 @@
...
 
-authFactory.getApplicationDefault(function(err, authClient) {
-  if (err) {
-    console.log('Authentication failed because of ', err);
-    return;
+  // The createScopedRequired method returns true when running on GAE or a local developer
+  // machine. In that case, the desired scopes must be passed in manually. When the code is
+  // running in GCE or a Managed VM, the scopes are pulled from the GCE metadata server.
+  // See https://cloud.google.com/compute/docs/authentication for more information.
+  if (client.createScopedRequired && client.createScopedRequired()) {
```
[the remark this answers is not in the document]
JustinBeckwith: It would probably clean things up a lot.  I'll do this in another PR.

JustinBeckwith:
@ofrobots yeah, I kind of started this PR without intending to make those changes, but it just kinda happened.  I split the origin contents of this one out into #202 (which landed) and #203 (which could use a review).  After that lands, I'll rebase this to be true docs/samples only.

JustinBeckwith:
Also, addresses #196

README.md
JustinBeckwith: Ah, good point.  Updated!

README.md
```diff
@@ -1,135 +1,287 @@
...
+const auth = new GoogleAuth();
+
+/**
+ * Acquire a client, and make a request to an API that's enabled by default.
+ */
+async function main() {
+  try {
+    const adc = await getADC();
+    const url = `https://www.googleapis.com/dns/v1/projects/${adc.projectId}`;
+    const res = await adc.client.request({url});
```
ofrobots: I don't have strong feelings here, but it would be preferable if we downplayed `client.request` in the examples. We really should not be encouraging the use of the auth library as a http client library. Ideally users use their own `request` or other client library.

This is definitely the most succinct example you can show and it seems convenient, but I don't think it scales when the needs of the user grow beyond what `client.request` can offer (e.g. multipart support). I've seen this questions about this show up many times.

A compromise might be if we added a section on how to use `authorizeRequest` with a user provided http client library.

README.md
```diff
@@ -1,135 +1,287 @@
...
+}
+
+main();
+```
+
+#### OAuth2 client
+
+This client comes with an [OAuth2][oauth] client that allows you to retrieve an access token and refreshes the token and retry the request seamlessly if you also provide an `expiry_date` and the token is expired. The basics of Google's OAuth2 implementation is explained on [Google Authorization and Authentication documentation][authdocs].
+
+In the following examples, you may need a `CLIENT_ID`, `CLIENT_SECRET` and `REDIRECT_URL`. You can find these pieces of information by going to the [Developer Console][devconsole], clicking your project --> APIs & auth --> credentials.
```
ofrobots: teeny nit: Higher up in the doc we use `>` arrows to show the navigation path. Here we use `-->`.

README.md
```diff
@@ -1,135 +1,287 @@
...
 #### Enable the API you want to use
 
-Before making your API call, you must be sure the API you're calling has been
-enabled. Go to **APIs & Auth** > **APIs** in the
-[Google Developers Console][devconsole] and enable the APIs you'd like to
-call. For the example below, you must enable the `DNS API`.
+Before making your API call, you must be sure the API you're calling has been enabled. Go to **APIs & Auth** > **APIs** in the [Google Developers Console][devconsole] and enable the APIs you'd like to call. For the example below, you must enable the `DNS API`.
+
+
+#### Choosing the correct credential type automatically
```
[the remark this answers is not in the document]
JustinBeckwith: I added a mini TOC at the top here that outlines the options, with links to the various sections.  Let me know what you think.

README.md
JustinBeckwith: As it stands today, this is the most common usage of the API, isn't it?  We specifically use this in the Google API Node.js Client.  I mean I hear what you're saying, but it just doesn't reflect the reality of the API as it's structured and used today.  

Lets open another issue for a future version (v2?) that cleans this up to avoid bundling the request client lib with the package. Thoughts?

README.md
JustinBeckwith: fixed

README.md
ofrobots: I know that the Google API Node.js client does this, but I wasn't able to find other examples. Is it really the most common usage?

JustinBeckwith:
Note:  test failures are unrelated, and fixed in #209

README.md
JustinBeckwith: Yeah, I did some spot checking and this stuff was all over the map.  For something like ADC - I'm not even sure how to show someone to use that without using the request object *scratches head*.  Maybe for the individual clients (JWT, OAuth2, Compute) it may make sense to show different examples.  To be honest - I'm very open to adding more samples and tweaking this stuff, but I do feel like this PR is burning a hole in my pocket :)  I'd like to get this branch merged into master soonish.  What do you think?

ofrobots: Since I already indicated that I don't have strong feelings, I am not going to block on this.

PR #202, opened 2017-12-01
BREAKING CHANGE: export class types and remove properties from GoogleAuth

JustinBeckwith:
When writing samples, I noticed that today I need to instantiate a new `GoogleAuth`, then from that instance you get access to the types needed to create new `JWT`, `UserRefreshClient`, and `OAuth2Client` objects.  That's silly.  Instead, I just added them to the exports on the index.


JustinBeckwith:
Good feedback :) Addressed all the things!

PR #203, opened 2017-12-01
fix: use urlencoded content-type for POST

JustinBeckwith:
Includes an update to gtoken, where the same bug was fixed.

### DIFF
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -214,8 +214,12 @@ export class OAuth2Client extends AuthClient {
       grant_type: 'authorization_code'
     };
 
-    const res = await this.transporter.request<CredentialRequest>(
-        {method: 'POST', url, data: values});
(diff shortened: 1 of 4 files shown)


src/auth/oauth2client.ts
```diff
@@ -214,8 +214,12 @@ export class OAuth2Client extends AuthClient {
       grant_type: 'authorization_code'
     };
 
-    const res = await this.transporter.request<CredentialRequest>(
-        {method: 'POST', url, data: values});
+    const res = await this.transporter.request<CredentialRequest>({
+      method: 'POST',
+      url,
+      data: querystring.stringify(values),
+      headers: {'Content-Type': 'application/x-www-form-urlencoded'}
```
ofrobots: Should there be a test modification that expects this header to be present?

JustinBeckwith: Yes :) I just included a change to the nock config in the test that will now make sure the content type is properly set.

Issue #133, opened 2017-04-13
Larger Safety Windows around Expiration Timestamp?


stephenplusplus:
_From @MarkHerhold on March 27, 2017 0:7_

Insert calls to BigQuery randomly result in auth errors.

I have a job that is responsible for sending data in bulk to BigQuery by using `.insert([ 'lots', 'of', 'items' ])` repeatedly in synchronous fashion. This process goes on for hours at a time, calling `insert` after the prior operation finishes.

After a few hours, my calls to BigQuery abruptly fail. I'm not sure if this issue is time-related or the call itself randomly fails. It does seem to occur after a few hours. This happens repeatedly.

**Error:**
```
Unhandled promise rejection (rejection id: 1): ApiError: Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.
```  

#### Environment details

  - OS: OSX El Capitan
  - Node.js version: 7.7.3
  - npm version: 4.1.2
  - `@google-cloud/bigquery` version: 0.8.0

#### Steps to reproduce

The above description tells it all, but I'm working on a script. I'll update this if I can find a way to reproduce in a more predictable manner.

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2139_

stephenplusplus:
_From @lukesneeringer on March 27, 2017 18:57_

@stephenplusplus This smells like a rate limit to me, other than the error message. Might this be a case where we are clobbering an error?

stephenplusplus:
I don't think we're clobbering it, just judging by the error message not being one that we've written. Before every request, we do a token grab, refreshing if necessary (that work is done by google-auth-library).

With this large amount of requests, maybe we need a larger safety window around the expiration timestamp for when the auth client makes the grab. @jmdobry does that sound possible? (please re-tag if someone else is closer to the auth library code)

stephenplusplus:
@ofrobots wdyt?

stephenplusplus:
Friendly re-ping.

merlinnot:
I came across the same issue. It's not a big deal for me because of a fault-tolerant design of my app, but I think it should have very high priority.

simonprickett:
[...]

PR #208, opened 2017-12-15

nolanmar511:
This can be used to address the issues described in #133, by allowing users to specify that tokens should be refreshed if they will expire soon. This is also necessary to fix https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/issues/94


ts/lib/auth/computeclient.ts
```diff
@@ -39,9 +39,11 @@ export class Compute extends OAuth2Client {
    *
    * Retrieve access token from the metadata server.
    * See: https://developers.google.com/compute/docs/authentication
+   *
+   * @param {number=} refreshTokeEarlyMillis The token should be refreshed if it will expire within this many milliseconds.
    */
-  constructor() {
-    super();
+  constructor(refreshTokenEarlyMillis?: number) {
+    super(undefined, undefined, undefined, undefined, refreshTokenEarlyMillis);
```
ofrobots: Oh my goodness. This doesn't have too great a readability.

I think the deficiency is in the constructor for `OAuth2Client`. Instead of taking so many optional properties as function parameters it should instead accept an options object. This would be a semver major change however, but perhaps it is worth making before 1.0? @JustinBeckwith WDYT?

ofrobots: actually there might be a way to avoid a semver major here. If arguments.length is 1 and it is an object that has one of the expected properties, we assume the 'new form'. Else we use the existing positional parameters form.

WDYT?

I still slightly favor a breaking change given the chance with 1.0.

nolanmar511:
If I run the the travis tests on my fork of the master branch, without modifying anything, the travis tests  fail in the same way.
The PRs which currently have passing tests seem to have different scripts in their package.json files. 

How should I address the failing tests?

ts/lib/auth/computeclient.ts
```diff
@@ -39,9 +39,11 @@ export class Compute extends OAuth2Client {
    *
    * Retrieve access token from the metadata server.
    * See: https://developers.google.com/compute/docs/authentication
+   *
+   * @param {number=} refreshTokeEarlyMillis The token should be refreshed if it will expire within this many milliseconds.
```
JustinBeckwith: Missing the `n` in `refreshTokenEarlyMillis`

ts/lib/auth/computeclient.ts
JustinBeckwith: I'm almost always in favor of accepting an options object over a mishmash of optional properties.  I would rather break it, and provide really nice error messages if we're going to do it.

ts/lib/auth/googleauth.ts
```diff
@@ -105,6 +105,11 @@ export class GoogleAuth {
    */
   public static DefaultTransporter = DefaultTransporter;
 
+  /**
+   * @param {number=} refreshTokeEarlyMillis The token should be refreshed if it will expire within this many milliseconds.
```
JustinBeckwith: Misspelled  `refreshTokenEarlyMillis` again :)

ts/lib/auth/googleauth.ts
```diff
@@ -400,9 +406,12 @@ export class GoogleAuth {
     // Set the JSON contents
     this.jsonContent = json;
     if (json.type === 'authorized_user') {
-      client = new UserRefreshClient();
+      client = new UserRefreshClient(
+          undefined, undefined, undefined, this.refreshTokenEarlyMillis);
```
JustinBeckwith: Similar to the discussion about switching to an options object in the oAuth2Client constructor, we should probably make that change here too.  WDYT @ofrobots ?

ts/lib/auth/googleauth.ts
```diff
@@ -105,6 +105,11 @@ export class GoogleAuth {
    */
   public static DefaultTransporter = DefaultTransporter;
 
+  /**
+   * @param {number=} refreshTokeEarlyMillis The token should be refreshed if it will expire within this many milliseconds.
+   */
+  constructor(private refreshTokenEarlyMillis?: number) {}
```
JustinBeckwith: Just curious - why is this private here, but public in the other classes?

ts/lib/auth/jwtclient.ts
```diff
@@ -44,12 +44,13 @@ export class JWT extends OAuth2Client {
    * @param {string=} key value of key
    * @param {(string|array)=} scopes list of requested scopes or a single scope.
    * @param {string=} subject impersonated account's email address.
+   * @param {number=} refreshTokeEarlyMillis The token should be refreshed if it will expire within this many milliseconds.
```
JustinBeckwith: Misspelled

ts/lib/auth/oauth2client.ts
```diff
@@ -52,17 +53,19 @@ export class OAuth2Client extends AuthClient {
    * @param {string=} clientSecret The authentication client secret.
    * @param {string=} redirectUri The URI to redirect to after completing the auth request.
    * @param {Object=} opt_opts optional options for overriding the given parameters.
+   * @param {number=} refreshTokeEarlyMillis The token should be refreshed if it will expire within this many milliseconds.
```
JustinBeckwith: Misspelled

ts/lib/auth/refreshclient.ts
```diff
@@ -29,10 +29,14 @@ export class UserRefreshClient extends OAuth2Client {
    * @param {string} clientId The authentication client ID.
    * @param {string} clientSecret The authentication client secret.
    * @param {string} refreshToken The authentication refresh token.
+   * @param {number=} refreshTokeEarlyMillis The token should be refreshed if it will expire within this many milliseconds.
```
JustinBeckwith: spelling

ts/test/test.oauth2.ts
```diff
@@ -931,6 +931,26 @@ describe('OAuth2 client', () => {
...
+         const oauth2client = new auth.OAuth2(
+             CLIENT_ID, CLIENT_SECRET, REDIRECT_URI, undefined, 5000);
+
+         oauth2client.credentials = {
+           access_token: 'initial-access-token',
+           refresh_token: 'refresh-token-placeholder',
+           expiry_date: (new Date()).getTime() + 3000
+         };
+
+         oauth2client.request({uri: 'http://example.com'}, () => {
```
JustinBeckwith: I hate to do this... but it probably makes more sense to rebase this PR against the `next` branch.

ts/lib/auth/computeclient.ts
nolanmar511: Fixed.

ts/lib/auth/googleauth.ts
nolanmar511: Fixed.

ts/lib/auth/googleauth.ts
nolanmar511: Switched to public here. There was no reason for this.

ts/lib/auth/jwtclient.ts
nolanmar511: Fixed

ts/lib/auth/oauth2client.ts
nolanmar511: Fixed

ts/lib/auth/refreshclient.ts
nolanmar511: Fixed

JustinBeckwith:
Instead of trying to get the tests to pass on the master branch, please take a look at rebasing against the next branch.

nolanmar511:
Opened PR #227, which has these changes on top of the next branch.

PR #209, opened 2017-12-19
fix: update code to support typescript@2.6.2

JustinBeckwith:
The tests for Node.js 4 and Node.js 6 randomly started failing due to what seems like a compiler change.  Our `package-lock.json` is pinning to `typescript@2.6.1`, which covers Node.js 8.  The `~2.6.0` tag in our `package.json` is letting new patch versions of typescript come through as designed, resulting in `typescript@2.6.2` getting installed.  I suspect TypeScript made a breaking change in a patch release :(

### DIFF
--- a/test/test.oauth2.ts
+++ b/test/test.oauth2.ts
@@ -17,6 +17,7 @@
 import * as assert from 'assert';
 import {AxiosRequestConfig} from 'axios';
 import * as crypto from 'crypto';
+import {randomBytes} from 'crypto';
 import * as fs from 'fs';
 import * as nock from 'nock';
 import * as path from 'path';
@@ -46,8 +47,10 @@ describe('OAuth2 client', () => {
         new OAuth2Client(CLIENT_ID, CLIENT_SECRET, REDIRECT_URI);
     const generated = oauth2client.generateAuthUrl(opts);
     const parsed = url.parse(generated);
+    if (typeof parsed.query !== 'string') {
+      throw new Error('Unable to parse querystring');
+    }
     const query = qs.parse(parsed.query);
-
     assert.equal(query.response_type, 'code token');
     assert.equal(query.access_type, ACCESS_TYPE);
     assert.equal(query.scope, SCOPE);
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
@alexander-fenster for review as well

ofrobots:
@JustinBeckwith is the version of `@types/node` different between what you got and what `package-lock.json` has? That could explain the breakage too.

test/test.oauth2.ts
```diff
@@ -46,14 +47,17 @@ describe('OAuth2 client', () => {
...
     const parsed = url.parse(generated);
-    const query = qs.parse(parsed.query);
-
-    assert.equal(query.response_type, 'code token');
-    assert.equal(query.access_type, ACCESS_TYPE);
-    assert.equal(query.scope, SCOPE);
-    assert.equal(query.client_id, CLIENT_ID);
-    assert.equal(query.redirect_uri, REDIRECT_URI);
-    done();
+    if (typeof parsed.query === 'string') {
```
ofrobots: nit: reverse this branch. It is preferable to write branches with error condition throw first. Errors return early and so neither the indent not the early return for the main logic would be necessary.

test/test.oauth2.ts
```diff
@@ -67,10 +71,13 @@ describe('OAuth2 client', () => {
         new OAuth2Client(CLIENT_ID, CLIENT_SECRET, REDIRECT_URI);
     const generated = oauth2client.generateAuthUrl(opts);
     const parsed = url.parse(generated);
-    const query = qs.parse(parsed.query);
-
-    assert.equal(query.scope, SCOPE_ARRAY.join(' '));
-    done();
+    if (typeof parsed.query === 'string') {
```
ofrobots: ditto

test/test.oauth2.ts
```diff
@@ -80,10 +87,13 @@ describe('OAuth2 client', () => {
            new OAuth2Client(CLIENT_ID, CLIENT_SECRET, REDIRECT_URI);
        const generated = oauth2client.generateAuthUrl();
        const parsed = url.parse(generated);
-       const query = qs.parse(parsed.query);
-
-       assert.equal(query.response_type, 'code');
-       done();
+       if (typeof parsed.query === 'string') {
```
ofrobots: ditto

test/test.oauth2.ts
```diff
@@ -793,17 +803,25 @@ describe('OAuth2 client', () => {
         new OAuth2Client(CLIENT_ID, CLIENT_SECRET, REDIRECT_URI);
     const generated = oauth2client.generateAuthUrl({});
     const parsed = url.parse(generated);
-    const query = qs.parse(parsed.query);
-    assert.equal(query.redirect_uri, REDIRECT_URI);
+    if (typeof parsed.query === 'string') {
```
ofrobots: I'll stop pointing these out.

test/test.oauth2.ts
JustinBeckwith: Yikes, weird I missed that.  Fixed.

Issue #111, opened 2016-12-26
Oauth2.refreshAccessToken returns invalid_request


0xmtn:
I'm having issues when I try to use "refreshAccessToken" function: `invalid_request`. So after so many hours wasted on this I tried to use the private "refreshToken_" function, which returned the main reason why it is an invalid request: `Could not determine client ID from request`.

Here is the code I'm trying:

```
    var auth = new google_auth();   
    var oauth2Client = new auth.OAuth2(conf.cliend_id, conf.client_secret, conf.red_url);
    var creds = {access_token: accessToken,  // **I tried without this too**
                refresh_token: refreshToken};
    oauth2Client.setCredentials(creds);
    oauth2Client.refreshAccessToken(function(err, tokens){
      console.log("ERR: ", err); // keeps saying invalid_request 400 and nothing more 
      console.log("TOKENS: ", tokens); // NULL
```
Couldn't find a solution, so I tried this:

```
    var auth = new google_auth();   
    var oauth2Client = new auth.OAuth2(conf.cliend_id, conf.client_secret, conf.red_url);
    var creds = {access_token: accessToken,  // **I tried without this too**
                refresh_token: refreshToken};
    oauth2Client.setCredentials(creds); // Tried without this too
    oauth2Client.refreshToken_(refreshToken, function(err, tokens, response){
      console.log("ERR: ", err); // keeps saying invalid_request 400 and nothing more.
      console.log("RESPONSE: ", response); //bunch of stuff here, also where I found the reason for the error
      console.log("TOKENS: ", tokens); // NULL
```
I am on server side using `google-auth-library` and `googleapis`.

How can I solve this? Is this a problem on my side or is there anything else I should know about this APIs?

ianmetcalf:
Not sure if the above code was copied verbatim or not, but you have a typo `conf.cliend_id` instead of `conf.client_id`. Again I'm not sure if this is just mis-copied, and I don't know anything about where your setting the `conf` object, but probably worth a double check.

JustinBeckwith:
Fixed in #215

Issue #120, opened 2017-01-10
OAuth2 getToken invalid_request


sejbr:
I'm quite new to both NodeJS and the new GoogleApis since I've been always using keys mostly.
I've had quite a stumble with [this](https://developers.google.com/identity/protocols/OAuth2InstalledApp) since it uses different oauth2 urls, be it /oauth2/v2/auth instead of /oauth2/auth and /oauth2/v4/token instead of /oauth2/token, I'm guessing because of the ability not to use secrets for installed apps in those versions.
Any way of changing the versions without editing the privates in oauth2client.js?
I've seen usage like require('googleapis').youtube('v3') but I've tried different things and can't seem to get it to work without changing the consts.

JustinBeckwith:
These urls are changed in #215, which will be in the upcoming 1.0 release of the library.

PR #215, opened 2017-12-22
chore: docs and samples for refresh token, update uris

JustinBeckwith:
Addresses #111

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,24 @@ oauth2Client.setCredentials(tokens);
 ```
 
 ##### Manually refreshing access token
-If you need to manually refresh the `access_token` associated with your OAuth2 client, make sure you have a `refresh_token` set in your credentials first and then call:
+If you need to manually refresh the `access_token` associated with your OAuth2 client, ensure the call to `generateAuthUrl` sets the `access_type` to `offline`.  The refresh token will only be returned for the first authorization by the user.  To force consent, set the `prompt` property to `consent`:
+
+```js
+// Generate the url that will be used for the consent dialog.
+const authorizeUrl = oAuth2Client.generateAuthUrl({
+  // To get a refresh token, you MUST set access_type to `offline`.
+  access_type: 'offline',
+  // set the appropriate scopes
+  scope: 'https://www.googleapis.com/auth/plus.me',
+  // A refresh token is only returned the first time the user
+  // consents to providing access.  For illustration purposes,
+  // setting the prompt to 'consent' will force this consent
+  // every time, forcing a refresh_token to be returned.
+  prompt: 'consent'
+});
+```
+
+If a refresh_token is set again on `OAuth2Client.credentials.refresh_token`, you can can `refreshAccessToken()`:
 
 ``` js
 const tokens = await oauth2Client.refreshAccessToken();
--- a/examples/refreshAccessToken.js
+++ b/examples/refreshAccessToken.js
@@ -1,0 +1,95 @@
+// Copyright 2017, Google, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const OAuth2Client = require('../build/src/index').OAuth2Client;
+const http = require('http');
+const url = require('url');
+const querystring = require('querystring');
+const opn = require('opn');
+
+// Download your OAuth2 configuration from the Google
+const keys = require('./oauth2.keys.json');
+
+/**
+ * Start by acquiring a pre-authenticated oAuth2 client.
+ */
+async function main() {
+  try {
+    const oAuth2Client = await getAuthenticatedClient();
+    // If you're going to save the refresh_token, make sure
+    // to put it somewhere safe!
+    console.log(`Refresh Token: ${oAuth2Client.credentials.refresh_token}`);
+    console.log(`Expiration: ${oAuth2Client.credentials.expiry_date}`);
+    console.log('Refreshing access token ...');
+    const res = await oAuth2Client.refreshAccessToken();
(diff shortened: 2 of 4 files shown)


examples/refreshAccessToken.js
```diff
@@ -0,0 +1,95 @@
...
+ */
+async function main() {
+  try {
+    const oAuth2Client = await getAuthenticatedClient();
+    // If you're going to save the refresh_token, make sure
+    // to put it somewhere safe!
+    console.log(`Refresh Token: ${oAuth2Client.credentials.refresh_token}`);
+    console.log(`Expiration: ${oAuth2Client.credentials.expiry_date}`);
+    console.log('Refreshing access token ...');
+    const res = await oAuth2Client.refreshAccessToken();
```
ofrobots: @nolanmar511 perhaps this pattern can be temporary workaround while #208 is figured out?

README.md
```diff
@@ -201,7 +201,24 @@ oauth2Client.setCredentials(tokens);
...
+  scope: 'https://www.googleapis.com/auth/plus.me',
+  // A refresh token is only returned the first time the user
+  // consents to providing access.  For illustration purposes,
+  // setting the prompt to 'consent' will force this consent
+  // every time, forcing a refresh_token to be returned.
+  prompt: 'consent'
+});
+```
+
+If a refresh_token is set again on `OAuth2Client.credentials.refresh_token`, you can cann `refreshAccessToken()`:
```
ofrobots: typo: cann?

src/auth/oauth2client.ts
```diff
@@ -25,11 +25,102 @@ import {CredentialRequest, Credentials} from './credentials';
 import {LoginTicket} from './loginticket';
 
 export interface GenerateAuthUrlOpts {
+  /**
+   * Recommended. Indicates whether your application can refresh access tokens
+   * when the user is not present at the browser. Valid parameter values are
+   * online, which is the default value, and offline. Set the value to offline
```
ofrobots: Put literal values in quotes.

src/auth/oauth2client.ts
```diff
@@ -25,11 +25,102 @@ import {CredentialRequest, Credentials} from './credentials';
...
+   * present at the browser. This is the method of refreshing access tokens
+   * described later in this document. This value instructs the Google
+   * authorization server to return a refresh token and an access token the
+   * first time that your application exchanges an authorization code for
+   * tokens.
+   */
   access_type?: string;
+
+  /**
+   * The response_type will always be set to CODE.
```
ofrobots: Put literal value in quotes.

examples/refreshAccessToken.js
nolanmar511: Possibly, but it would require this in @google-cloud/common, which would require exposing this in google-auto-auth. 
Depending on the time to resolve #208, I'm not certain that exposing this would be worth the effort.

README.md
JustinBeckwith: fixed

src/auth/oauth2client.ts

src/auth/oauth2client.ts

src/auth/oauth2client.ts
```diff
@@ -25,11 +25,103 @@ import {CredentialRequest, Credentials} from './credentials';
...
+
+  /**
+   * The 'response_type' will always be set to 'CODE'.
+   */
   response_type?: string;
+
+  /**
+   * Required. The client ID for your application. You can find this value in
+   * the API Console.
+   */
```
ofrobots: Dissonance. If this required, why is it marked optional?

JustinBeckwith: Good catch.  I was copying the documentation from Google's OAuth docs.  This value is required in the HTTP call, but will be set for you based on the value passed into the constructor.  I made this explicit in the jsdoc.

examples/refreshAccessToken.js
```diff
@@ -0,0 +1,95 @@
...
+const opn = require('opn');
+
+// Download your OAuth2 configuration from the Google
+const keys = require('./oauth2.keys.json');
+
+/**
+ * Start by acquiring a pre-authenticated oAuth2 client.
+ */
+async function main() {
+  try {
```
ofrobots: nit: use the `main().catch(console.error)` pattern instead?

src/auth/oauth2client.ts
```diff
@@ -130,13 +223,13 @@ export class OAuth2Client extends AuthClient {
...
   private static readonly GOOGLE_OAUTH2_AUTH_BASE_URL_ =
-      'https://accounts.google.com/o/oauth2/auth';
+      'https://accounts.google.com/o/oauth2/v2/auth';
 
   /**
    * The base endpoint for token retrieval.
    */
   private static readonly GOOGLE_OAUTH2_TOKEN_URL_ =
-      'https://accounts.google.com/o/oauth2/token';
+      'https://www.googleapis.com/oauth2/v4/token';
```
ofrobots: What is the semver impact of this change?

ofrobots: Also: the justification for this change needs to be documented in the PR and the commit message.

examples/refreshAccessToken.js
JustinBeckwith: This sent me down a dark rabbit hole from which I've yet to emerge.  I'm having a lot of trouble getting the process to exit naturally, without explicitly calling `process.exit()`.  The http server is hanging onto active connections, so I had to explicitly track all incoming connections, and manually close them to get the server to close.  I'm also running into issues with the `opn` npm module hanging onto it's handle, so I had to do a bunch of hacking to make that work.  In the end, I had added like 50 lines of code to make the process exit naturally, and it really added nothing to the sample.  

Are you ok with me leaving this one the way it is?  I'd be happy to open an issue to dig into this deeper after we get next merged into master.

ofrobots: sgtm

src/auth/oauth2client.ts
JustinBeckwith: I would say this is a semver minor change.  I have no reason to believe it breaks any functionality, but it does (I think) bring us up to date with the most recent documentation:
https://developers.google.com/identity/protocols/OAuth2InstalledApp

It's also suggested in #120 that this is causing problems for users as well.  I will update the PR and commit message to call these both out.

JustinBeckwith:
Note:  This is a semver minor change (though it will be included in the 1.0 release). I have no reason to believe it breaks any functionality, but it does (I think) bring us up to date with the most recent documentation:
https://developers.google.com/identity/protocols/OAuth2InstalledApp

It's also suggested in #120 that this is causing problems for users as well. I will update the PR and commit message to call these both out.

Issue #187, opened 2017-11-23
1000s of calls to child_process.exec


kolodny:
When a new instance of a google auth instance is created, it looks calls the `getDefaultProjectId` method. This method does caching, however the cache only really works once the first call settles. More aggressive caching can be called for if there is a loop creating hundreds of instances. The following is something I threw together which solves the problem. I'm not really familiar with ts enough yet to create a PR but the patch should look something like this:

<strike>

```diff
diff --git a/ts/lib/auth/googleauth.ts b/ts/lib/auth/googleauth.ts
index deb2b90..77b999b 100644
--- a/ts/lib/auth/googleauth.ts
+++ b/ts/lib/auth/googleauth.ts
@@ -32,6 +32,8 @@ import {JWT} from './jwtclient';
 import {OAuth2Client} from './oauth2client';
 import {UserRefreshClient} from './refreshclient';

+let getDefaultProjectIdPromise = null;
+
 export interface ProjectIdCallback { (err: Error, projectId: string): void }

 export interface CredentialBody {
@@ -126,14 +128,33 @@ export class GoogleAuth {
         this.callback(callback, null, this._cachedProjectId);
       });
     } else {
+      let resolve = null;
+      let reject = null;
       const my_callback = (err: Error, projectId: string) => {
         if (!err && projectId) {
           this._cachedProjectId = projectId;
+          if (resolve) {
+            resolve(projectId);
+          }
+        } else {
+          if (reject) {
+            reject(error);
+          }
         }
         setImmediate(() => {
           this.callback(callback, err, projectId);
         });
       };
+      if (getDefaultProjectIdPromise) {
+        getDefaultProjectIdPromise.then(
+          (projectId) => my_callback(null, projectId),
+          (error) => my_callback(error),
+        );
+        return;
+      }
+      getDefaultProjectIdPromise = new Promise((res, rej) => {
+        resolve = res;
+        reject = rej;
+      });

       // environment variable
       if (this._getProductionProjectId(my_callback)) {
```

</strike>

<sub>Edit: That last bit was an overly complex fix, see the next comment instead</sub>

The following file highlights the calls to exec from this library:

```js
const childProcess = require('child_process');

const oldExec = childProcess.exec;
childProcess.exec = function() {
  console.log(`exec called ${new Error().stack.split('\n')[2].trim()}`);
  return oldExec.apply(this, arguments);
}

const BigTable = require('@google-cloud/bigtable');
const rowCount = 10;

[...]

PR #216, opened 2017-12-23
fix: cache promise instead of ProjectId

JustinBeckwith:
Addresses #187

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -74,13 +74,8 @@ export class GoogleAuth {
     return this.checkIsGCE;
   }
 
-  private _cachedProjectId: string;
+  private _getDefaultProjectIdPromise: Promise<string|null>;
 
-  // Note:  this properly is only public to satisify unit tests.
-  // https://github.com/Microsoft/TypeScript/issues/5228
-  set cachedProjectId(projectId: string) {
-    this._cachedProjectId = projectId;
-  }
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
The way this code works today, it can be kind of a bad bug.  Take a look at #187 - we're queuing up thousands of calls, waiting for the initial projectId to populate.  Generally - I'd say it's more important to fix the bug than to make it work with Stackdriver (totally willing to debate it of course). 

On the discussion of this library vs google-auto-auth - I view that library as a band aid, because this one was basically abandoned.  I would like to reach a place where it isn't needed anymore, but it is going to take time. 

If we wanted to cache the promise, **and** not break context propagation - what can I do?

ofrobots:
> Generally - I'd say it's more important to fix the bug than to make it work with Stackdriver

It is not about Stackdriver. This is about breaking long stack traces, NewRelic and all other async debugging tools out this. This module is the basis of all `@google-cloud/` modules which means that all of them will stop working with these debug / tracing tools.

> If we wanted to cache the promise, and not break context propagation - what can I do?

Similar to what was done in https://github.com/stephenplusplus/google-auto-auth/pull/21. Basically, use the promise as a queue while it is pending. Once resolved, cache the value too, and stop using the promise.

JustinBeckwith:
I think I got it :)  @ofrobots can you take another look?

PR #219, opened 2017-12-25
fix: improve typing around tokens and add sample

JustinBeckwith:
While writing a sample I noticed that we were missing typing and properties around a few objects.  Much better now :)


JustinBeckwith:
Ping @google/google-node-team can someone take a look?

examples/verifyIdToken.js
```diff
@@ -0,0 +1,97 @@
...
+    console.log(ticket);
+
+    // You can use this info to get user information too.
+    const url = `https://www.googleapis.com/plus/v1/people/me`;
+    const res = await oAuth2Client.request({url})
+    console.log(res.data);
+  } catch (e) {
+    console.error(e);
+  }
+  process.exit();
```
jinwoo: I'm not sure why this is needed.

src/auth/loginticket.ts
```diff
@@ -62,3 +61,106 @@ export class LoginTicket {
...
+   * presented only once.
+   */
+  nonce?: string;
+
+  /**
+   * The hosted G Suite domain of the user. Provided only if the user belongs to
+   * a hosted domain.
+   */
+  hd?: string;
+}
```
jinwoo: Newline at the end?

JustinBeckwith: fixed

examples/verifyIdToken.js
JustinBeckwith: There is some problem with the http server and opn modules grabbing handles that prevent the app from finishing naturally.  I could fix it, but it adds another ~50 LOC to the sample that don't add much value.

Issue #152, opened 2017-07-15
Should allow for providing credential JSON in different ways


askdesigners:
This library currently cannot work on hosting providers that have ephemeral file systems like Heroku. It would be handy if there were a way to provide this object directly, rather than by file path. 

Since these systems have an ephemeral fs you cannot easily include files there without having them checked into the git repo.

ofrobots:
I believe this should already work:

```js
const GoogleAuth = require('google-auth-library');
const googleAuth = new GoogleAuth();

googleAuth.fromJSON("your JSON key", (err, authClient) => { .... });
```

@askdesigners, @logicoder, @fhuel: does this address your use cases?

I can adjust the docs which don't seem to provide details on these. I would also suggest checking out [google-auto-auth](https://www.npmjs.com/package/google-auto-auth) which provides a higher level API to work with google auth that might be a bit more convenient to use.

JustinBeckwith:
Fixed in #220

PR #220, opened 2017-12-26
docs: add example and readme for using fromJSON

JustinBeckwith:
Addresses #152

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,56 @@ main();
 
 The parameters for the JWT auth client including how to use it with a `.pem` file are explained in [samples/jwt.js](samples/jwt.js).
 
+#### Loading credentials from environment variables
+
+Instead of loading credentials from a key file, you can also provide them using an environment variable and the `GoogleAuth.fromJSON()` method.  This is particularly convenient for systems that deploy directly from source control (Heroku, App Engine, etc).
+
+Start by exporting your credentials:
+
+```
+$ export CREDS='{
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -245,6 +245,56 @@ main();
...
+  try {
+    const auth = new GoogleAuth();
+    // load the JWT or UserRefreshClient from the keys
+    const client = auth.fromJSON(keys);
+    client.scopes = ['https://www.googleapis.com/auth/cloud-platform'];
+    await client.authorize();
+    const url = `https://www.googleapis.com/dns/v1/projects/${keys.project_id}`;
+    const res = await client.request({url});
+    console.log(res.data);
+  } catch(e) {
```
ofrobots: nit: I think the example will be simpler if you dropped the try-catch from here and changed the `main` invocation to:

```js
main().catch(console.error);
```

Or you can drop the try-catch completely, and assume the user knows how to deal with errors.

README.md
```diff
@@ -245,6 +245,56 @@ main();
...
+    console.error(e);
+  }
+  process.exit();
```
ofrobots: Why is this here?

README.md
JustinBeckwith: Nice!  I went with the `main().catch` approach.

README.md
JustinBeckwith: Not needed, got rid of it.

Issue #65, opened 2015-11-10
A proxy setting in getToken


jojonki:
I faced a trouble of setting proxy during getting token. When this library tries to get token, it looks it doesn't consider proxy setting. My code works if I set a proxy as a request property directly.
https://github.com/google/google-auth-library-nodejs/blob/master/lib/auth/oauth2client.js#L143

Or are there any ways to setting proxy even if trying to getting tokens?

runk:
Stuck at the same issue.

You can set `HTTP_PROXY` and `HTTPS_PROXY` environment vars to corresponding value. Those vars are recognized by `request` module which `google-auth-library-nodejs` uses internally

JustinBeckwith:
Turns out this works out of the box via the `HTTP_PROXY` and `HTTPS_PROXY` env vars :)  Adding  docs around that in #221.

PR #221, opened 2017-12-27
docs: document proxy behavior and verify with a test

JustinBeckwith:
The axios npm module supports setting an HTTP or HTTPS proxy via the http_proxy and https_proxy environment variables.  It also happens to be undocumented, which I didn't figure out until writing my own implementation  😂

Here I'm just adding a few tests, and updating our docs.  I will follow up with a PR against the axios docs to save others from my fate.  

Addresses #65.

### DIFF
--- a/test/test.transporters.ts
+++ b/test/test.transporters.ts
@@ -98,6 +98,19 @@ describe('Transporters', () => {
           done();
         });
   });
+});
+
+describe('transporter proxy', () => {
+  let savedEnv: NodeJS.ProcessEnv;
+
+  beforeEach(() => {
+    savedEnv = process.env;
+    process.env = {};
+  });
+
+  afterEach(() => {
+    process.env = savedEnv;
+  });
 
   it('should use the http proxy if one is configured', async () => {
     process.env['http_proxy'] = 'http://han:solo@proxy-server:1234';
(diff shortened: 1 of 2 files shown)


test/test.transporters.ts
```diff
@@ -98,4 +98,42 @@ describe('Transporters', () => {
...
+            'accept': /.*/g,
+            'user-agent': /google-api-nodejs-client\/.*/g,
+            'proxy-authorization': /.*/g
+          }
+        })
+        .reply(200);
+    const url = 'http://example.com/fake';
+    const result = await transporter.request({url});
+    assert.equal(result.status, 200);
+    process.env['http_proxy'] = undefined;
```
ofrobots: This clobbers the original value of `process.env.http_proxy`

test/test.transporters.ts
```diff
@@ -98,4 +98,42 @@ describe('Transporters', () => {
           done();
         });
   });
+
+  it('should use the http proxy if one is configured', async () => {
```
ofrobots: nit: put these two tests in a describe section. Use something like the following to ensure you're not leaving side effects between/after tests:

```ts
let savedEnv;
beforeEach(() => {
  savedEnv = process.env;
  process.env = {};
});

afterEach(() => { process.env = savedEnv; });
```

test/test.transporters.ts
JustinBeckwith: Fixed!

test/test.transporters.ts
JustinBeckwith: fixed!

PR #227, opened 2018-01-03

nolanmar511:
This can be used to address the issues described in #133, by allowing users to specify that tokens should be refreshed if they will expire soon. This is also necessary to fix GoogleCloudPlatform/cloud-profiler-nodejs#94

For earlier discussion of these changes, see PR #208.


JustinBeckwith:
Thanks!  We need to land #230 first.

JustinBeckwith:
@nolanmar511 So sorry for all of the churn here.  We're finally done with all of the crazy, and you should be free to get this moving again.  Now we can rebase this against the master branch, which is mostly what we had in next.  I ran into some trouble porting my PRs, and ended up using `git format-patch` and `git apply` to port mine over.

nolanmar511:
Now rebasing against master. See #242

PR #230, opened 2018-01-04
chore: accept options objects in constructors

JustinBeckwith:
Should make #227 much more palatable.

### DIFF
--- a/src/auth/jwtclient.ts
+++ b/src/auth/jwtclient.ts
@@ -23,12 +23,20 @@ import {GetTokenResponse, OAuth2Client, RequestMetadataResponse} from './oauth2c
 
 const isString = require('lodash.isstring');
 
+export interface JWTOptions {
+  email?: string;
+  keyFile?: string;
+  key?: string;
+  scopes?: string|string[];
+  subject?: string;
+}
+
 export class JWT extends OAuth2Client {
   email?: string;
-  keyFile?: string|null;
-  key?: string|null;
-  scopes?: string|string[]|null;
-  scope?: string|null;
+  keyFile?: string;
+  key?: string;
+  scopes?: string|string[];
+  scope?: string;
   subject?: string;
   gtoken: GoogleToken;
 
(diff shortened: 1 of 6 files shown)


src/auth/jwtclient.ts
```diff
@@ -44,15 +52,22 @@ export class JWT extends OAuth2Client {
...
-    this.scopes = scopes;
-    this.subject = subject;
+    const opts = (optionsOrEmail && typeof optionsOrEmail === 'object') ?
+        optionsOrEmail :
+        {email: optionsOrEmail, keyFile, key, scopes, subject};
+    this.email = opts.email;
+    this.keyFile = opts.keyFile;
+    this.key = opts.key;
+    this.scopes = opts.scopes;
+    this.subject = opts.subject;
```
ofrobots: I was thinking that there perhaps there is a clever way to employ [destructuring](https://stackoverflow.com/questions/17781472/how-to-get-a-subset-of-a-javascript-objects-properties) here. The result is overcomplicated. Worst of all it repeats property names.

```js
Object.assign(this, 
  (({email, keyFile, key, scopes, subject}) => 
   ({email, keyFile, key, scopes, subject}))(opts));
```

An alternatively would be `lodash.pick`: 
```js
Object.assign(this, pick(opts, ['email', 'keyFile', 'key', 'scopes', 'subject']));
```

On second thought; I think your version is fine. Fun exercise though.

src/auth/oauth2client.ts
```diff
@@ -241,14 +245,20 @@ export class OAuth2Client extends AuthClient {
    * @param {Object=} opts optional options for overriding the given parameters.
    * @constructor
    */
+  constructor(options: OAuth2ClientOptions);
+  constructor(clientId?: string, clientSecret?: string, redirectUri?: string);
   constructor(
-      clientId?: string, clientSecret?: string, redirectUri?: string,
-      opts: AuthClientOpts = {}) {
```
ofrobots: Making sure: the opts object was added in the `next` branch? The delta here looks semver major to me.

test/test.jwt.ts
```diff
@@ -73,9 +73,12 @@ describe('Initial credentials', () => {
 describe('JWT auth client', () => {
   describe('.authorize', () => {
     it('should get an initial access token', done => {
-      const jwt = new JWT(
-          'foo@serviceaccount.com', PEM_PATH, null,
-          ['http://bar', 'http://foo'], 'bar@subjectaccount.com');
+      const jwt = new JWT({
```
ofrobots: As far as I can tell, all the constructions in the tests are using the new options object form. Please add a test that makes sure that the other form of the `emailOrOptions is object` branch is covered.

Let me know if this is already covered and I missed it.

test/test.oauth2.ts
```diff
@@ -44,8 +44,12 @@ describe('OAuth2 client', () => {
       response_type: 'code token'
     };
 
-    const oauth2client =
-        new OAuth2Client(CLIENT_ID, CLIENT_SECRET, REDIRECT_URI);
+    const oauth2client = new OAuth2Client({
```
ofrobots: likewise.

src/auth/jwtclient.ts
JustinBeckwith: I had the same exact thought :)

src/auth/oauth2client.ts
JustinBeckwith: Yikes, good catch.  Going to go add a test to make sure I don't break this.

test/test.oauth2.ts
JustinBeckwith: The first test uses the object format, the others use the exploded format.  I also added a test that overrides authbaseUrl and tokenUrl - because no test covered those at all :(

test/test.jwt.ts
JustinBeckwith: Updated the first test to make sure that the non-object form is covered.

JustinBeckwith:
@ofrobots ready for another pass!

Issue #228, opened 2018-01-03
Apply linting and format rules to the examples


JustinBeckwith:
The examples are all in JavaScript, so we don't have linting for formatting rules enabled.  We should do that :)

JustinBeckwith:
Fixed in #233

PR #233, opened 2018-01-05
chore: apply code style rules to javascript

JustinBeckwith:
Addresses #228

### DIFF
--- a/examples/adc.js
+++ b/examples/adc.js
@@ -42,28 +42,27 @@ async function main() {
 
 /**
  * Instead of specifying the type of client you'd like to use (JWT, OAuth2, etc)
- * this library will automatically choose the right client based on the
- * environment.
+ * this library will automatically choose the right client based on the environment.
  */
 async function getADC() {
-  // Acquire a client and the projectId based on the environment. This method
-  // looks for the GCLOUD_PROJECT and GOOGLE_APPLICATION_CREDENTIALS environment
-  // variables.
+  // Acquire a client and the projectId based on the environment. This method looks
+  // for the GCLOUD_PROJECT and GOOGLE_APPLICATION_CREDENTIALS environment variables.
   const res = await auth.getApplicationDefault();
   let client = res.credential;
 
-  // The createScopedRequired method returns true when running on GAE or a local
-  // developer machine. In that case, the desired scopes must be passed in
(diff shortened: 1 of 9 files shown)


examples/adc.js
```diff
@@ -43,27 +42,29 @@ async function main() {
...
  * Instead of specifying the type of client you'd like to use (JWT, OAuth2, etc)
- * this library will automatically choose the right client based on the environment.
+ * this library will automatically choose the right client based on the
+ * environment.
  */
-async function getADC() {
-  // Acquire a client and the projectId based on the environment. This method looks
-  // for the GCLOUD_PROJECT and GOOGLE_APPLICATION_CREDENTIALS environment variables.
+async function
+getADC() {
```
jinwoo: It looks quite weird to me to split `async function` and the function name across two lines, but if this is what the tool does, ok.

examples/fromJSON.js
```diff
@@ -33,8 +33,9 @@ const {GoogleAuth} = require('../build/src/index');
  *       "client_id": "your-client-id",
  *       "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  *       "token_uri": "https://accounts.google.com/o/oauth2/token",
- *       "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
- *       "client_x509_cert_url": "your-cert-url"
+ *       "auth_provider_x509_cert_url":
+ *"https://www.googleapis.com/oauth2/v1/certs", "client_x509_cert_url":
```
jinwoo: This doesn't look ok to me though. Is there any way to convince clang-format otherwise?

JustinBeckwith: Yeah I have no idea what clang-format is trying to do here.  I kind of wonder out loud if it supports the `async` keyword....

jinwoo: Doesn't seem related to `async`. The same thing happens without `async` too. See https://github.com/google/google-auth-library-nodejs/pull/233/files#diff-7538725028be787890f7305df2148f61R43

JustinBeckwith:
@jinwoo this is ready for another look.  I switched over to prettier, which seems to work way better.

JustinBeckwith:
CI failures are unrelated, and discussed in #236

Issue #55, opened 2015-08-28
API conventions


stephenplusplus:
Oh boy, it's me again.

I think this library would be easier to fit into a modern Node developer's code if it was a bit easier to use.

Using the example from the readme:

``` js
(new GoogleAuth).getApplicationDefault(function(err, authClient) {
```

It's very rare you find the use of `new Something()` anymore. OO is still an important tool for structuring the code, but I think an approach like [gcloud](http://gitnpm.com/gcloud) is better for the end user; basically, use OO internally, but don't require the user to see it. Alias by camelCase (as opposed to UpperCamelCase), and auto-initialize for them:

``` js
function GoogleAuth() {
  if (!(this instanceof GoogleAuth)) {
    return new GoogleAuth();
  }
}

module.exports.googleAuth = GoogleAuth;

// for the user:
googleAuth()
// is equivalent to:
new googleAuth()
```
## 

``` js
  if (err === null) {
    // Inject scopes if they have not been injected by the environment
    if (authClient.createScopedRequired && authClient.createScopedRequired()) {
```

`createScopedRequired` being a method that may or may not exist is confusing. Can't this always be defined? Or maybe from the docs, it can be declared in what cases scopes are required up front?

*\* Edit *\*  Looking at this again, couldn't it just be a bool? If the function is sync, it should just be `if (authClient.createScopedRequired) {}`
## 

``` js
      var scopes = [
        'https://www.googleapis.com/auth/cloud-platform',
        'https://www.googleapis.com/auth/compute'
      ];
      authClient = authClient.createScoped(scopes);
    }
```

The naming here is a bit confusing -- maybe it's just me. `createScopedRequired` and `createScoped`. Maybe this could be `scopesRequired` and `addScopes`.

Using `addScopes` would also avoid the re-assignment of the `authClient`. The pattern there feels a bit backwards. Get an authClient then get another one from it. It would be great if I could have gotten the authClient right from the start (as mentioned in the last suggestion with docs + requiring docs up front) or modify it so that it's good to go (`addScopes`).
## 

``` js
  // Fetch the access token
    var _ = require(lodash);
    var optionalUri = null;  // optionally specify the URI being authorized
    var reqHeaders = {};
    authClient.getRequestMetadata(optionalUri, function(err, headers)) {
      if (err === null) {
        // Use authorization headers
        reqHeaders = _.merge(allHeaders, headers);
      }
    });
```

[...]

Issue #54, opened 2015-08-28
Add convenience methods for common use cases


stephenplusplus:
Hello again!

I was wondering if it would be possible to design some convenience wrappers around common uses for this library?

I can't appreciate all the nuances and complexities this library has to support, so I can't comment on all of the use cases that can be simplified. But as an example, I made google-auto-auth to make authing with application default, JSON/p12 keyfile, and credentials JS objects easier: https://github.com/stephenplusplus/google-auto-auth/blob/0bea71c76c46158a4a4911b3f16af35f7a16ae9d/index.js#L81-L92

I'm still not entirely sure I'm using this library's API the right way (#53), but a developer using google-auto-auth is surely going to have an easier time getting started.

Thanks!

ofrobots:
@stephenplusplus do you still care about this?

stephenplusplus:
This is what started all these issues: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/832

Ideally, this library would be easier to use and better documented so we could replace and deprecate google-auto-auth.

ofrobots:
Hmm.. I quite like `google-auto-auth` – it is nice and easy to use. Do we still want to deprecate it?

/cc @JustinBeckwith

stephenplusplus:
I don't mind either way :)

PR #242, opened 2018-01-10

nolanmar511:
This can be used to address the issues described in #133, by allowing users to specify that tokens should be refreshed if they will expire soon. This is also necessary to fix GoogleCloudPlatform/cloud-profiler-nodejs#94

For earlier discussion of these changes, see PR #208.

### DIFF
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -33,8 +33,8 @@ export class Compute extends OAuth2Client {
    * Retrieve access token from the metadata server.
    * See: https://developers.google.com/compute/docs/authentication
    */
-  constructor() {
-    super();
+  constructor(refreshTokenEarlyMillis?: number) {
+    super({refreshTokenEarlyMillis});
     // Start with an expired refresh token, which will automatically be
     // refreshed before the first API call is made.
     this.credentials = {expiry_date: 1, refresh_token: 'compute-placeholder'};
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -20,6 +20,8 @@ import {RequestError} from './../transporters';
 import {CredentialRequest, Credentials} from './credentials';
 import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
 
+export interface ComputeOptions extends RefreshOptions {}
+
 export class Compute extends OAuth2Client {
   /**
    * Google Compute Engine metadata server token endpoint.
@@ -33,7 +35,7 @@ export class Compute extends OAuth2Client {
    * Retrieve access token from the metadata server.
    * See: https://developers.google.com/compute/docs/authentication
    */
-  constructor(options?: RefreshOptions) {
+  constructor(options?: ComputeOptions) {
     super(options);
     // Start with an expired refresh token, which will automatically be
     // refreshed before the first API call is made.
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -37,7 +37,7 @@ export class Compute extends OAuth2Client {
    */
   constructor(options?: ComputeOptions) {
     options = options || {};
-    super({refreshTokenEarlyMillis: options.refreshTokenEarlyMillis});
+    super({eagerRefreshThresholdMillis: options.eagerRefreshThresholdMillis});
     // Start with an expired refresh token, which will automatically be
     // refreshed before the first API call is made.
     this.credentials = {expiry_date: 1, refresh_token: 'compute-placeholder'};
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -20,8 +20,6 @@ import {RequestError} from './../transporters';
 import {CredentialRequest, Credentials} from './credentials';
 import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
 
-export interface ComputeOptions extends RefreshOptions {}
-
 export class Compute extends OAuth2Client {
   /**
    * Google Compute Engine metadata server token endpoint.
@@ -35,9 +33,8 @@ export class Compute extends OAuth2Client {
    * Retrieve access token from the metadata server.
    * See: https://developers.google.com/compute/docs/authentication
    */
-  constructor(options?: ComputeOptions) {
-    options = options || {};
-    super({eagerRefreshThresholdMillis: options.eagerRefreshThresholdMillis});
+  constructor(options?: RefreshOptions) {
+    super(options);
     // Start with an expired refresh token, which will automatically be
     // refreshed before the first API call is made.
     this.credentials = {expiry_date: 1, refresh_token: 'compute-placeholder'};
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -20,6 +20,8 @@ import {RequestError} from './../transporters';
 import {CredentialRequest, Credentials} from './credentials';
 import {GetTokenResponse, OAuth2Client} from './oauth2client';
 
+export interface ComputeOptions { refreshTokenEarlyMillis?: number; }
+
 export class Compute extends OAuth2Client {
   /**
    * Google Compute Engine metadata server token endpoint.
@@ -33,8 +35,9 @@ export class Compute extends OAuth2Client {
    * Retrieve access token from the metadata server.
    * See: https://developers.google.com/compute/docs/authentication
    */
-  constructor(refreshTokenEarlyMillis?: number) {
-    super({refreshTokenEarlyMillis});
+  constructor(options?: ComputeOptions) {
+    options = options || {};
+    super({refreshTokenEarlyMillis: options.refreshTokenEarlyMillis});
     // Start with an expired refresh token, which will automatically be
     // refreshed before the first API call is made.
     this.credentials = {expiry_date: 1, refresh_token: 'compute-placeholder'};
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -18,9 +18,9 @@ import {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios
 
 import {RequestError} from './../transporters';
 import {CredentialRequest, Credentials} from './credentials';
-import {GetTokenResponse, OAuth2Client} from './oauth2client';
+import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
 
-export interface ComputeOptions { refreshTokenEarlyMillis?: number; }
+export interface ComputeOptions extends RefreshOptions {}
 
 export class Compute extends OAuth2Client {
   /**
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -58,8 +58,6 @@ interface CredentialResult {
   default: {email: string;};
 }
 
-export interface GoogleAuthOptions extends RefreshOptions {}
-
 export class GoogleAuth {
   transporter: Transporter;
 
@@ -90,11 +88,6 @@ export class GoogleAuth {
    * Export DefaultTransporter as a static property of the class.
    */
   static DefaultTransporter = DefaultTransporter;
-
-  constructor(options?: GoogleAuthOptions) {
-    options = options || {};
-    this.eagerRefreshThresholdMillis= options.eagerRefreshThresholdMillis;
-  }
 
   /**
    * Obtains the default project ID for the application..
@@ -236,6 +229,14 @@ export class GoogleAuth {
           'Unexpected error while acquiring application default credentials: ' +
           e.message);
     }
+  }
+
+  /**
+   * Sets time to eagerly refresh tokens before they are expired.
+   * @param {number=} eagerRefreshThresholdMillis Time in ms to refresh an unexpired token before it expires.
+   */
+  setEagerRefreshThresholdMillis(eagerRefreshThresholdMillis: number) {
+    this.eagerRefreshThresholdMillis = eagerRefreshThresholdMillis;
   }
 
   /**
@@ -388,7 +389,8 @@ export class GoogleAuth {
       client = new UserRefreshClient(
           {eagerRefreshThresholdMillis: this.eagerRefreshThresholdMillis});
     } else {
-      client = new JWT({eagerRefreshThresholdMillis: this.eagerRefreshThresholdMillis});
+      client = new JWT(
+          {eagerRefreshThresholdMillis: this.eagerRefreshThresholdMillis});
     }
     client.fromJSON(json);
     return client;
@@ -442,8 +444,8 @@ export class GoogleAuth {
    * @returns A JWT loaded from the key
    */
   fromAPIKey(apiKey: string): JWT {
-    const client =
-        new JWT({eagerRefreshThresholdMillis: this.eagerRefreshThresholdMillis});
+    const client = new JWT(
+        {eagerRefreshThresholdMillis: this.eagerRefreshThresholdMillis});
     client.fromAPIKey(apiKey);
     return client;
   }
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -219,7 +219,8 @@ export class GoogleAuth {
         // TODO: cache the result
         return {
           projectId: null,
-          credential: new Compute(this.refreshTokenEarlyMillis)
+          credential: new Compute(
+              {refreshTokenEarlyMillis: this.refreshTokenEarlyMillis})
         };
       } else {
         // We failed to find the default credentials. Bail out with an error.
@@ -381,7 +382,7 @@ export class GoogleAuth {
     this.jsonContent = json;
     if (json.type === 'authorized_user') {
       client = new UserRefreshClient(
-          undefined, undefined, undefined, this.refreshTokenEarlyMillis);
+          {refreshTokenEarlyMillis: this.refreshTokenEarlyMillis});
     } else {
       client = new JWT({refreshTokenEarlyMillis: this.refreshTokenEarlyMillis});
     }
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -82,8 +82,6 @@ export class GoogleAuth {
 
   cachedCredential: OAuth2Client|null = null;
 
-  private eagerRefreshThresholdMillis: number|undefined;
-
   /**
    * Export DefaultTransporter as a static property of the class.
    */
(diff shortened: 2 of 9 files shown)


src/auth/computeclient.ts
```diff
@@ -33,8 +33,8 @@ export class Compute extends OAuth2Client {
    * Retrieve access token from the metadata server.
    * See: https://developers.google.com/compute/docs/authentication
    */
-  constructor() {
-    super();
+  constructor(refreshTokenEarlyMillis?: number) {
```
ofrobots: Can you make this an options object rather than a positional parameter? Similar to JWT.

nolanmar511: Done -- also did the same for UserRefreshClient.

nolanmar511:
PTAL

src/auth/googleauth.ts
```diff
@@ -87,6 +87,11 @@ export class GoogleAuth {
    */
   static DefaultTransporter = DefaultTransporter;
 
+  /**
+   *  @param {number=} refreshTokenEarlyMillis The token should be refreshed if it will expire within this many milliseconds.
+   */
+  constructor(public refreshTokenEarlyMillis?: number) {}
```
ofrobots: and here?

src/auth/googleauth.ts
```diff
@@ -87,6 +87,11 @@ export class GoogleAuth {
    */
   static DefaultTransporter = DefaultTransporter;
 
+  /**
+   *  @param {number=} refreshTokenEarlyMillis The token should be refreshed if it will expire within this many milliseconds.
```
ofrobots: Omit JSDoc type annotations in TypeScript code

src/auth/computeclient.ts
```diff
@@ -20,6 +20,8 @@ import {RequestError} from './../transporters';
 import {CredentialRequest, Credentials} from './credentials';
 import {GetTokenResponse, OAuth2Client} from './oauth2client';
 
+export interface ComputeOptions { refreshTokenEarlyMillis?: number; }
```
ofrobots: It would be good to add documentation on what the purpose of this configuration option is and how it behaves.

nit: it would be good if this type was not duplicated in three places. Can you see if you can declare a single interface with the common option (just this one) and make others inherit from it.

src/auth/refreshclient.ts
```diff
@@ -32,9 +40,32 @@ export class UserRefreshClient extends OAuth2Client {
    * @param {string} refreshToken The authentication refresh token.
    * @constructor
    */
-  constructor(clientId?: string, clientSecret?: string, refreshToken?: string) {
-    super(clientId, clientSecret);
-    this._refreshToken = refreshToken;
+  constructor(
+      clientId?: string, clientSecret?: string, refreshToken?: string,
+      refreshTokenEarlyMillis?: number);
+
```
ofrobots: nit: omit this newline.

test/test.compute.ts
```diff
@@ -66,6 +66,19 @@ describe('Compute auth client', () => {
     });
   });
 
+  it('should refresh if access token has expired', (done) => {
```
ofrobots: it would good to add another check exercises the other path – namely these case where`refreshTokenEarlyMillis` has been specified but the token should not be expired.

src/auth/googleauth.ts
nolanmar511: done.

src/auth/googleauth.ts
nolanmar511: Done.

src/auth/computeclient.ts
nolanmar511: Done.

src/auth/refreshclient.ts
nolanmar511: Done

test/test.compute.ts
nolanmar511: Done.

src/auth/oauth2client.ts
```diff
@@ -216,14 +216,20 @@ export interface VerifyIdTokenOptions {
...
   clientId?: string;
   clientSecret?: string;
   redirectUri?: string;
   authBaseUrl?: string;
   tokenUrl?: string;
 }
 
+export interface RefreshOptions {
+  // The token should be refreshed if it will expire within this many
+  // milliseconds.
```
ofrobots: Use jsdoc style comments so that this gets emitted into the docs.

nit: `refreshTokenEarlyMillis` doesn't seem obviously descriptive to me. Perhaps you can name it 'eagerRefreshThresholdMillis'?

I would suggest phrasing the comment as something like: "Eagerly refresh unexpired tokens when they are within this many milliseconds from expiring".

ofrobots: nit: the default value should be documented. Right now the default value is buried below in the initialization in the constructor. Is 0 a reasonable default value? /cc @JustinBeckwith

src/auth/computeclient.ts
```diff
@@ -18,7 +18,9 @@ import {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios
 
 import {RequestError} from './../transporters';
 import {CredentialRequest, Credentials} from './credentials';
-import {GetTokenResponse, OAuth2Client} from './oauth2client';
+import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
+
+export interface ComputeOptions extends RefreshOptions {}
```
JustinBeckwith: Do we need to create specific ComputeOptions here?  Why not just use RefreshOptions directly?

src/auth/computeclient.ts
```diff
@@ -33,8 +35,9 @@ export class Compute extends OAuth2Client {
    * Retrieve access token from the metadata server.
    * See: https://developers.google.com/compute/docs/authentication
    */
-  constructor() {
-    super();
+  constructor(options?: ComputeOptions) {
+    options = options || {};
+    super({refreshTokenEarlyMillis: options.refreshTokenEarlyMillis});
```
JustinBeckwith: I think we should make refreshTokenEarlyMillis optional here.

src/auth/googleauth.ts
```diff
@@ -58,6 +58,8 @@ interface CredentialResult {
   default: {email: string;};
 }
 
+export interface GoogleAuthOptions extends RefreshOptions {}
```
JustinBeckwith: Same comment as above, not sure this is worth it

src/auth/googleauth.ts
```diff
@@ -82,11 +84,18 @@ export class GoogleAuth {
...
   cachedCredential: OAuth2Client|null = null;
 
+  private refreshTokenEarlyMillis: number|undefined;
+
   /**
    * Export DefaultTransporter as a static property of the class.
    */
   static DefaultTransporter = DefaultTransporter;
 
+  constructor(options?: GoogleAuthOptions) {
```
JustinBeckwith: I'm hesitant to rely on a constructor with this class.  In the near future, I was hoping to look deeply at #4, #54, and #55.  All of these involve the idea of not requiring the user to instantiate this object, but rather providing default exported methods.  Can we experiment with hanging this option of off the relevant methods?

src/auth/oauth2client.ts
JustinBeckwith: SGTM

src/auth/oauth2client.ts
```diff
@@ -936,4 +932,16 @@ export class OAuth2Client extends AuthClient {
     const buffer = new Buffer(b64String, 'base64');
     return buffer.toString('utf8');
   }
+
+  /**
+   * Returns true iff a token is expired or will expire within
```
JustinBeckwith: Typo `iff`

src/auth/refreshclient.ts
```diff
@@ -32,9 +38,31 @@ export class UserRefreshClient extends OAuth2Client {
    * @param {string} refreshToken The authentication refresh token.
    * @constructor
    */
-  constructor(clientId?: string, clientSecret?: string, refreshToken?: string) {
-    super(clientId, clientSecret);
-    this._refreshToken = refreshToken;
+  constructor(
+      clientId?: string, clientSecret?: string, refreshToken?: string,
+      refreshTokenEarlyMillis?: number);
```
JustinBeckwith: Lets not add this as an option here.  I mostly kept the multiple argument style signature around purely for backwards compatibility.  As we move forward, lets just use the options objects for new properties.

src/auth/refreshclient.ts
```diff
@@ -32,9 +38,31 @@ export class UserRefreshClient extends OAuth2Client {
...
+          clientSecret,
+          refreshToken,
+          refreshTokenEarlyMillis
+        };
+    super({
+      clientId: opts.clientId,
+      clientSecret: opts.clientSecret,
+      refreshTokenEarlyMillis: opts.refreshTokenEarlyMillis
+    });
+    this._refreshToken = opts.refreshToken;
```
JustinBeckwith: I must have missed this class... thank you!

test/test.compute.ts
```diff
@@ -66,6 +66,40 @@ describe('Compute auth client', () => {
...
+         ' before expiration is set',
+     (done) => {
+       nock('http://metadata.google.internal')
+           .get(
+               '/computeMetadata/v1beta1/instance/service-accounts/default/token')
+           .reply(200, {access_token: 'abc123', expires_in: 10000});
+       compute = new Compute({refreshTokenEarlyMillis: 10000});
+       compute.credentials.access_token = 'initial-access-token';
+       compute.credentials.expiry_date = (new Date()).getTime() + 5000;
+       compute.request({url: 'http://foo'}, () => {
```
JustinBeckwith: for new tests please use async/await style syntax

test/test.compute.ts
```diff
@@ -66,6 +66,40 @@ describe('Compute auth client', () => {
...
+         ' refresh before expiration is set',
+     (done) => {
+       const scope =
+           nock('http://metadata.google.internal')
+               .get(
+                   '/computeMetadata/v1beta1/instance/service-accounts/default/token')
+               .reply(200, {access_token: 'abc123', expires_in: 10000});
+       compute.credentials.access_token = 'initial-access-token';
+       compute.credentials.expiry_date = (new Date()).getTime() + 11000;
+       compute.request({url: 'http://foo'}, () => {
```
JustinBeckwith: async/await

test/test.jwt.ts
```diff
@@ -228,6 +228,60 @@ describe('JWT auth client', () => {
...
+         });
+
+         jwt.credentials = {
+           access_token: 'woot',
+           refresh_token: 'jwt-placeholder',
+           expiry_date: (new Date()).getTime() + 800
+         };
+
+         createGTokenMock({access_token: 'abc123'});
+         jwt.request({url: 'http://bar'}, () => {
```
JustinBeckwith: async/await

test/test.jwt.ts
```diff
@@ -228,6 +228,60 @@ describe('JWT auth client', () => {
...
+           refreshTokenEarlyMillis: 1000
+         });
+
+         jwt.credentials = {
+           access_token: 'initial-access-token',
+           refresh_token: 'jwt-placeholder',
+           expiry_date: (new Date()).getTime() + 5000
+         };
+
+         jwt.request({url: 'http://bar'}, () => {
```
JustinBeckwith: will stop calling these out now

src/auth/oauth2client.ts
```diff
@@ -936,4 +932,16 @@ export class OAuth2Client extends AuthClient {
...
+   * refreshTokenEarlyMillis milliseconds.
+   * If there is no expiry time, assumes the token is not expired or expiring.
+   */
+  protected isTokenExpiring(): boolean {
```
JustinBeckwith: So here's a funny question.  Does `refreshTokenEarlyMillis` really need to be configurable?  I had just submitted #248, which sorta blindly adds a 1 second buffer.  Why do we need to make this configurable?  cc @ofrobots @jonparrott

JustinBeckwith:
Does this address #67?

src/auth/oauth2client.ts
ofrobots: 1 second may not be enough. For example Cloud Profiler may keep a request open for 1 hour (!). The profiler agent library wants to use 1hr as the timeout to make sure that a token expiring isn't likely going to fail.

src/auth/oauth2client.ts
JustinBeckwith: Given the issues raised in #62 I would rather see this defaulted to at least 1 second.

src/auth/oauth2client.ts
JustinBeckwith: Fair point :)   I do think we should have a reasonable default here, to cut off the issues in #62.

src/auth/oauth2client.ts
nolanmar511: Renamed, updated the comment, and increased the default value 1 second.

src/auth/computeclient.ts
nolanmar511: Done.

src/auth/computeclient.ts
nolanmar511: Done.

src/auth/googleauth.ts
nolanmar511: Done.

src/auth/googleauth.ts
nolanmar511: I added a setter for `refreshTokenEarlyMillis`

src/auth/computeclient.ts
ofrobots: I see value in a separate type in case `ComputeOptions` adds more properties in the future. Having an empty type alias here avoids a potential breaking change in the future.

src/auth/googleauth.ts
ofrobots: Same. I'd prefer to keep it.

src/auth/oauth2client.ts
nolanmar511: Meant "if and only if", but changed to "if".

src/auth/refreshclient.ts
nolanmar511: Done.

test/test.compute.ts
nolanmar511: I believe that this is an easy change in test.compute.ts or in test.jwt.ts.

When I re-wrote this test as 
```
  it('should not refresh if access token will not expire soon and time to' +
         ' refresh before expiration is set',
     async () => {
       const scope =
           nock('http://metadata.google.internal')
               .get(
                   '/computeMetadata/v1beta1/instance/service-accounts/default/token')
               .reply(200, {access_token: 'abc123', expires_in: 10000});
       compute.credentials.access_token = 'initial-access-token';
       compute.credentials.expiry_date = (new Date()).getTime() + 11000;
       await compute.request({url: 'http://foo'});
       assert.equal(compute.credentials.access_token, 'initial-access-token');
       nock.cleanAll();
     });
```

it failed with:
```
NetConnectNotAllowedError: Nock: Not allow net connect for "foo:80/"
      at node_modules/nock/lib/intercept.js:404:21
      at Object.module.request (node_modules/nock/lib/common.js:140:14)
      at RedirectableRequest._performRequest (node_modules/follow-redirects/index.js:73:20)
      at new RedirectableRequest (node_modules/follow-redirects/index.js:57:7)
      at Object.wrappedProtocol.request (node_modules/follow-redirects/index.js:229:10)
      at dispatchHttpRequest (node_modules/axios/lib/adapters/http.js:132:25)
      at new Promise (<anonymous>)
      at httpAdapter (node_modules/axios/lib/adapters/http.js:18:10)
      at dispatchRequest (node_modules/axios/lib/core/dispatchRequest.js:59:10)
      at <anonymous>
```

I have changed tests in test.oauth2.ts

nolanmar511:
This change does not address #67, though I assume that it would make errors like those in #67 less common.

jonparrott:
Somehow the comment about expiry buffer got lost in the fold, but it should be 5 minutes.

See the canonical implementation:
https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/dcbf1e0af0c79f1424f03ad27b071eccfdbc80d3/google/auth/credentials.py#L62
https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/dcbf1e0af0c79f1424f03ad27b071eccfdbc80d3/google/auth/_helpers.py#L25

src/auth/googleauth.ts
ofrobots: I am not very thrilled with this turn of events. I would prefer to keep this to be a construction time option, and not a mutable property.

I would like to understand the needs from the linked issues more deeply before we do a lot more in this direction.

JustinBeckwith: @ofrobots and I had a good discussion here.  I think we're both in favor of accepting optional options in the public factory methods on the class.  This has a few benefits:
- We can get rid of the constructor requirement for the `GoogleAuth` class
- We can still make the expiry buffer configurable
- It's still a non-breaking change
- There are no global ramifications 

I think that means extending these methods:
`getApplicationDefault`, `fromJSON`, `fromStream`

Thank you for bearing with us here :)

src/auth/computeclient.ts
nolanmar511: @ofrobots, @JustinBeckwith -- What's the decision here?
I'd originally added the separate types for the reasons ofrobots suggested, but I have no strong opinions here.

JustinBeckwith: I am happy to be outvoted here :)  I don't feel terribly strongly either.

src/auth/googleauth.ts
nolanmar511: For getApplicationDefaultAsync, should I modify the value for `eagerRefreshThresholdMillis` if cached credentials were used?
Right now I do not.

nolanmar511:
PTAL

JustinBeckwith:
LGTM!  Please rebase to master, and I'll merge this in. Thank you!

nolanmar511:
Just rebased against master.

tnguyen14:
Is there any update to the docs? How should this change be used?

JustinBeckwith:
Greetings!  You can learn how to us this in the reference docs:
https://google.github.io/google-auth-library-nodejs/classes/_auth_googleauth_.googleauth.html#getapplicationdefault

rublev:
> Greetings! You can learn how to us this in the reference docs:
> https://google.github.io/google-auth-library-nodejs/classes/_auth_googleauth_.googleauth.html#getapplicationdefault

404

PR #243, opened 2018-01-10
Fix usage of OAuth2Client.verifyIdToken in example

aknuds1:
examples/verifyIdToken.js uses the old signature for `OAuth2Client.verifyIdToken`, this PR upgrades it to the new API.

### DIFF
--- a/examples/verifyIdToken.js
+++ b/examples/verifyIdToken.js
@@ -30,10 +30,10 @@ async function main() {
     const oAuth2Client = await getAuthenticatedClient();
 
     // Verify the id_token, and access the claims.
-    const ticket = await oAuth2Client.verifyIdToken(
-      oAuth2Client.credentials.id_token,
-      keys.web.client_id
-    );
+    const ticket = await oAuth2Client.verifyIdToken({
+      idToken: oAuth2Client.credentials.id_token,
+      audience: keys.web.client_id,
+    });
     console.log(ticket);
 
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
Thank you for catching this :)  Could you also add the audience in the new options?  

```js
const ticket = await oAuth2Client.verifyIdToken({
  idToken: oAuth2Client.credentials.id_token,
  audience: keys.web.client_id
});
```

aknuds1:
@JustinBeckwith I added a part `audience: 'audience'` to the `verifyIdToken` options, as I don't know would be a sensible value for `audience`. Does this look OK to you or should I use some other value?

aknuds1:
@JustinBeckwith I'm also wondering if `verifyIdToken` should be documented somewhere? I only find it in examples and tests.

JustinBeckwith:
The audience should be set to `keys.web.client_id` :) On the docs piece - the readme can always be better, but at least we have a proper reference doc now at google.github.io/google-auth-library-nodejs/

aknuds1:
@JustinBeckwith Oh, alright, I've updated the PR. The `verifyIdToken` test (test.oauth2.ts) doesn't follow this pattern though, should it be changed to read `const result = await client.verifyIdToken({idToken, audience: CLIENT_ID, maxExpiry});` to be consistent?

Boboss74:
omg this is why it doesn't work... T-T

PR #248, opened 2018-01-11
fix: add buffer for expiry_date

### LINKED ISSUE #62: low error margin on expiry_date => isTokenExpired false negatives
My app saves refresh tokens in long-term storage and caches the transient access tokens. I make calls to to the google drive api, polling the changes endpoint in ~1 minute intervals.

Every once in a while (~24 hours), I get a 401 Invalid Credentials error. The refresh token is definitely fine - it hasn't expired or been revoked, e.g. using the same refresh token a minute later (on the next poll cycle), it again continues to work again, despite the auth failure in the previous poll cycle.

Because it occurs pretty randomly, I think this issue originates from the following lines in the auth client library:

``` javascript
tokens.expiry_date = ((new Date()).getTime() + (tokens.expires_in * 1000));
```

``` javascript
var isTokenExpired = expiryDate ? expiryDate <= (new Date()).getTime() : false;
```

If there's any significant network latency in the remote calls, then the true expiry date will be earlier than the computed expiry_date, and isTokenExpired may be set to false even when it should really be true. This would result in an API call with the expired access token, leading to a 401 Invalid Credentials error.

In my specific case, it would also makes sense in that on the next poll cycle, we would have passed the computed expiry_date, and the refresh token is then used to get a new access token, and the system continues to work fine.

Let me know if I'm totally off base, and thanks for this great client library!

JustinBeckwith:
Fixes #62

### DIFF
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -69,8 +69,7 @@ export class Compute extends OAuth2Client {
     }
     const tokens = res.data as Credentials;
     if (res.data && res.data.expires_in) {
-      tokens.expiry_date =
-          ((new Date()).getTime() + (res.data.expires_in * 1000));
+      tokens.expiry_date = this.getExpirationDate(res.data.expires_in);
       delete (tokens as CredentialRequest).expires_in;
     }
     return {tokens, res};
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -404,8 +404,7 @@ export class OAuth2Client extends AuthClient {
     });
     const tokens = res.data as Credentials;
     if (res.data && res.data.expires_in) {
-      tokens.expiry_date =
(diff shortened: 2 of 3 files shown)


ofrobots:
How does this related to #242 ?

JustinBeckwith:
A fair question that I also asked over there :)  I should have reviewed that before submitting this one.  A few differences:
- That PR is making the refresh window configurable, while this PR just sets it to 1 second to account for the potential request time.  I'm not sure why we would need it to be configurable, or if 1 second is enough time. 
- I am not doing any additional automatic refreshing, that PR is

ofrobots:
Replied here: https://github.com/google/google-auth-library-nodejs/pull/242#discussion_r160865940. There is value in users being able to specify the threshold.

JustinBeckwith:
Closing in favor of #242

PR #249, opened 2018-01-12
feat: export static GoogleAuth

JustinBeckwith:
Addresses #4

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import {GoogleAuth} from './auth/googleauth';
+
 export {Compute} from './auth/computeclient';
 export {GoogleAuth} from './auth/googleauth';
 export {IAMAuth} from './auth/iam';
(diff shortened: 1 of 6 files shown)


JustinBeckwith:
My initial approach here was to try using es6 default exports:

```js
export default new GoogleAuth();
```

However - given the lack of compat between CommonJS and es6 modules, consumers would have to use this syntax:

```js
const auth = require('google-auth-library').default;
```

That doesn't feel terribly discoverable, and isn't much better from what we have today.  Webpack works around this, but apparently TypeScript isn't really planning on muddying the waters:
https://github.com/Microsoft/TypeScript/issues/2719

Instead, I ended up taking the named export approach.  I export both the `GoogleAuth` class, AND an `auth` object with a single instance.  It means consumers using Node 6 and above can do this:

```js
const {auth} = require('google-auth-library');
auth.getApplicationDefault()...
```

The `module.exports` and `exports =` syntax aren't terribly es6 compatible, and would have caused a whole other set of issues.

This seems like a fair compromise, but I'm looking for opinions :)

JustinBeckwith:
@ofrobots and @mylesborins, would love to hear your thoughts too :)

src/index.ts
```diff
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import {GoogleAuth} from './auth/googleauth';
+
 export {Compute} from './auth/computeclient';
 export {GoogleAuth} from './auth/googleauth';
```
ofrobots: nit: avoid second reference to './auth/googleauth'. Export in the newly added stanza below.

Issue #234, opened 2018-01-05
Changelog outdated


zoellner:
Changelog is outdated. last documented change is from 12/29/2016
However there are changes in this library that introduce breaking changes in google/google-api-nodejs-client. see google/google-api-nodejs-client#894

JustinBeckwith:
Apologies.  This was an oversight, and was addressed in #198.  It will be part of the 1.0 release, and you can track that over in #197.

JustinBeckwith:
This was fixed in the 1.0 release, and is now updated in https://github.com/google/google-api-nodejs-client/pull/891

nopnop:
Hi @JustinBeckwith, maybe @zoellner issue was related to this file https://github.com/google/google-auth-library-nodejs/blob/master/CHANGELOG.md which is still outdated.
It is a bit confusing at first, found this issue, then ... I eventually scrolled the README.md

zoellner:
that's correct. the CHANGELOG.md file is what I stumbled across when I was debugging that issue.

JustinBeckwith:
Ah, that makes a lot more sense :)  That's an easy change.  Will submit a PR soon!

JustinBeckwith:
The change log has now been entirely moved to GitHub release notes, and it includes these changes :) Thanks folks!

PR #250, opened 2018-01-15
chore: update changelog

JustinBeckwith:
Addresses #234.  I don't like keeping this in two places, so the alternative approach is to delete the file all together to reduce confusion and just rely on the "Releases" feature in GitHub.

@ofrobots thoughts?


JustinBeckwith:
Decided to backfill all the releases (including new tags), and delete the file :)   Thanks!

Issue #52, opened 2015-08-10
Jwt Access Credentials should cache JWTs


anthmgoogle:
The JWT Access Credentials should temporarily cache JWTs. Suggested algorithm:
- Hash JWTs using Audience as key.
- Also store the timestamp of last use with each JWT.
- On access clear any JWT unused for more than 1 hour.

JustinBeckwith:
Out of curiosity, @jonparrott / @lukesneeringer is this something you do in the python client?

jonparrott:
Yes, see https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/google/auth/jwt.py#L497

JustinBeckwith:
Interesting!  Out of curiosity - what are the use cases for the configurable cache size and expiration window?

jonparrott:
The cache size and expiration window are not configurable for Python. It's rare for an application to talk to any memory-significant number audiences with the same set of credentials, and the expiration time is just the same as the JWT expiration time.

JustinBeckwith:
Addressed in #254, will go out in the upcoming 1.1.0 release.

PR #254, opened 2018-01-16
feat: cache JWT tokens in JWTAccess

JustinBeckwith:
addresses #52


JustinBeckwith:
ping @google/google-node-team :) . Also @anthmgoogle and @jonparrott if you want to take a look.

src/auth/jwtaccess.ts
```diff
@@ -71,10 +82,24 @@ export class JWTAccess {
...
-    return {headers: {Authorization: 'Bearer ' + signedJWT}};
+    const res = {headers: {Authorization: 'Bearer ' + signedJWT}};
+    this.cache.set(authURI, {res, expires: Date.now() + 60 * 60 * 1000});
+    return res;
+  }
+
+  /**
+   * Delete any token in the cache which has expired.
+   */
+  private cleanupCache() {
```
jinwoo: `expireCache()` seems a better name...

src/auth/jwtaccess.ts
```diff
@@ -71,10 +82,24 @@ export class JWTAccess {
...
+    const now = Date.now();
+    this.cache.forEach((value, key) => {
```
jinwoo:
```typescript
this.cache.forEach(({expires}, key) => {
  if (now > expires) {
    ...
  }
});
```
?

test/test.jwtaccess.ts
```diff
@@ -51,6 +60,25 @@ describe('.getRequestMetadata', () => {
...
+
+  it('should not return cached tokens older than an hour', () => {
+    const realDateNow = Date.now;
+    try {
+      // go forward in time one hour (plus a little)
+      Date.now = () => realDateNow() + (1000 * 60 * 60) + 10;
+      const res2 = client.getRequestMetadata(testUri);
+      assert.notEqual(res, res2);
+    } catch (e) {
+      assert.fail(e);
```
jinwoo: If anything throws, doesn't the test fail automatically? I think you can just delete the `try-catch` from here.

src/auth/jwtaccess.ts
```diff
@@ -51,13 +58,17 @@ export class JWTAccess {
...
-   * Get a non-expired access token, after refreshing if necessary
+   * Get a non-expired access token, after refreshing if necessary.
    *
    * @param {string} authURI the URI being authorized
-   * @param {function} metadataCb a callback invoked with the jwt request metadata.
-   * @returns a Promise that resolves with the request metadata response
+   * @returns An object that includes the authorization header.
    */
   getRequestMetadata(authURI: string): RequestMetadataResponse {
+    this.cleanupCache();
```
jonparrott: This seems unnecessary, just check the expiration on the key you get from the cache and drop it if needed. O(1) vs O(n).

(This makes me yearn for Node.js to have a LRU Cache in it's "standard library")

test/test.jwtaccess.ts
JustinBeckwith: I did that to make sure that we return Date.now back to the original function.

jinwoo: How does it work if you call `assert.fail(e)` on exceptions?

Did you mean to do
```typescript
   const realDateNow = Date.now;
   try {
     // go forward in time one hour (plus a little)
     Date.now = () => realDateNow() + (1000 * 60 * 60) + 10;
     const res2 = client.getRequestMetadata(testUri);
     assert.notEqual(res, res2);
   } finally {
     // return date.now to it's normally scheduled programming
     Date.now = realDateNow;
   }
```
?

JustinBeckwith: Fair point :)

src/auth/jwtaccess.ts
JustinBeckwith: The danger there is that the cache can grow over time to a large size, leading to leaks.  I am taking a shot at using the canonical LRU cache module.

JustinBeckwith:
@jinwoo I made a pretty substantial change, so just wanted to make sure this was better :)

src/auth/jwtaccess.ts
```diff
@@ -24,6 +25,9 @@ export class JWTAccess {
   key?: string|null;
   projectId?: string;
 
+  private cache =
```
jinwoo: Same preference of mine as before :)

```typescript
private cache: LRU<string, RequestMetadataResponse> = LRU({max: 500, maxAge: 60 * 60 * 1000});
```

But up to you!

src/auth/jwtaccess.ts
```diff
@@ -71,10 +78,12 @@ export class JWTAccess {
       secret: this.key
     };
 
-    // Sign the jwt and invoke metadataCb with it.
+    // Sign the jwt and add it to the cache
     const signedJWT =
         jws.sign({header: {alg: 'RS256'}, payload, secret: this.key});
-    return {headers: {Authorization: 'Bearer ' + signedJWT}};
+    const res = {headers: {Authorization: 'Bearer ' + signedJWT}};
```
ofrobots: nit: as long as you're changing this, switch to using a template string instead of using concatenation.

src/auth/jwtaccess.ts
```diff
@@ -24,6 +25,9 @@ export class JWTAccess {
   key?: string|null;
   projectId?: string;
 
+  private cache =
+      LRU<string, RequestMetadataResponse>({max: 500, maxAge: 60 * 60 * 1000});
```
ofrobots: question: why is 500 a reasonable number here?

src/auth/jwtaccess.ts
JustinBeckwith: fixed!

src/auth/jwtaccess.ts
JustinBeckwith: I will admit to choosing this at complete random.  Totally open to a better number :)

PR #255, opened 2018-01-17
chore: delete createScopedRequired

JustinBeckwith:
**This may be a terrible idea, and it's possible if not likely I'm missing something very important**

Now with that out of the way...  addresses #55.  I was reading through that issue, and looking more closely at the code when I realized...  why does this whole mess even exist?  

I think we can reasonably just delete createScoped, and createdScopedRequired.  This of course would be a semver major, and it's possible it's just not worth it right now.  But I do want to get this in front of folks and find out what y'all think...  or more importantly why this is a bad idea :)

cc @stephenplusplus @lukesneeringer @alexander-fenster


README.md
```diff
@@ -52,44 +52,26 @@ Rather than manually creating an OAuth2 client, JWT client, or Compute client, t
...
-    client = client.createScoped(scopes);
-  }
-  return {
-    client: client,
-    projectId: res.projectId
-  }
+  const client = res.credential;
+  client.scopes = ['https://www.googleapis.com/auth/cloud-platform'];
+  const url = `https://www.googleapis.com/dns/v1/projects/${res.projectId}`;
+  const res2 = await client.request({ url });
```
ofrobots: nit: Let's rename the variables to avoid the name collision. It would be fine to call this one `res`, but I would suggest renaming the `res` above to something more descriptive.

examples/adc.js
```diff
@@ -28,35 +28,14 @@ const { auth } = require('google-auth-library');
...
-    client = client.createScoped(scopes);
-  }
-  return {
-    client: client,
-    projectId: res.projectId
-  };
+  const client = res.credential;
+  client.scopes = ['https://www.googleapis.com/auth/cloud-platform'];
+  const url = `https://www.googleapis.com/dns/v1/projects/${res.projectId}`;
+  const res2 = await client.request({ url });
```
ofrobots: nit: ditto

examples/verifyIdToken.js
```diff
@@ -32,7 +32,7 @@ async function main() {
     // Verify the id_token, and access the claims.
     const ticket = await oAuth2Client.verifyIdToken({
       idToken: oAuth2Client.credentials.id_token,
-      audience: keys.web.client_id,
+      audience: keys.web.client_id
```
ofrobots: Did you manually make this change or was this clang-format? Seems like an unnecessary change, but perhaps the version of clang-format has changed?

README.md
```diff
@@ -52,44 +52,26 @@ Rather than manually creating an OAuth2 client, JWT client, or Compute client, t
...
-    // Scopes can be specified either as an array or as a single, space-delimited string.
-    const scopes = ['https://www.googleapis.com/auth/cloud-platform'];
-    client = client.createScoped(scopes);
-  }
-  return {
-    client: client,
-    projectId: res.projectId
-  }
+  const client = res.credential;
+  client.scopes = ['https://www.googleapis.com/auth/cloud-platform'];
```
ofrobots: I would prefer that the scopes (if any) be provided to the constructor/factory rather than being attached to a properly like hence. Would that work?

examples/verifyIdToken.js
JustinBeckwith: This is prettier.  I don't have prettier linting blocking npm test right now.  I will go fix that!

JustinBeckwith:
I'm going to go ahead and close this one out for now.  I opened up #264 to track it, and I figure we can come back in a few months when 2.0 is on the horizon :)

PR #257, opened 2018-01-17
chore: move to CircleCI and fix codecov

JustinBeckwith:
This uses js-green-licenses as an example, and moves us over to CircleCI. Feedback welcomed!

cc @alexander-fenster


.circleci/config.yml
```diff
@@ -0,0 +1,86 @@
...
+
+version: 2.0
+workflows:
+  version: 2
+  tests:
+    jobs:
+      - node4:
+          filters:
+            tags:
+              only: /^v\d+(\.\d+){2}(-.*)?$/
```
jinwoo: I learned a trick from @kjin for defining this regexp only once. https://github.com/google/node-h2-auto-push/blob/9b09ae27c31a637ae32a23c05957944a5d530ef6/.circleci/config.yml#L2

.circleci/config.yml
```diff
@@ -0,0 +1,86 @@
...
+
+  publish_npm:
+    docker:
+      - image: node:8
+        user: node
+    steps:
+      - checkout
+      - run:
+          name: Set npm authentication.
+          command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
```
jinwoo: If you want npm auto-publish on version tags, you need to add the NPM_TOKEN env var to the CircleCI settings. The token is in valentine.

test/test.googleauth.ts
```diff
@@ -1176,7 +1176,8 @@ describe('.getApplicationDefault', () => {
        // tslint:disable-next-line no-any
        assert.equal(null, (result3 as any).specialTestBit);
        assert.notEqual(cachedCredential, result3);
-     });
+     })
+      .timeout(5000);
```
jinwoo: Curious. Why is this needed?

.circleci/config.yml
alexander-fenster: Why do we need to filter by tag for these tasks at all?

.circleci/config.yml
```diff
@@ -0,0 +1,86 @@
...
+      - run:
+          name: Install modules and dependencies.
+          command: npm install
+      - run:
+          name: Publish the module to npm.
+          command: npm publish
```
alexander-fenster: To make CI run for all commits, but publish only for v.* tags, all tasks like node4, node6, etc. should have a filter /.*/, but the publish task should be restricted to v.*.  See example here: https://github.com/googleapis/nodejs-os-login/blob/master/.circleci/config.yml - we run tests on all versions for all commits, run lint and docs for all commits, but run publish only on v* tags.

.circleci/config.yml
jinwoo: This is so that these tests also run for those tags as well as changes in branches. That's needed because `node4`, `node6`, etc. are requirements of `publish_npm`, which is triggered only for such tags.

CircleCI build is performed for branches by default, but is not for tags by default.

.circleci/config.yml
jinwoo: No. The build runs for branches by default, so you don't need a filter `/.*/`. `node-h2-auto-push` has the same configuration and npm publish works fine: https://github.com/google/node-h2-auto-push/blob/master/.circleci/config.yml

alexander-fenster:
Also, in circleci -> project settings -> Advanced settings, enable builds for forked repos, that way, you'll get CI run for this PR as well.

.circleci/config.yml
jinwoo: See this doc: https://github.com/google/node-h2-auto-push/blob/master/.circleci/config.yml

jinwoo: Oops. Sorry. See this doc: https://circleci.com/docs/2.0/workflows/#git-tag-job-execution

.circleci/config.yml
alexander-fenster: hmm. we have different setup in our other repos (see my comment below)

.circleci/config.yml
alexander-fenster: hmm. interesting
probably it's because we also have extra tasks (lint, docs, and system and sample tests for some repos) that depend on the branches?

alexander-fenster: so what happens in our case is, {node4,node6,node7,node8,node9} -> {lint,docs} runs for all commits, but {node4,node6,node7,node8,node9} -> {lint,docs} -> publish_npm runs for tagged only, hence we probably need to have a filter of /.*/ on all tasks and a filter of /v.*/ on publish_npm.

test/test.googleauth.ts
JustinBeckwith: This test will randomly time out from time to time.  I am dropping it from this PR and doing another.

.circleci/config.yml
JustinBeckwith: Woah this is rad.  Thanks!

JustinBeckwith:
@jinwoo I mentioned green licenses because I copied your config.yml :)

.circleci/config.yml
JustinBeckwith: 👍 will do this as soon as this lands

JustinBeckwith: Which secret in valentine am I looking for?  Can you make sure I have access?

jinwoo: It's stored as "google-node-team@npm's NPM_TOKEN". I think you have access. Let me know if you don't.

PR #260, opened 2018-01-19
Add given_name and family_name to TokenPayload interface

dhritzkiv:
I can't find the documentation that lists this field, but it is returned by the API.

### DIFF
--- a/src/auth/loginticket.ts
+++ b/src/auth/loginticket.ts
@@ -133,6 +133,24 @@ export interface TokenPayload {
    * records. Note that this claim is never guaranteed to be present.
    */
   name?: string;
+  
+  /**
+   * The user's given name, in a displayable form. Might be provided when:
+   * - The request scope included the string "profile"
+   * - The ID token is returned from a token refresh
+   * - When name claims are present, you can use them to update your app's user
+   * records. Note that this claim is never guaranteed to be present.
+   */
+  given_name?: string;
+  
+  /**
+   * The user's family name, in a displayable form. Might be provided when:
+   * - The request scope included the string "profile"
+   * - The ID token is returned from a token refresh
+   * - When name claims are present, you can use them to update your app's user
+   * records. Note that this claim is never guaranteed to be present.
+   */
+  family_name?: string;
 
   /**
    * Identifies the audience that this ID token is intended for. It must be one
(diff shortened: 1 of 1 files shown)


dhritzkiv:
Not sure why the tests are failing. What is gts check and what about my additions cause it to fail?

JustinBeckwith:
Thanks for the PR!  Silly question - how do you end up in a situation where you're given a token with that claim defined?  The list here was largely drawn from this:
https://developers.google.com/identity/protocols/OpenIDConnect

Do you have an example of code that uses this?

dhritzkiv:
Hmm. Curious that it's not on there.

When writing the PR, I referenced the response from the Google API Explorer, specifically this: https://developers.google.com/apis-explorer#p/oauth2/v2/oauth2.userinfo.v2.me.get

The response for my account was:

```json
{
 "id": "108219257014879587292",
 "email": "daniel@matterandform.net",
 "verified_email": true,
 "name": "Daniel Hritzkiv",
 "given_name": "Daniel",
 "family_name": "Hritzkiv",
 "picture": "https://lh3.googleusercontent.com/-UJs4r4xP9Lc/AAAAAAAAAAI/AAAAAAAAACw/2uuOHxpjfwk/photo.jpg",
 "locale": "en",
 "hd": "matterandform.net"
}
```

I was able to get `given_name` and `family_name` using just the scope `"https://www.googleapis.com/auth/userinfo.profile"`

I actually haven't tested/confirmed that that's what is returned with this library, but I imagine it is.

JustinBeckwith:
Thanks for the details here!  I'm still just a tad turned around - can you share the code snippet you're trying write, that you can't write today?

dhritzkiv:
It’s not that I can’t write a bit of code, it’s more-so that the typescript definitions (that are automatically generated from this typescript source file) currently suggest to my IDE that given_name is not accessible from objects of this shape. Fortunately for me, my code still runs fine (since I’m using vanilla JS), but it might fail to compile if someone were using TS.

In any case, here is a sample snippet of code:

‘’’js
const payload = login.getPayload();//payload is of type TokenPayload

const {
    given_name,
    family_name
} = payload;//warnings in my IDE about these fields not existing on the TokenPayload interface
‘’’

Hope I understood your question!

PR #266, opened 2018-01-23
chore: update to latest version of gtoken

JustinBeckwith:


test/test.jwt.ts
```diff
@@ -49,11 +49,9 @@ interface GTokenResult {
   expires_in?: number;
 }
 
-function createGTokenMock(body: GTokenResult = {
-  access_token: 'initial-access-token'
-}) {
-  nock('https://accounts.google.com:443')
-      .post('/o/oauth2/token')
+function createGTokenMock(body: GTokenResult) {
+  return nock('https://www.googleapis.com')
```
jinwoo: Hmm.... So it looks like gtoken 2.1.0 was a breaking change for at least one of its clients :) I'm not sure how many projects will have broken tests once they do a fresh `npm install`...

Not sure if we have to release 3.0.0 or just live with this.

JustinBeckwith: I'm not sure how to feel about this one.  It's a breaking change in that the unit tests were reaching inside of the module, and mocking a URL that's not part of the public contract.  I could reasonably argue this one either way. 

@ofrobots this is an interesting discussion :)  What do you think?

ofrobots: I don't think this was a breaking change. The only thing effected is _tests_ that were reaching into internals.

PR #268, opened 2018-01-25
feat: add IAP and additional claims support

### LINKED ISSUE #134: Allow custom claims in JWTAccess `payload`
[Google Cloud IAP](https://cloud.google.com/iap) provides the ability to use bearer tokens as a means of authentication. However, a separate claim needs to be provided (`target_audience`) to do this. The current code in lib/auth/jwtaccess.js can do this but needs support for custom claims.

JustinBeckwith:
Adds support for additionalClaims to JWT, and an example of using it to work with the Identity Aware Proxy (IAP). Resolves #134  🤘

### DIFF
--- a/src/auth/jwtaccess.ts
+++ b/src/auth/jwtaccess.ts
@@ -18,7 +18,7 @@ import * as jws from 'jws';
 import * as LRU from 'lru-cache';
 import * as stream from 'stream';
 import {JWTInput} from './credentials';
-import {RequestMetadataCallback, RequestMetadataResponse} from './oauth2client';
+import {RequestMetadataResponse} from './oauth2client';
 
 export class JWTAccess {
   email?: string|null;
@@ -74,6 +74,8 @@ export class JWTAccess {
     // The payload used for signed JWT headers has:
     // iss == sub == <client email>
     // aud == <the authorization uri>
+    // NOTE: Users can also override the values for iss / sub / aud / etc
+    // with additionalClaims.
     const payload = Object.assign(
         {iss: this.email, sub: this.email, aud: authURI, exp, iat},
         additionalClaims);
--- a/src/auth/jwtaccess.ts
+++ b/src/auth/jwtaccess.ts
@@ -57,10 +57,13 @@ export class JWTAccess {
   /**
    * Get a non-expired access token, after refreshing if necessary.
    *
-   * @param {string} authURI the URI being authorized
+   * @param authURI The URI being authorized.
+   * @param additionalClaims An object with a set of additional claims to
+   * include in the payload.
    * @returns An object that includes the authorization header.
    */
-  getRequestMetadata(authURI: string): RequestMetadataResponse {
+  getRequestMetadata(authURI: string, additionalClaims?: {}):
+      RequestMetadataResponse {
     const cachedToken = this.cache.get(authURI);
     if (cachedToken) {
       return cachedToken;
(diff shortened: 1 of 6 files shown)


src/auth/jwtclient.ts
```diff
@@ -93,11 +98,26 @@ export class JWT extends OAuth2Client {
...
+        // /oauth2/v4/token endpoint and specifically IAP. If that claim is
+        // included, the token endpoint will return an id_token instead of
+        // an access_token.
+        const {tokens} = await this.refreshToken();
+        return {headers: {Authorization: `Bearer ${tokens.access_token}`}};
+      } else {
+        // TODO: There is some duplication between the code in google/node-gtoken
+        // and the code in JWTAccess.  This particular code path needs to be
+        // further rationalized.
+        if (!this.access) {
```
kjin: Why does this value now have to be cached? (Is that what you meant by 'needs to be further rationalized'?)

JustinBeckwith: Hah, that's a great question.  So in #254 I added caching inside of the JWTAccess object for gtokens.  However - we create a new instance of that object every time `getRequestMetadata` is called, so the cache does nothing 😂 . This actually enables the cache.

src/auth/jwtaccess.ts
```diff
@@ -71,12 +74,9 @@ export class JWTAccess {
...
     // aud == <the authorization uri>
-    const payload = {iss: this.email, sub: this.email, aud: authURI, exp, iat};
-    const assertion = {
-      header: {alg: 'RS256'} as jws.Header,
-      payload,
-      secret: this.key
-    };
+    const payload = Object.assign(
+        {iss: this.email, sub: this.email, aud: authURI, exp, iat},
+        additionalClaims);
```
ofrobots: What's the behavior you want to have when the user has a `iss` property in `additionalClaims`? Are they allowed to override `this.email`?

I think you should decide on a behavior, document it, and add some tests.

JustinBeckwith: Decided to support overriding, and add a test specifically for it.  Good call 👍

src/auth/jwtaccess.ts
```diff
@@ -71,12 +74,11 @@ export class JWTAccess {
...
     // iss == sub == <client email>
     // aud == <the authorization uri>
-    const payload = {iss: this.email, sub: this.email, aud: authURI, exp, iat};
-    const assertion = {
-      header: {alg: 'RS256'} as jws.Header,
-      payload,
-      secret: this.key
-    };
+    // NOTE: Users can also override the values for iss / sub / aud / etc
+    // with additionalClaims.
```
ofrobots: This worries me somewhat. Are these potentially overridable properties documented anywhere? I am worried about accidental overriding of these properties because the user got the `additionalClaims` object from somewhere else and passed it on. It might lead to spurious errors. We even have strong types here to avoid the trap. `additionalClaims` is simply an `any`.

JustinBeckwith: You're right, and this is a really good catch.  I updated the code to validate that users don't pass any of the pre-populated properties, and changed the test to do the opposite.  PTAL!

src/auth/jwtaccess.ts
JustinBeckwith: Reverse that, see comment below 😂

JustinBeckwith:
FYI @matthewg

JustinBeckwith:
@ofrobots *ping* :)

src/auth/jwtaccess.ts
```diff
@@ -57,26 +57,38 @@ export class JWTAccess {
...
     if (cachedToken) {
       return cachedToken;
     }
     const iat = Math.floor(new Date().getTime() / 1000);
     const exp = iat + 3600;  // 3600 seconds = 1 hour
 
+    // if additionalClaims are provided, ensure they do not collide with
+    // other required claims.
+    if (additionalClaims) {
+      ['iss', 'sub', 'aud', 'exp', 'iat'].forEach(claim => {
```
ofrobots: nit: If a new property is added to the inline object at line 90, someone will need to remember to come back to this list and update it. Instead move the object at line 90 to a named object above, and then iterate the keys of the object to make sure none of them are in additional claims.

JustinBeckwith: That is awesome feedback.

PR #269, opened 2018-01-30
Update token endpoint + fix failing tests

ace-n:
When I try to use this library in an `npm` Container Builder step (via the [Cloud Storage API](https://cloud.google.com/storage/docs)), I get the following error:
```
{ Error: 404 page not found
at new RequestError (/workspace/functions/autodeploy/node_modules/google-auth-library/lib/transporters.js:34:42)
at Request._callback (/workspace/functions/autodeploy/node_modules/google-auth-library/lib/transporters.js:108:23)
at Request.self.callback (/workspace/functions/autodeploy/node_modules/request/request.js:186:22)
at emitTwo (events.js:126:13)
at Request.emit (events.js:214:7)
at Request.<anonymous> (/workspace/functions/autodeploy/node_modules/request/request.js:1163:10)
at emitOne (events.js:116:13)
at Request.emit (events.js:211:7)
at IncomingMessage.<anonymous> (/workspace/functions/autodeploy/node_modules/request/request.js:1085:12)
at Object.onceWrapper (events.js:313:30)
at emitNone (events.js:111:20)
at IncomingMessage.emit (events.js:208:7)
at endReadableNT (_stream_readable.js:1056:12)
at _combinedTickCallback (internal/process/next_tick.js:138:11)
at process._tickCallback (internal/process/next_tick.js:180:9) code: 404 }
```

To (potentially) fix this, I've changed the `v1beta1` token endpoint URL to use `v1`, which in my tests returned a token (as opposed to 404ing like the `v1beta1` URL).

I've also updated the tests accordingly, and added timeouts to a few unrelated-but-flaky tests to ensure that they pass.

### DIFF
--- a/test/test.googleauth.ts
+++ b/test/test.googleauth.ts
@@ -1114,7 +1114,6 @@ describe('GoogleAuth', () => {
 describe('.getApplicationDefault', () => {
   it('should return a new credential the first time and a cached credential the second time',
      async () => {
-
        // Create a function which will set up a GoogleAuth instance to match
        // on an environment variable json file, but not on anything else.
        const setUpAuthForEnvironmentVariable = (creds: GoogleAuth) => {
@@ -1179,7 +1178,7 @@ describe('.getApplicationDefault', () => {
        assert.notEqual(cachedCredential, result3);
      });
 
-  it('should use environment variable when it is set', function(done) {
+  it('should use environment variable when it is set', (done) => {
 
     // We expect private.json to be the file that is used.
     const fileContents =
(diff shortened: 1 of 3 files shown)


test/test.googleauth.ts
```diff
@@ -1113,7 +1113,9 @@ describe('GoogleAuth', () => {
 
 describe('.getApplicationDefault', () => {
   it('should return a new credential the first time and a cached credential the second time',
-     async () => {
+     async function() {
```
JustinBeckwith: Lets drop this change and the one below.  If we need to increase the timeouts, we should do that in another PR :)

PR #273, opened 2018-01-31
chore: improve env var mocking

JustinBeckwith:
This change removes the not-private private function on GoogleAuth that was there only for testing purposes, and makes things a little easier to understand.

### DIFF
--- a/test/test.googleauth.ts
+++ b/test/test.googleauth.ts
@@ -130,7 +130,7 @@ function blockGoogleApplicationCredentialEnvironmentVariable() {
 
 // Intercepts the specified environment variable, returning the specified value.
 function mockEnvVar(name: string, value?: string) {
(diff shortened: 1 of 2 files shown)


test/test.googleauth.ts
```diff
@@ -119,23 +124,13 @@ function pathJoin(item1: string, item2: string) {
...
-  const originalGetEnvironmentVariableFunction = auth._getEnv;
-  auth._getEnv = (name: string) => {
-    if (name === environmentVariableName) {
-      return environmentVariableValue;
-    }
-
-    return originalGetEnvironmentVariableFunction(name);
-  };
+function mockEnvVar(name: string, value?: string) {
+  process.env[name] = value ? value : '';
```
ofrobots: nit: `process.env[name] = value || ''`;

PR #278, opened 2018-02-03
chore: use gcp-metadata

### LINKED ISSUE #190: Use gcp-metadata to talk to the metadata service
We currently have our own code to talk to the metadata service in GCE.  We should use this instead:
https://yarnpkg.com/en/package/gcp-metadata

JustinBeckwith:
First draft of integrating the new and shiny gcp-metadata.  First https://github.com/stephenplusplus/gcp-metadata/pull/24 needs to land, and we need to ship it :)

Resolves #190.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
   ],
   "dependencies": {
     "axios": "^0.17.1",
-    "gcp-metadata": "^0.4.1",
     "gtoken": "^2.1.0",
     "jws": "^3.1.4",
     "lodash.isstring": "^4.0.1",
@@ -40,6 +39,7 @@
     "@types/tmp": "0.0.33",
     "clang-format": "^1.0.50",
     "codecov": "^3.0.0",
+    "gcp-metadata": "^0.5.0",
     "gh-pages": "^1.1.0",
     "gts": "^0.5.3",
     "js-green-licenses": "^0.4.0",
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   ],
   "dependencies": {
     "axios": "^0.17.1",
+    "gcp-metadata": "^0.4.1",
     "gtoken": "^2.1.0",
     "jws": "^3.1.4",
     "lodash.isstring": "^4.0.1",
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -15,7 +15,7 @@
  */
 
 import {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios';
-
+import * as gcpMetadata from 'gcp-metadata';
 import {RequestError} from './../transporters';
 import {CredentialRequest, Credentials} from './credentials';
 import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
@@ -27,7 +27,7 @@ export class Compute extends OAuth2Client {
    * Google Compute Engine metadata server token endpoint.
    */
   protected static readonly _GOOGLE_OAUTH2_TOKEN_URL =
-      'http://metadata.google.internal/computeMetadata/v1beta1/instance/service-accounts/default/token';
+      '/computeMetadata/v1beta1/instance/service-accounts/default/token';
 
   /**
    * Google Compute Engine service account credentials.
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -21,6 +21,7 @@
   ],
   "dependencies": {
     "axios": "^0.17.1",
+    "gcp-metadata": "^0.4.1",
```
ofrobots: Leaving a reminder here to make sure this gets updated after the semver major changes in gcp-metadata land.

src/auth/computeclient.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 import {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios';
-
+import * as gcpMetadata from 'gcp-metadata';
```
ofrobots: nit: could you do a named import? AFAICT you only need `HOST_ADDRESS`.

src/auth/googleauth.ts
```diff
@@ -243,34 +244,11 @@ export class GoogleAuth {
...
-        if (!isRetry) {
-          return await this._checkIsGCE(true);
-        }
-        throw e;
-      } else if (!isDNSError) {
-        throw e;
-      }
-      this.checkIsGCE = false;
-    }
+    this.checkIsGCE = await gcpMetadata.isAvailable();
```
ofrobots: :+1: on getting rid of so much duplicated functionality. I think the retry handling is subtly different (I think 5xx errors used to retry only once). Should be fine though.

src/auth/googleauth.ts
```diff
@@ -645,19 +607,13 @@ export class GoogleAuth {
     }
 
     // For GCE, return the service account details from the metadata server
-    const result = await this.transporter.request<CredentialResult>({
-      url:
-          'http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/?recursive=true',
-      headers: {'Metadata-Flavor': 'Google'}
-    });
+    const res = await gcpMetadata.instance(
```
ofrobots: teeny nit: no need to abbreviate. Is `res` an abbreviation of `result` or `response`? I think you intended the latter, so let's just use the full name.

test/test.googleauth.ts
```diff
@@ -35,61 +37,35 @@ afterEach(() => {
   process.env = envCache;
 });
 
-function createIsGCENock(isGCE = true) {
-  nock('http://metadata.google.internal').get('/').reply(200, null, {
-    'metadata-flavor': 'Google'
-  });
+const host = gcpMetadata.HOST_ADDRESS;
+const instancePath = '/computeMetadata/v1beta1/instance';
```
ofrobots: What's going on with the `v1` vs `v1beta1`? That needs to be sorted out before this lands (hopefully with `v1` being used uniformly everywhere). Alternative, can't you use the/an export from `gcp-metadata`?

src/auth/computeclient.ts
```diff
@@ -60,11 +61,15 @@ export class Compute extends OAuth2Client {
    */
   protected async refreshToken(refreshToken?: string|
                                null): Promise<GetTokenResponse> {
-    const url = this.tokenUrl || Compute._GOOGLE_OAUTH2_TOKEN_URL;
+    const url =
```
JustinBeckwith: @ofrobots I really don't like this one bit.  Because of the fact that OAuth2Client lets you override the `tokenUrl`, we still have this stank in this one spot.  I see 3 options:

1. We add the ability to override the `tokenUrl` to `gcp-metadata`
2. We yank this option, and do a semver major release (doesn't have to be today)
3. We keep it around

I am very, very against options 3 🌮 What are your thoughts?

JustinBeckwith:
@ofrobots this is ready for a look 🎉

package.json
ofrobots: Do you still need to update this?

src/auth/googleauth.ts
ofrobots: nit: Can this method be simplified to:

```js
return this.checkIsGCE || await gcpMetadata.isAvailable();
```

JustinBeckwith: That would get rid of the caching.  I was able to simplify it a bit though :)

ofrobots: Why would it get rid of caching? The right hand side isn't evaluated if the left hand side is truthy.

JustinBeckwith: This would return the value before stashing it in `this.checkIsGCE`, wouldn't it?

ofrobots: You're right. Your version is better.

```js
this.checkIsGCE = this.checkIsGCE || await gcpMetadata.isAvailable();
return this.checkIsGCE;
```

JustinBeckwith: Fair point :) Will update in a follow up PR!

ofrobots: Seriously, I think your version as committed is better.

PR #279, opened 2018-02-07
fix: do not retry if a post with readable stream

### LINKED ISSUE #171: oAuth2client: Callback never called with media upload because of retry logic and readStream
Hello,

When trying to upload a media file to the youtube.videos.insert endpoint using a readStream, when receiving a 401 it will automatically retry the request, unfortunately the readStream has already been read, which result in never getting a callback from mikael/request.

The problem does not lie with the token but the actual service endpoint, youtube in this case, having bad permissions. Which is a problem that is solvable but I still like to have my callback invoked.

Is it possible to turn off the retry functionality in oauth2client.js? Retry logic will never work with readable streams imho.

```javascript
const google = require('googleapis');
const youtube = google.youtube('v3');

const jwtClient = new google.auth.JWT(
  credentials.client_email,
  null,
  credentials.private_key,
  [
    'https://www.googleapis.com/auth/youtubepartner',
    'https://www.googleapis.com/auth/youtube',
    'https://www.googleapis.com/auth/youtube.force-ssl',
    'https://www.googleapis.com/auth/youtube.upload',
  ],
);

jwtClient.authorize((err) => {
  if(err) throw err;

  youtube.videos.insert({
    auth: jwtClient,
    onBehalfOfContentOwner: ownerId,
    onBehalfOfContentOwnerChannel: data.channelId,
    part: 'snippet, status',
    body: {
      snippet: {
        title: data.title,
        description: data.description,
        tags: '',
        categoryId: '',
      },
      status: {
        privacyStatus: 'private',
      },
    },
    resource: {
      name: 'test',
      mimeType: 'video/mp4',
    },
    media: {
      mimeType: 'video/mp4',
      body: fs.createReadStream(data.videoFilePath),
    },
  }, (err, resp) => {

    // this gets never called

    console.log('err', err);
    console.log('resp', resp);
  });
});

```

Current **_very dirty_** monkey patching solution:
```javascript
const fs = require('fs');
const Stream = require('stream');
const google = require('googleapis');

google.auth.OAuth2.prototype._makeRequest = function (opts, callback) {
  if (opts.multipart) {
    opts.multipart = opts.multipart.map((params) => {
      if (params.body instanceof Stream.Readable) params.body = fs.createReadStream(params.body.path);
      return params;
    });
  }

  return this.transporter.request(opts, callback);
};
```

JustinBeckwith:
If users are calling `.request()` with a POST, and it's pushing a readableStream, we probably shouldn't retry the request.  This PR resolves #171, though we probably should eliminate this retry logic all up in 2.0.

### DIFF
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -18,6 +18,8 @@ import {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios
 import * as crypto from 'crypto';
 import * as http from 'http';
 import * as querystring from 'querystring';
+import * as stream from 'stream';
+
 import {PemVerifier} from './../pemverifier';
 import {BodyResponseCallback} from './../transporters';
 import {AuthClient} from './authclient';
@@ -691,7 +693,9 @@ export class OAuth2Client extends AuthClient {
         // Automatically retry 401 and 403 responses if err is set and is
         // unrelated to response then getting credentials failed, and retrying
         // won't help
-        if (!retry && (statusCode === 401 || statusCode === 403)) {
+        const isReadableStream = res.config.data instanceof stream.Readable;
+        if (!retry && (statusCode === 401 || statusCode === 403) &&
+            !isReadableStream) {
           /* It only makes sense to retry once, because the retry is intended
            * to handle expiration-related failures. If refreshing the token
            * does not fix the failure, then refreshing again probably won't
--- a/test/test.oauth2.ts
+++ b/test/test.oauth2.ts
@@ -15,12 +15,13 @@
  */
 
 import * as assert from 'assert';
-import {AxiosRequestConfig} from 'axios';
+import {AxiosError, AxiosRequestConfig} from 'axios';
 import * as crypto from 'crypto';
 import {randomBytes} from 'crypto';
 import * as fs from 'fs';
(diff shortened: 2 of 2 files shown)


src/auth/oauth2client.ts
```diff
@@ -691,7 +693,9 @@ export class OAuth2Client extends AuthClient {
         // Automatically retry 401 and 403 responses if err is set and is
         // unrelated to response then getting credentials failed, and retrying
         // won't help
-        if (!retry && (statusCode === 401 || statusCode === 403)) {
+        const isReadableStream = res.config.data instanceof stream.Readable;
+        if (!retry && (statusCode === 401 || statusCode === 403) &&
+            !isReadableStream) {
```
ofrobots: Where do you check if this was a `POST`?

JustinBeckwith:
Not sure what I was thinking with the commit message.  The same should apply to a PUT or POST.  No readable stream retries will work here.

ofrobots:
Thinking about this some more, shouldn't we restrict ourselves to [safe](https://tools.ietf.org/html/rfc7231#section-4.2.1) and [idempotent](https://tools.ietf.org/html/rfc7231#section-4.2.2) methods for retries?

ofrobots:
https://tools.ietf.org/html/rfc7230#section-6.3.1

PR #281, opened 2018-02-10
feat: add auto auth features

### LINKED ISSUE #54: Add convenience methods for common use cases

JustinBeckwith:
This adds most of the features available in [google-auto-auth](https://github.com/stephenplusplus/google-auto-auth).  

Resolves #54!


JustinBeckwith:
@stephenplusplus could you take a look?

stephenplusplus:
Nice! Would you mind showing some demos of authing with ADC, a key file, credentials object, pem/p12?

JustinBeckwith:
Sure!  Does this cover it?

stephenplusplus:
LGTM. Here's a general look at how this would be used from GC/* libs. Let me know if anything looks wack.

```js
const GoogleAuth = require('google-auth-library').GoogleAuth;

function Service(cfg) {
  // cfg may or may not have a `keyFilename` or `credentials` object:
  this.authClient = new GoogleAuth(cfg);
}

Service.prototype.request = function(reqOpts) {
  this.authClient.authorizeRequest(reqOpts, function(err, authorizedReqOpts) {
    // make request with authorizedReqOpts
  });
};
```

JustinBeckwith:
Looks great.  Only thing to note is that I decided to not support callbacks with new APIs.  This is totally up for debate though.  Is using promises a showstopper?

stephenplusplus:
Oops. No, not at all, it shouldn't make a difference 👍

examples/auto/adc.js
```diff
@@ -0,0 +1,32 @@
+// Copyright 2017, Google, Inc.
```
ofrobots: 2018. Google LLC. Here and elsewhere.

src/auth/envDetect.ts
```diff
@@ -0,0 +1,65 @@
...
+  KUBERNETES_ENGINE,
+  CLOUD_FUNCTIONS,
+  COMPUTE_ENGINE,
+  NONE
+}
+
+let env: GCPEnv;
+
+export async function getEnv() {
+  if (!env) {
```
ofrobots: Note that for the `APP_ENGINE` case, `env` will be falsy, meaning that the cache doesn't work on App Engine.

src/auth/googleauth.ts
```diff
@@ -83,13 +107,24 @@ export class GoogleAuth {
...
+
   /**
    * Export DefaultTransporter as a static property of the class.
    */
   static DefaultTransporter = DefaultTransporter;
 
+  constructor(opts?: GoogleAuthOptions) {
+    opts = opts || {};
+    this._cachedProjectId = opts.projectId || null;
+    this.keyfileName = opts.keyFilename;
```
ofrobots: cringe cringe cringe. `keyFilename` is the correct capitalization as per the style guide. Is there a way to fix the field?

src/auth/googleauth.ts
```diff
@@ -59,6 +61,28 @@ interface CredentialResult {
   default: {email: string;};
 }
 
+export interface GoogleAuthOptions {
+  /**
+   * Path to a .json, .pem, or .p12 key file
+   */
+  keyFilename?: string;
```
ofrobots: Doesn't google-auto-auth also accept `keyFile`?

src/auth/googleauth.ts
```diff
@@ -615,4 +651,64 @@ export class GoogleAuth {
...
+
+  /**
+   * Automatically obtain a client based on the provided configuration.  If no
+   * options were passed, use Application Default Credentials.
+   */
+  async getClient() {
+    if (!this.cachedCredential) {
+      if (this.keyfileName) {
+        const filePath = path.resolve(process.cwd(), this.keyfileName);
+        console.log(filePath);
```
ofrobots: delete this?

src/auth/googleauth.ts
```diff
@@ -615,4 +651,64 @@ export class GoogleAuth {
...
+    const client = await this.getClient();
+    const {headers} = await client.getRequestMetadata(url);
+    opts.headers = Object.assign(opts.headers || {}, headers);
+    return opts;
+  }
+
+  /**
+   * Determine the compute environment in which the code is running.
+   */
+  async getEnv(): Promise<GCPEnv> {
```
ofrobots: Does this need to be `async`?

src/auth/googleauth.ts
```diff
@@ -615,4 +651,64 @@ export class GoogleAuth {
...
+    if (!this.cachedCredential) {
+      if (this.keyfileName) {
+        const filePath = path.resolve(process.cwd(), this.keyfileName);
+        console.log(filePath);
+        const stream = fs.createReadStream(filePath);
+        this.cachedCredential = await this.fromStreamAsync(stream);
+      } else if (this.jsonContent) {
+        this.cachedCredential = await this.fromJSON(this.jsonContent);
+      } else {
+        await this.getApplicationDefaultAsync();
```
ofrobots: no assignment on this path?

examples/auto/adc.js
JustinBeckwith: fixed!

src/auth/googleauth.ts
JustinBeckwith: fixed!

src/auth/googleauth.ts
JustinBeckwith: It does!  This wasn't documented and I was mostly basing things off of the readme.  Added :)

src/auth/googleauth.ts
JustinBeckwith: oops

src/auth/googleauth.ts
JustinBeckwith: Yup.  The checks for GCE/GKE are async since they contact the metadata server.

ofrobots: I meant: you don't have an await and it is a single line function, so you can drop the `async` keyword and avoid some unnecessary promise wrapping.

src/auth/googleauth.ts
JustinBeckwith: It's a little gross, but that method actually caches the result itself so the assignment isn't needed.

src/auth/googleauth.ts
JustinBeckwith: Ahh, I got it.  Removed.

src/auth/envDetect.ts
ofrobots:
```ts
env = GCPEnv.APP_ENGINE;
if (env) {
  console.log('env is truthy');
} else {
  console.log('env is falsy'); // <<-- this is printed
}
```

The reason is that `APP_ENGINE` will correspond to the value 0 being the first thing in the enum. 0 is falsy.

JustinBeckwith:
I think this is in it's final form, and ready for a good look over :) @google/google-node-team

examples/authRequest.js
```diff
@@ -0,0 +1,34 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+/**
+ * Import the GoogleAuth library, and create a new GoogleAuth client.
+ */
+const { auth } = require('google-auth-library');
```
DominicKramer: Why is the following usage used here:
```js
const { auth } = require('google-auth-library');
...
```
Wouldn't it be necessary to create a `GoogleAuth` object here like in the other code examples?
```js
const { GoogleAuth } = require('google-auth-library');
...
const auth = new GoogleAuth({ ... });
```

src/auth/envDetect.ts
```diff
@@ -0,0 +1,70 @@
...
+function isCloudFunction() {
+  return !!process.env.FUNCTION_NAME;
+}
+
+async function isKubernetesEngine() {
+  try {
+    await gcpMetadata.instance('attributes/cluster-name');
+    return true;
+  } catch (e) {
+    console.error(e);
```
alexander-fenster: does it mean that it will log error every time getEnv() is called outside of Kubernetes? (e.g. when `isKubernetesEngine()` is false and `isComputeEngine()` is true)

examples/authRequest.js
JustinBeckwith: Nope :)  In the last release, we started exporting a top level `auth` object that is just an instance of `GoogleAuth`.  In most cases - that's fine to use :)

src/auth/googleauth.ts
```diff
@@ -616,4 +657,63 @@ export class GoogleAuth {
...
   }
+
+  /**
+   * Automatically obtain a client based on the provided configuration.  If no
+   * options were passed, use Application Default Credentials.
+   */
+  async getClient() {
+    if (!this.cachedCredential) {
+      if (this.keyFilename) {
+        const filePath = path.resolve(process.cwd(), this.keyFilename);
```
alexander-fenster: I guess it's the same as just `path.resolve(this.keyFilename)`, `process.cwd()` is appended by default

src/auth/envDetect.ts
JustinBeckwith: oooops.  Fixed :)

src/auth/googleauth.ts
JustinBeckwith: I believe you're right :)

PR #284, opened 2018-02-14
fix: retry connections to gce metadata server

### LINKED ISSUE #283: Auth error: ETIMEDOUT
###### Copied from original issue: https://github.com/googleapis/nodejs-pubsub/issues/30

<a href="/jamesholcomb"></a>@jamesholcomb<br>January 5, 2018 3:50 AM

#### Environment details

Note: App is running external from GCP

  - OS: Ubuntu 16.04.3 LTS
  - Node.js version: 8.9.1
  - npm version: 5.5.1
  - @google-cloud/pubsub version: 0.14.5

#### Steps to reproduce
UNKNOWN

````
Auth error:Error: connect ETIMEDOUT 172.217.4.109:443
Auth error:Error: connect ETIMEDOUT 172.217.4.109:443
Auth error:Error: connect ETIMEDOUT 172.217.4.109:443
Auth error:Error: connect ETIMEDOUT 172.217.4.109:443
Dec 31, 2017 7:05 AM ERROR  { message: 'Getting metadata from plugin failed with error: con
nect ETIMEDOUT 172.217.4.109:443',
  stack: 'Error: Getting metadata from plugin failed with error: connect ETIMEDOUT 172.217.4.109:44
3\n    at ClientDuplexStream.onConnectionStatus (/home/jholcomb/ridealert.server/node_modules/@goog
le-cloud/pubsub/src/connection-pool.js:270:21)\n    at emitOne (events.js:116:13)\n    at ClientDup
lexStream.emit (events.js:211:7)\n    at ClientDuplexStream._emitStatusIfDone (/home/jholcomb/ridea
lert.server/node_modules/grpc/src/node/src/client.js:260:10)\n    at ClientDuplexStream._receiveSta
tus (/home/jholcomb/ridealert.server/node_modules/grpc/src/node/src/client.js:233:8)\n    at /home/
jholcomb/ridealert.server/node_modules/grpc/src/node/src/client.js:757:12',
  code: 16 }
````

This error occured after the server app had been connected and processing messages for several days.  Is it expected that the pubsub lib automatically reconnects (eventually) or does a reconnect in this situation have to be handled by the client?

### PR DESCRIPTION
**Pull request by JustinBeckwith**
Resolves #283.

cc @stephenplusplus @jonparrott

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
     "gtoken": "^2.1.0",
     "jws": "^3.1.4",
     "lodash.isstring": "^4.0.1",
-    "lru-cache": "^4.1.1"
+    "lru-cache": "^4.1.1",
+    "retry-axios": "^0.3.0"
   },
   "devDependencies": {
     "@justinbeckwith/typedoc": "^0.10.1",
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -14,14 +14,20 @@
  * limitations under the License.
  */
 
-import {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios';
+import axios, {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios';
 import {BASE_PATH, HEADER_NAME, HOST_ADDRESS} from 'gcp-metadata';
+import * as rax from 'retry-axios';
 
 import {RequestError} from './../transporters';
 import {CredentialRequest, Credentials} from './credentials';
 import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
 
 export interface ComputeOptions extends RefreshOptions {}
+
+// Create a scoped axios instance that will retry 3 times by default
+const ax = axios.create();
+rax.attach(ax);
+
 
 export class Compute extends OAuth2Client {
   /**
@@ -68,8 +74,11 @@ export class Compute extends OAuth2Client {
     try {
       // TODO: In 2.0, we should remove the ability to configure the tokenUrl,
       // and switch this over to use the gcp-metadata package instead.
-      res = await this.transporter.request<CredentialRequest>(
-          {url, headers: {'Metadata-Flavor': 'Google'}});
+      res = await ax.request<CredentialRequest>({
+        url,
+        headers: {'Metadata-Flavor': 'Google'},
+        raxConfig: {noResponseRetries: 8, retry: 8, instance: ax}
+      } as rax.RaxConfig);
     } catch (e) {
       e.message = 'Could not refresh access token.';
       throw e;
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -17,8 +17,6 @@
 import axios, {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios';
 import {BASE_PATH, HEADER_NAME, HOST_ADDRESS} from 'gcp-metadata';
 import * as rax from 'retry-axios';
-
-import {RequestError} from './../transporters';
 import {CredentialRequest, Credentials} from './credentials';
 import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
 
@@ -27,7 +25,6 @@ export interface ComputeOptions extends RefreshOptions {}
 // Create a scoped axios instance that will retry 3 times by default
 const ax = axios.create();
 rax.attach(ax);
-
 
 export class Compute extends OAuth2Client {
   /**
@@ -77,7 +74,7 @@ export class Compute extends OAuth2Client {
       res = await ax.request<CredentialRequest>({
         url,
         headers: {'Metadata-Flavor': 'Google'},
-        raxConfig: {noResponseRetries: 8, retry: 8, instance: ax}
+        raxConfig: {noResponseRetries: 3, retry: 3, instance: ax}
       } as rax.RaxConfig);
     } catch (e) {
       e.message = 'Could not refresh access token.';
@@ -118,7 +115,7 @@ export class Compute extends OAuth2Client {
             e.message = helpfulMessage;
           } else {
             e = new Error(helpfulMessage);
-            (e as RequestError).code = res.status.toString();
+            (e as NodeJS.ErrnoException).code = res.status.toString();
           }
         }
       }
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -15,7 +15,7 @@
  */
 
 import axios, {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios';
-import {BASE_PATH, HEADER_NAME, HOST_ADDRESS} from 'gcp-metadata';
+import * as gcpMetadata from 'gcp-metadata';
 import * as rax from 'retry-axios';
 import {CredentialRequest, Credentials} from './credentials';
 import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -24,7 +24,8 @@
     "gtoken": "^2.1.0",
     "jws": "^3.1.4",
     "lodash.isstring": "^4.0.1",
-    "lru-cache": "^4.1.1"
+    "lru-cache": "^4.1.1",
+    "retry-axios": "^0.3.0"
```
jonparrott: Is this the same thing we use in our client libraries to do retries? If not, could we use that? If not, why not?

src/auth/computeclient.ts
```diff
@@ -68,8 +74,11 @@ export class Compute extends OAuth2Client {
     try {
       // TODO: In 2.0, we should remove the ability to configure the tokenUrl,
       // and switch this over to use the gcp-metadata package instead.
-      res = await this.transporter.request<CredentialRequest>(
```
jonparrott: So is `this.transporter` no longer used?

src/auth/computeclient.ts
```diff
@@ -68,8 +74,11 @@ export class Compute extends OAuth2Client {
...
-          {url, headers: {'Metadata-Flavor': 'Google'}});
+      res = await ax.request<CredentialRequest>({
+        url,
+        headers: {'Metadata-Flavor': 'Google'},
+        raxConfig: {noResponseRetries: 8, retry: 8, instance: ax}
```
jonparrott: 8 times is a bit extreme. Python is 3, I believe.

src/auth/computeclient.ts
JustinBeckwith: It's used in other places, just not in this particular case.  All that transporter thing really does is attach user agent headers.   This is important in places like the apiary client, but not relevant in this case.

src/auth/computeclient.ts
JustinBeckwith: Hah, Cunningham's Law proves itself again

package.json
jonparrott: Do you think it's an okay state to be in that our libraries use multiple http libraries and multiple retry libraries?

JustinBeckwith: It's not ideal, and should probably change.  I'd love to have a plan to migrate us 100% of axios, but it's beyond the scope of what I can accomplish in this library.  Happy to offer guidance and input though :)

ofrobots: The alternatives are:
* Continue using a client library that has problems.
* Rewrite all code to switch atomically to a different library.

There may be *some* downsides of multiple client libraries; but those are unspecified in this thread and assumed to be terrible. IMHO they aren't terrible.

jonparrott: It was an honest question - I am ramping up on all things in all languages and trying to figure out what matters and what doesn't.

For context, in Python we were faced with a *very* similar issue - we used httplib2 throughout and it was horrible. We needed to rapidly move to requests, starting by allowing the auth library to use [*multiple* transports](http://google-auth.readthedocs.io/en/latest/reference/google.auth.transport.html). We then updated our libraries one by one to use requests instead of httplib2. There was a little overlap where two different libraries would use different transports, but within a single library one transport was used. It took ~month to get everything moved over.

This was a much bigger concern in Python than my current understanding leads me to believe it is in node.

src/auth/computeclient.ts
```diff
@@ -68,8 +71,11 @@ export class Compute extends OAuth2Client {
     try {
       // TODO: In 2.0, we should remove the ability to configure the tokenUrl,
       // and switch this over to use the gcp-metadata package instead.
-      res = await this.transporter.request<CredentialRequest>(
-          {url, headers: {'Metadata-Flavor': 'Google'}});
+      res = await ax.request<CredentialRequest>({
+        url,
+        headers: {'Metadata-Flavor': 'Google'},
+        raxConfig: {noResponseRetries: 3, retry: 3, instance: ax}
```
ofrobots: The `instance` property of `raxConfig` is not [documented](https://github.com/JustinBeckwith/retry-axios/blob/a8defe01059fff98f1b2a056fd3df62e9bc22021/README.md). What does it do?

JustinBeckwith: Yeah, so this turned out to be a funny bug.  If you use `axios.create` to create a scoped instance, you need to pass a reference to that instance into the config.  Otherwise, there's no way to figure out which instance throws the error.  I need to go update the docs and submit a PR to fix this in gcp-metadata.

src/auth/computeclient.ts
```diff
@@ -68,8 +71,11 @@ export class Compute extends OAuth2Client {
     try {
       // TODO: In 2.0, we should remove the ability to configure the tokenUrl,
       // and switch this over to use the gcp-metadata package instead.
-      res = await this.transporter.request<CredentialRequest>(
-          {url, headers: {'Metadata-Flavor': 'Google'}});
+      res = await ax.request<CredentialRequest>({
+        url,
+        headers: {'Metadata-Flavor': 'Google'},
```
ofrobots: You have a typo in those headers, and without an end-to-end deployment test we wouldn't have caught this.  We will only find out when something is actually deployed by a user.

Actually there is no typo (I hope), but I had you for a second.

Can you use an exported value of this constant from `gcp-metadata` rather than typing it again, potentially with typos.

test/test.compute.ts
```diff
@@ -120,6 +118,44 @@ describe('Compute auth client', () => {
...
+  it('should retry calls to the metadata service if there are network errors',
+     (done) => {
+       const scopes = [
+         nock(HOST_ADDRESS).get(tokenPath).times(2).replyWithError({
+           code: 'ENOTFOUND'
+         }),
+         nock(HOST_ADDRESS).get(tokenPath).reply(200, {
+           access_token: 'abc123',
+           expires_in: 10000
+         })
```
ofrobots: If I am reading `nock` documentation correctly, I think this can be simplified:

```js
const scope = nock(HOST_ADDRESS).get(tokenPath).times(2)
  .replyWithError({ code: 'ENOTFOUND' }),
  .reply(200, {
    access_token: 'abc123',
    expires_in: 10000
});
```

And then you can get rid of the `foreach` below.

test/test.compute.ts
```diff
@@ -120,6 +118,44 @@ describe('Compute auth client', () => {
...
+       });
+     });
+
+  it('should retry calls to the metadata service if it returns non-200 errors',
+     (done) => {
+       const scopes = [
+         nock(HOST_ADDRESS).get(tokenPath).times(2).reply(500),
+         nock(HOST_ADDRESS).get(tokenPath).reply(200, {
+           access_token: 'abc123',
+           expires_in: 10000
```
ofrobots: likewise.

src/auth/computeclient.ts
JustinBeckwith: The change to the docs (and test) has been made to retry-axios:
https://github.com/JustinBeckwith/retry-axios/commit/544c2a6563c3c4df69d5d441bbaf872a0a59d83f

src/auth/computeclient.ts
JustinBeckwith: I fixed this to use the existing HEADER_NAME, and submitted a PR to gcp-metadata to export the whole header 👍 
https://github.com/stephenplusplus/gcp-metadata/pull/33

test/test.compute.ts
JustinBeckwith: fixed

JustinBeckwith:
@ofrobots @jonparrott Does this look good?

src/auth/computeclient.ts
```diff
@@ -61,15 +64,18 @@ export class Compute extends OAuth2Client {
...
     let res: AxiosResponse<CredentialRequest>|null = null;
     // request for new token
     try {
       // TODO: In 2.0, we should remove the ability to configure the tokenUrl,
       // and switch this over to use the gcp-metadata package instead.
-      res = await this.transporter.request<CredentialRequest>(
-          {url, headers: {'Metadata-Flavor': 'Google'}});
+      res = await ax.request<CredentialRequest>({
+        url,
+        headers: {[gcpMetadata.HEADER_NAME]: 'Google'},
```
ofrobots: Could you use `headers: gcpMetadata.HEADERS` or is that not released yet?

ofrobots: or rather `headers: Object.assign({}, gcpMetadata.HEADERS)`

JustinBeckwith: It's not released yet.  There's one more bug in there I'd like to land before 0.6.0.  Tracking here: https://github.com/stephenplusplus/gcp-metadata/issues/35

PR #301, opened 2018-02-25
test: flatten tests

JustinBeckwith:
This change flattens out the structure of our tests, dropping the `describe` calls.  Doing this in preparation for a move over to ava, which doesn't support nested describe blocks.  Did a little cleanup along the way.


JustinBeckwith:
The code coverage went down because I "fixed a bug" in the tests.  Filed #302 to make sure we come back around and actually test this code correctly after this lands.

PR #305, opened 2018-03-02
chore: drop support for node.js 4

JustinBeckwith:
BREAKING CHANGE: This change removes support for Node.js 4.x.  

Note sure if we want to put these changes in another branch or not.  (Probably a good idea)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,8 @@
   "extends": "./node_modules/gts/tsconfig-google.json",
   "compilerOptions": {
     "rootDir": ".",
-    "outDir": "build"
+    "outDir": "build",
+    "target": "es2016"
   },
   "include": [
     "src/*.ts",
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,8 +23,6 @@ workflows:
   version: 2
   tests:
     jobs:
-      - node4:
-          filters: *release_tags
       - node6:
           filters: *release_tags
       - node8:
@@ -33,7 +31,6 @@ workflows:
           filters: *release_tags
       - publish_npm:
           requires:
-            - node4
             - node6
             - node8
             - node9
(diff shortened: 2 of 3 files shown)


ofrobots:
Separate branch is probably a good idea in case we need to release a fix in the next 8 weeks.

JustinBeckwith:
@ofrobots I updated next to match master, and pointed the PR at next.  @google/google-node-team ready for another look!

tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "./node_modules/gts/tsconfig-google.json",
   "compilerOptions": {
     "rootDir": ".",
-    "outDir": "build"
+    "outDir": "build",
+    "target": "es2016"
```
jinwoo: Is es2016 equivalent to Node 6?

JustinBeckwith: No :(  All the tests pass here pass, but I can't promise that we won't run into trouble down the road.

jinwoo: Then why not be more conservative and use `es6` (aka `es2015`)?

JustinBeckwith: There's no direct relation between Node.js releases and a specific TypeScript target.  Here's my rejected feature request on the TypeScript repo:
https://github.com/Microsoft/TypeScript/issues/22306

For the moment, everything works with es2016.  I was optimizing for the most current feature set that passes all the tests.  The code that gets generated here is likely closer to the origin code, which makes for better debugging and likely better perf.

jinwoo: I know that there's no direct relation between node and a TS target. But AFAIK Node 6 is closer to es2015 vs es2016. Having said that, it's probably fine because we have a set of tests with a pretty good coverage.

ofrobots: Looking at https://node.green, as far as I can tell, the only thing missing for ES2016 for Node 6 is exponentiation operator. I think we're pretty safe to use es2016 as the target here.

jinwoo: Great. Good to know!

PR #310, opened 2018-03-05
chore: cleanup iam interface

JustinBeckwith:
BREAKING CHANGE:  The `getRequestMetadata` of `IAMAuth` now directly returns http headers instead of requiring a callback function.

#### Old code

```js
IAMAuthClient.getRequestMetadata(null, (err, headers) => {
   console.log(headers);
});
```

#### New code

```js
const headers = IAMAuthClient.getRequestMetadata();
console.log(headers);
```


alexander-fenster:
If it's a breaking change, it's definitely not a `chore` :)

JustinBeckwith:
Fair enough :) Made it a fix!

PR #316, opened 2018-03-10
fix: increase test timeout

JustinBeckwith:
Fixing up the random test failures happening in 4 and 6.


JustinBeckwith:
TBR = @ofrobots

PR #318, opened 2018-03-10
feat: get rid of createScopedRequired

### LINKED ISSUE #264: Remove createScopedRequired and createScope methods
Part of the feedback in #55 was to re-rationalize the createScopedRequired and createScope methods.  I wrote #255 to try addressing this, but it was way too early for a semver major change after going 1.0.

JustinBeckwith:
Resolves #264

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,12 +52,14 @@ Rather than manually creating an OAuth2 client, JWT client, or Compute client, t
 
 For example, a JWT auth client will be created when your code is running on your local developer machine, and a Compute client will be created when the same code is running on Google Cloud Platform.
 
-The code below shows how to retrieve a default credential type, depending upon the runtime environment. The createScopedRequired must be called to determine when you need to pass in the scopes manually, and when they have been set for you automatically based on the configured runtime environment.
+The code below shows how to retrieve a default credential type, depending upon the runtime environment.
 
 ```js
 const {auth} = require('google-auth-library');
 async function main() {
-  const client = await auth.getClient();
+  const client = await auth.getClient({
+    scopes: 'https://www.googleapis.com/auth/cloud-platform'
+  });
   const projectId = await auth.getDefaultProjectId();
   const url = `https://www.googleapis.com/dns/v1/projects/${projectId}`;
   const res = await client.request({ url });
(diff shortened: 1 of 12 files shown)


JustinBeckwith:
I am very confused and do not know why its showing all these commits.  "get rid of createScopedRequired" should be the only one.

test/test.googleauth.ts
```diff
@@ -1229,6 +1230,13 @@ it('should accept credentials to get a client', async () => {
   assert.equal(client.email, 'hello@youarecool.com');
 });
 
+it('should allow passing scopes to get a client', async () => {
+  const scopes = ['http://examples.com/is/a/scope'];
```
jinwoo: `scopes` can be an array or a string, right? Maybe it's worthwhile to have a test for the latter case too?

README.md
```diff
@@ -52,12 +52,14 @@ Rather than manually creating an OAuth2 client, JWT client, or Compute client, t
 
 For example, a JWT auth client will be created when your code is running on your local developer machine, and a Compute client will be created when the same code is running on Google Cloud Platform.
 
-The code below shows how to retrieve a default credential type, depending upon the runtime environment. The createScopedRequired must be called to determine when you need to pass in the scopes manually, and when they have been set for you automatically based on the configured runtime environment.
+The code below shows how to retrieve a default credential type, depending upon the runtime environment.
```
jinwoo: It'd be good to mention that `scopes` can be either a string array or just a string.

test/test.googleauth.ts
JustinBeckwith: good call :)

test/test.googleauth.ts
```diff
@@ -1233,6 +1233,13 @@ it('should allow passing scopes to get a client', async () => {
   assert.equal(client.scopes, scopes);
 });
 
+it('should allow passing a scope to get a client', async () => {
```
jinwoo: Better to have a different description than the test above.

test/test.googleauth.ts
```diff
@@ -1225,6 +1226,20 @@ it('should accept credentials to get a client', async () => {
...
 });
 
+it('should allow passing scopes to get a client', async () => {
+  const scopes = ['http://examples.com/is/a/scope'];
+  const keyFilename = './test/fixtures/private.json';
+  const client = await auth.getClient({scopes, keyFilename}) as JWT;
+  assert.equal(client.scopes, scopes);
+});
+
+it('should allow passing a scope to get a client', async () => {
```
jinwoo: Put the comment in a wrong commit. Trying again.
Better to have a description than the test above.

PR #319, opened 2018-03-10
chore: remove tokenUrl option

### LINKED ISSUE #280: Remove tokenUrl override
There's a tokenUrl config setting that makes it hard to use `gcp-metadata` to request tokens in the ComputeClient:
https://github.com/google/google-auth-library-nodejs/blob/master/src/auth/computeclient.ts#L69

We should remove this code, and switch over to using gcp-metadata for all communication in the client.

JustinBeckwith:
Resolves #280

### DIFF
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -234,8 +234,6 @@ export class OAuth2Client extends AuthClient {
   private redirectUri?: string;
   private certificateCache: {}|undefined;
   private certificateExpiry: Date|undefined;
-  protected authBaseUrl?: string;
-  protected tokenUrl?: string;
 
   // TODO: refactor tests to make this private
   _clientId?: string;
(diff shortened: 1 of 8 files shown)


src/auth/oauth2client.ts
```diff
@@ -342,8 +338,7 @@ export class OAuth2Client extends AuthClient {
     if (opts.scope instanceof Array) {
       opts.scope = opts.scope.join(' ');
     }
-    const rootUrl =
-        this.authBaseUrl || OAuth2Client.GOOGLE_OAUTH2_AUTH_BASE_URL_;
+    const rootUrl = OAuth2Client.GOOGLE_OAUTH2_AUTH_BASE_URL_;
```
kjin: Why remove `authBaseUrl`?

JustinBeckwith: It's another one of those properties that doesn't really have a reason to exist.  I can only assume it was here for unit testing purposes, but there are far better ways to do that.

PR #322, opened 2018-03-13
fix: don't duplicate user-agent

JustinBeckwith:
When deciding if we should add the google api user agent we were checking both the name *and* the value. In reality, we shouldn't duplicate the user-agent if the name matches, and the value doesn't.

### DIFF
--- a/src/transporters.ts
+++ b/src/transporters.ts
@@ -19,6 +19,7 @@ import {validate} from './options';
 
 // tslint:disable-next-line no-var-requires
 const pkg = require('../../package.json');
+const PRODUCT_NAME = 'google-api-nodejs-client';
 
 export interface Transporter {
   request<T>(opts: AxiosRequestConfig): AxiosPromise<T>;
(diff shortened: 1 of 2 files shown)


src/transporters.ts
```diff
@@ -50,13 +50,12 @@ export class DefaultTransporter {
...
     // set transporter user agent
     opts.headers = opts.headers || {};
-    if (!opts.headers['User-Agent']) {
+    const uaValue: string = opts.headers['User-Agent'];
+    if (!uaValue) {
       opts.headers['User-Agent'] = DefaultTransporter.USER_AGENT;
-    } else if (
-        opts.headers['User-Agent'].indexOf(DefaultTransporter.USER_AGENT) ===
-        -1) {
+    } else if (!uaValue.includes('google-api-nodejs-client/')) {
```
jinwoo: Can you define a constant for `'google-api-nodejs-client/'` and use it here and also in the `DefaultTransporter.USER_AGENT` definition?

JustinBeckwith: 👍

JustinBeckwith:
Great question! So, the way it is now is kinda a bug.  This code was lifted (a long time ago) from google-api-nodejs-client, and they included this header with the npm module version.  Problem is - it pulls the version from google-auth-library :(

To fix this, I submitted this PR over in the other repo:
https://github.com/google/google-api-nodejs-client/pull/1065/files

With that change, new versions of the google-api-nodejs-client should be reporting the correct version, while only older versions will fall back to the code here.  At some point in the future (year, two) we can further clean this up.

PR #323, opened 2018-03-13
feat: add JWT.getCredentials method

### LINKED ISSUE #313: JWT: Can I call `getCredentials`?
The JWTClient doesn't appear to let me call a "getCredentials()" function. Ideally, the JWTClient could recognize the p12/pem file, parse it, and internally store then return the private key from `getCredentials()`.

JustinBeckwith:
Resolves #313


src/auth/jwtclient.ts
```diff
@@ -168,6 +168,20 @@ export class JWT extends OAuth2Client {
    * @private
    */
   async refreshToken(refreshToken?: string|null): Promise<GetTokenResponse> {
+    this.createGToken();
+    const token = await this.gtoken!.getToken();
```
jinwoo: What about making `createGToken()` return a string? Then you won't need the non-null annotation here.

src/auth/jwtclient.ts
```diff
@@ -261,4 +268,19 @@ export class JWT extends OAuth2Client {
     }
     this.apiKey = apiKey;
   }
+
+  /**
+   * Using the key or keyFile on the JWT client, obtain an object that contains
+   * the key and the client email.
+   */
+  async getCredentials() {
```
jinwoo: I'd like public methods to have their return type (and param types) explicitly specified.

JustinBeckwith: 👍

src/auth/jwtclient.ts
JustinBeckwith: 👍

jinwoo: Not done here and in line 177 below ?

JustinBeckwith: Ooops, sorry bout that.  Cleared it up.

PR #324, opened 2018-03-14
fix: ensure GCF can read metadata requests

### LINKED ISSUE #321: GCF: `getCredentials` returning "Failure from metadata server."
*Originally reported here: https://github.com/googleapis/nodejs-storage/issues/150*

I'm receiving this error in GCF using `google-auth-library`@^1.3.1:

```js
Error: Failure from metadata server.
    at GoogleAuth.<anonymous> (/user_code/node_modules/google-auth-library/build/src/auth/googleauth.js:634:35)
    at step (/user_code/node_modules/google-auth-library/build/src/auth/googleauth.js:47:23)
    at Object.next (/user_code/node_modules/google-auth-library/build/src/auth/googleauth.js:28:53)
    at fulfilled (/user_code/node_modules/google-auth-library/build/src/auth/googleauth.js:19:58)
    at process._tickDomainCallback (internal/process/next_tick.js:135:7)
```

Cloud function source:

```js
const GoogleAuth = require('google-auth-library').GoogleAuth

exports.test = function test (req, res) {
  function end(err) {
    console.error(err)
    res.status(500).end()
  }

  const authClient = new GoogleAuth()
  authClient.getApplicationDefault(err => {
    if (err) return end(err)

    authClient.getCredentials((err, credentials) => {
      if (err) return end(err)
      console.log(credentials)
      res.send('Ok')
    })
  })
}
```

Am I doing anything wrong this way?

JustinBeckwith:
Resolves #321 

Turns out that the GCE metadata server isn't picky about the trailing slash, but GCF is 🤷‍♂️

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -631,6 +631,9 @@ export class GoogleAuth {
     }
 
     // For GCE, return the service account details from the metadata server
+    // NOTE: The trailing '/' at the end of service-accounts/ is very important!
+    // The GCF metadata server doesn't respect querystring params if this / is
+    // not included.
     const {data} = await gcpMetadata.instance(
         {property: 'service-accounts/', params: {recursive: true}});
 
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -632,7 +632,7 @@ export class GoogleAuth {
 
     // For GCE, return the service account details from the metadata server
     const {data} = await gcpMetadata.instance(
-        {property: 'service-accounts', params: {recursive: true}});
+        {property: 'service-accounts/', params: {recursive: true}});
 
     if (!data || !data.default || !data.default.email) {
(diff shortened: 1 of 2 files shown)


src/auth/googleauth.ts
```diff
@@ -632,7 +632,7 @@ export class GoogleAuth {
 
     // For GCE, return the service account details from the metadata server
     const {data} = await gcpMetadata.instance(
-        {property: 'service-accounts', params: {recursive: true}});
+        {property: 'service-accounts/', params: {recursive: true}});
```
jinwoo: It'd be great to add some comments here that the slash at the end is important, and describes briefly what happened due to this, so that future maintainers make sure to not delete it.

JustinBeckwith: Great idea :) Added!

PR #328, opened 2018-03-19
fix: use id_token for IAP requests

### LINKED ISSUE #327: IAP authentication fails due to gtoken not returning a refresh token
Followed the IAP example, provided all the data, found out that when jwtclient calls refresh token, we get:

```
    { res: null,
      tokens: 
       { access_token: undefined,
         token_type: 'Bearer',
         expiry_date: NaN } }
```

Which doesnt raise an error, culprit seems to be here, because it returns `id_token` and not `access_token`, which the library expects by default. 

Related issue here: https://github.com/google/node-gtoken/issues/58

JustinBeckwith:
This fixes an issue where we look at the access_token instead of the id_token for IAP based requests.  The `any` will be resolved when https://github.com/google/node-gtoken/pull/59 lands. 

Resolves #327

### DIFF
--- a/src/auth/jwtclient.ts
+++ b/src/auth/jwtclient.ts
@@ -103,7 +103,7 @@ export class JWT extends OAuth2Client {
                                      target_audience: string
                                    }).target_audience) {
         const {tokens} = await this.refreshToken();
-        return {headers: {Authorization: `Bearer ${tokens.access_token}`}};
+        return {headers: {Authorization: `Bearer ${tokens.id_token}`}};
       } else {
         // no scopes have been set, but a uri has been provided. Use JWTAccess
         // credentials.
@@ -174,7 +174,9 @@ export class JWT extends OAuth2Client {
     const tokens = {
       access_token: token,
       token_type: 'Bearer',
-      expiry_date: gtoken.expiresAt
+      expiry_date: gtoken.expiresAt,
+      // tslint:disable-next-line no-any
+      id_token: (gtoken.rawToken! as any).id_token
     };
     return {res: null, tokens};
   }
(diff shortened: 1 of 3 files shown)


AVVS:
expiresAt from gtoken would be NaN if I’m not mistaken in case of id_token

AVVS:
nvm, checking https://github.com/google/node-gtoken/pull/59/files and it should fix this in conjunction

JustinBeckwith:
Ping @google/google-node-team for some PR love :)

src/auth/jwtclient.ts
```diff
@@ -103,7 +103,7 @@ export class JWT extends OAuth2Client {
                                      target_audience: string
                                    }).target_audience) {
         const {tokens} = await this.refreshToken();
-        return {headers: {Authorization: `Bearer ${tokens.access_token}`}};
+        return {headers: {Authorization: `Bearer ${tokens.id_token}`}};
```
jinwoo: I'm not familiar with IAP authentication. So now `access_token` is not used anymore? Should it be deleted from `Credentials` and `CredentialRequest` then?

JustinBeckwith: Nah.  IAP is weird.  If you pass a "target_audience" claim to the Google token endpoint, it will return an id_token instead of an access_token.  For most cases, users won't be passing that claim, except for IAP users.  We need to keep the access_token and such in both those interfaces.

jinwoo: I see. A follow-up question. I don't see any usages of `access_token` in jwtclient.ts. Does it mean that JWT uses only IAP authentication?

JustinBeckwith: Nope!  JWTClient inherits from OAuth2Client, which *does* make use of the access_token that's returned in other cases.

PR #337, opened 2018-03-30
feat: add getTokenInfo method

### LINKED ISSUE #336: What is the correct way to use OAuth2?
_From @megamindbrian on October 12, 2017 1:4_

There seems to be some ambiguity in using the Google APIs.  How do I check if an old token has access to a new scope?

My code looks like this example:
https://developers.google.com/drive/v3/web/quickstart/nodejs

But this example shows the client.init() function using scopes:
https://developers.google.com/api-client-library/javascript/samples/samples#authorizing-and-making-authorized-requests

Is there a way to validate the scope using the OAuth2 layer, or do I have to connect to the drive service (for the first example), get and error, and then return back to setting up a new OAuth2 connection?  Seems a little counter-intuitive to get an authentication error from the service layer, that should happen when I create an OAuth2 client with right the scopes, right?

_Copied from original issue: google/google-api-nodejs-client#822_

JustinBeckwith:
This adds a new `getTokenInfo` method to `OAuth2Client`, which calls the tokeninfo endpoint described here:
https://developers.google.com/identity/protocols/OAuth2UserAgent#validate-access-token

Resolves #336.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,20 @@ const tokens = await oauth2Client.refreshAccessToken();
 // store these new tokens in a safe place (e.g. database)
 ```
 
+##### Checking access_token information
+After obtaining and storing an access_token, at a later time you may want to go check the expiration date,
+original scopes, or audience for the token.  To get the token info, you can use the `getTokenInfo` method:
+
+```js
+// after acquiring an oAuth2Client...
+const tokenInfo = await oAuth2client.getTokenInfo('my-access-token');
+
+// take a look at the scopes originally provisioned for the access token
+console.log(tokenInfo.scopes);
+```
+
+This method will throw if the token is invalid.
+
 ##### OAuth2 with Installed Apps (Electron)
 If you're authenticating with OAuth2 from an installed application (like Electron), you may not want to embed your `client_secret` inside of the application sources. To work around this restriction, you can choose the `iOS` application type when creating your OAuth2 credentials in the [Google Developers console][devconsole]:
 
--- a/README.md
+++ b/README.md
@@ -221,8 +221,8 @@ const tokens = await oauth2Client.refreshAccessToken();
 // store these new tokens in a safe place (e.g. database)
 ```
 
-##### Checking access_token information
-After obtaining and storing an access_token, at a later time you may want to go check the expiration date,
+##### Checking `access_token` information
+After obtaining and storing an `access_token`, at a later time you may want to go check the expiration date,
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -221,6 +221,20 @@ const tokens = await oauth2Client.refreshAccessToken();
 // store these new tokens in a safe place (e.g. database)
 ```
 
+##### Checking access_token information
```
kjin: Nit: Codify `access_token`? Here and below

examples/jwt.js
```diff
@@ -31,10 +31,14 @@ async function main() {
...
     scopes: ['https://www.googleapis.com/auth/cloud-platform']
   });
-  await client.authorize();
   const url = `https://www.googleapis.com/dns/v1/projects/${keys.project_id}`;
   const res = await client.request({ url });
   console.log(res.data);
+
+  // After acquiring an access_token, you may want to check on the audience, expiration,
+  // or original scopes requested.  You can do that with the `getTokenInfo` method.
+  const info = await client.getTokenInfo(client.credentials.access_token);
```
kjin: Nit: Use `tokenInfo` to match the oauth2.js and README example.

src/auth/oauth2client.ts
```diff
@@ -36,6 +36,75 @@ export interface GetTokenOptions {
...
+   * is not present at the browser. This value instructs the Google
+   * authorization server to return a refresh token and an access token the
+   * first time that your application exchanges an authorization code for
+   * tokens.
+   */
+  access_type?: string;
+}
+
+
+export interface TokenInfoRequest {
```
kjin: - From usage it looks like `azp`, `sub, and `access_type` can be optional. (Or rather, the corresponding fields in `TokenInfo` should be not undefined.)
- `exp` seems to disappear in `TokenInfo`. Is this intentional?

README.md
JustinBeckwith: Fixed!

examples/jwt.js
JustinBeckwith: fixed

src/auth/oauth2client.ts
JustinBeckwith: First one fixed to be consistent.  With exp - yes this was intentional.  This is the expiration date for the token (which seems to be a copy of expires_in...).  There's practically no documentation for this property, so I don't want to explicitly delete it, and I want users to use `expiry_date` instead.

PR #339, opened 2018-04-02
fix: cache refreshToken promises

### LINKED ISSUE #338: Cache promise for refreshToken requests
We're getting reports of 1000s of connections to googleapis when a flood of new requests come in with an expired access token.  We should cache a single promise for refreshToken requests, and return that when the refresh is in transit.

JustinBeckwith:
Resolves #338

### DIFF
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -504,26 +504,26 @@ export class OAuth2Client extends AuthClient {
    */
   protected async refreshToken(refreshToken?: string|
                                null): Promise<GetTokenResponse> {
-    if (refreshToken) {
-      // If a request to refresh using the same token has started,
-      // return the same promise.
-      if (this.refreshTokenPromises.has(refreshToken)) {
-        return this.refreshTokenPromises.get(refreshToken)!;
-      }
-
-      const p = this.refreshTokenNoCache(refreshToken)
-                    .then(r => {
-                      this.refreshTokenPromises.delete(refreshToken);
-                      return r;
-                    })
-                    .catch(e => {
-                      this.refreshTokenPromises.delete(refreshToken);
-                      throw e;
-                    });
-      this.refreshTokenPromises.set(refreshToken, p);
-      return p;
+    if (!refreshToken) {
+      return this.refreshTokenNoCache(refreshToken);
     }
-    return this.refreshTokenNoCache(refreshToken);
+    // If a request to refresh using the same token has started,
+    // return the same promise.
+    if (this.refreshTokenPromises.has(refreshToken)) {
+      return this.refreshTokenPromises.get(refreshToken)!;
+    }
+
+    const p = this.refreshTokenNoCache(refreshToken)
+                  .then(r => {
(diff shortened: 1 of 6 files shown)


src/auth/oauth2client.ts
```diff
@@ -503,6 +504,30 @@ export class OAuth2Client extends AuthClient {
    */
   protected async refreshToken(refreshToken?: string|
                                null): Promise<GetTokenResponse> {
+    if (refreshToken) {
```
ofrobots: nit: reverse this branch to flatten the meat of this method.

src/auth/refreshclient.ts
```diff
@@ -65,9 +65,9 @@ export class UserRefreshClient extends OAuth2Client {
    * @param refreshToken An ignored refreshToken..
    * @param callback Optional callback.
    */
-  protected async refreshToken(refreshToken?: string|
-                               null): Promise<GetTokenResponse> {
-    return super.refreshToken(this._refreshToken);
+  protected async refreshTokenNoCache(refreshToken?: string|
+                                      null): Promise<GetTokenResponse> {
+    return super.refreshTokenNoCache(this._refreshToken!);
```
ofrobots: Why do you need the non-null operator? I don't quite follow how you can guarantee it is non-null anyway.

test/test.jwt.ts
```diff
@@ -244,6 +244,54 @@ it('should refresh token if missing access token', (done) => {
...
+    nock('http://example.com').get('/').thrice().reply(200)
+  ];
+  const jwt = new JWT({
+    email: 'foo@serviceaccount.com',
+    keyFile: PEM_PATH,
+    scopes: ['http://bar', 'http://foo'],
+    subject: 'bar@subjectaccount.com'
+  });
+  jwt.credentials = {refresh_token: 'jwt-placeholder'};
+  jwt.credentials = {refresh_token: 'refresh-token-placeholder'};
```
ofrobots: Both these lines needed?

src/auth/oauth2client.ts
```diff
@@ -503,6 +504,30 @@ export class OAuth2Client extends AuthClient {
...
+        return this.refreshTokenPromises.get(refreshToken)!;
+      }
+
+      const p = this.refreshTokenNoCache(refreshToken)
+                    .then(r => {
+                      this.refreshTokenPromises.delete(refreshToken);
+                      return r;
+                    })
+                    .catch(e => {
+                      this.refreshTokenPromises.delete(refreshToken);
```
ofrobots: Is this catch path exercised by the tests?

JustinBeckwith: Nope.  Added one!

src/auth/refreshclient.ts
JustinBeckwith: This wasn't needed

JustinBeckwith:
Just tested out the fixed module with firebase, and I'm only seeing 5 connections over 1000 simultaneous calls.  Ready for another look @ofrobots :)

image

PR #341, opened 2018-04-04
feat: emit an event when new tokens are obtained

### LINKED ISSUE #340: README doesn’t describe how to detect when automatically refreshed access token should be stored
_From @binki on July 24, 2017 0:32_

The README states:

> This client comes with an [OAuth2](https://developers.google.com/identity/protocols/OAuth2) client that allows you to retrieve an access token and refreshes the token and retry the request seamlessly if you also provide an`expiry_date` and the token is expired.

Then later there are examples of how to initially obtain the expiry date, access token, and refresh token. And [instructions to save these to the database](https://github.com/google/google-api-nodejs-client/tree/b4c694b8a96de90c4ea414c4df27ada449a698ab#retrieve-access-token).

However, it is my understanding that when the library automatically retries a request after the access token has expired and uses the refresh token to get a new access code, this new access code and expiry should be stored into the database. Shouldn’t the examples in the README describe how to subscribe to the “automatic refresh has happened” event (if there is one) so that future requests made by the application after it restarts (or by other processes sharing an access token through a database) can use the new access token instead of always needing to refresh on startup?

_Copied from original issue: google/google-api-nodejs-client#783_

JustinBeckwith:
Today we really have no way to tell users when their access_token was refreshed.  Some people store these on the server across multiple requests, so giving them some way to know when it happens would be nice.  Users are asking for this over in google-api-nodejs-client.  

Resolves #340.

### DIFF
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -494,6 +494,7 @@ export class OAuth2Client extends AuthClient {
           ((new Date()).getTime() + (res.data.expires_in * 1000));
       delete (tokens as CredentialRequest).expires_in;
     }
+    this.emit('tokens', tokens);
     return {tokens, res};
   }
 
@@ -551,6 +552,9 @@ export class OAuth2Client extends AuthClient {
           ((new Date()).getTime() + (res.data.expires_in * 1000));
       delete (tokens as CredentialRequest).expires_in;
     }
+
+    this.emit('tokens', tokens);
+
     return {tokens, res};
   }
 
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -552,9 +552,7 @@ export class OAuth2Client extends AuthClient {
           ((new Date()).getTime() + (res.data.expires_in * 1000));
       delete (tokens as CredentialRequest).expires_in;
     }
-
     this.emit('tokens', tokens);
-
     return {tokens, res};
   }
 
--- a/test/test.jwt.ts
+++ b/test/test.jwt.ts
@@ -133,28 +133,20 @@ it('can get obtain new access token when scopes are set', (done) => {
   });
 });
 
-
 it('should emit an event for tokens', (done) => {
+  const accessToken = 'initial-access-token';
+  const scope = createGTokenMock({access_token: accessToken});
   const jwt = new JWT({
     email: 'foo@serviceaccount.com',
     keyFile: PEM_PATH,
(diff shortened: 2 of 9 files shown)


src/auth/oauth2client.ts
```diff
@@ -551,6 +552,9 @@ export class OAuth2Client extends AuthClient {
           ((new Date()).getTime() + (res.data.expires_in * 1000));
       delete (tokens as CredentialRequest).expires_in;
     }
+
+    this.emit('tokens', tokens);
+
```
alexander-fenster: nit: inconsistent formatting, can you remove this empty line?

test/test.jwt.ts
```diff
@@ -133,6 +133,30 @@ it('can get obtain new access token when scopes are set', (done) => {
...
+    email: 'foo@serviceaccount.com',
+    keyFile: PEM_PATH,
+    scopes: ['http://bar', 'http://foo'],
+    subject: 'bar@subjectaccount.com'
+  });
+
+  let raisedEvent = false;
+  jwt.on('tokens', tokens => {
+    assert.equal(tokens.access_token, 'initial-access-token');
+    raisedEvent = true;
```
alexander-fenster: do we need this flag? why not just call `done()` here?

alexander-fenster: I feel this test adds some extra assertion on the order of events (`tokens` is expected to be emitted before the 'getAccessToken' callback is called). Even if it's how things are supposed to work, I feel it's wrong to rely on this in the unit test (I can be wrong here of course, don't know the code well enough)

test/test.oauth2.ts
```diff
@@ -736,9 +736,19 @@ function mockExample() {
...
 it('should refresh token if missing access token', (done) => {
   const scopes = mockExample();
-  client.credentials = {refresh_token: 'refresh-token-placeholder'};
+  let raisedEvent = false;
+  const refreshToken = 'refresh-token-placeholder';
+  client.credentials = {refresh_token: refreshToken};
+
+  // ensure the tokens event is raised
+  client.on('tokens', tokens => {
+    assert.equal(tokens.access_token, 'abc123');
```
alexander-fenster: would be great to have `abc123` in a variable and use it both in mock in the top of the file and in all asserts.

test/test.jwt.ts
JustinBeckwith: You win my best code review feedback of the day medal 🥇 

I simplified and fixed this up quite a bit.

PR #344, opened 2018-04-05
feat: allow passing options to auth.getClient

JustinBeckwith:
This change lets us stop pushing people towards createScopedRequired, while not breaking the API for those already using it.  It will also really simplify the getApplicationDefault scenario for users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "opn": "^5.2.0",
     "pify": "^3.0.0",
     "prettier": "^1.11.1",
-    "sinon": "^5.0.0",
+    "sinon": "^4.5.0",
     "source-map-support": "^0.5.3",
     "tmp": "^0.0.33",
     "typescript": "~2.7.2"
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
     "opn": "^5.2.0",
     "pify": "^3.0.0",
     "prettier": "^1.11.1",
-    "sinon": "^5.0.0",
+    "sinon": "^4.5.0",
```
ofrobots: Downgrade? why?

test/test.googleauth.ts
```diff
@@ -984,6 +984,7 @@ it('getApplicationDefault should use GCE when well-known file and env const are
          'compute-placeholder', res.credential.credentials.refresh_token);
    });
 
+
```
ofrobots: nit: unnecessary extra line.

package.json
JustinBeckwith: If you install 5.0, they say it's deprecated.  I checked `npm info sinon`, and it looks like they accidentally released it at one point, then quickly switched back the dist tag to 4.5. I could pull this out to another PR if you'd like :)

test/test.googleauth.ts
JustinBeckwith: fixed

package.json
ofrobots: Thanks for explaining. Sounds fine here.

PR #346, opened 2018-04-06
feat: add auth.request method

JustinBeckwith:
Another one that come up while trying to swap out google-auto-auth.  This way consumers can make a single call to get the client and make the request, with client caching.

### DIFF
--- a/test/test.googleauth.ts
+++ b/test/test.googleauth.ts
@@ -1295,3 +1295,14 @@ it('should get the current environment if GAE', async () => {
   const env = await auth.getEnv();
   assert.equal(env, envDetect.GCPEnv.APP_ENGINE);
 });
+
+it('should make the request', async () => {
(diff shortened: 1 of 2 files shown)


test/test.googleauth.ts
```diff
@@ -1295,3 +1295,14 @@ it('should get the current environment if GAE', async () => {
   const env = await auth.getEnv();
   assert.equal(env, envDetect.GCPEnv.APP_ENGINE);
 });
+
+it('should make the request', async () => {
+  const url = 'http://example.com';
+  const {auth, scopes} = mockGCE();
+  const data = {breakfast: 'coffee'};
```
alexander-fenster: just coffee? no eggs? :)

JustinBeckwith: 🤷‍♂️  I worked from home this morning, too lazy to make my own.

PR #349, opened 2018-04-11
Allow redirect_uri and client_id overrides in getToken opts

TimJohns:
The current `generateAuthUrl()` implementation allows for the `redirect_uri` opt to be overridden. Without the ability to specify the same override in the corresponding call to `getToken()`, the code exchange endpoint returns an error indicating a mismatch.

This change allows the `redirect_uri` and `client_id` used in the subsequent call to `getToken()` (and passed to the code exchange endpoint) to similarly be overridden.

A `redirect_uri` use case is straightforward, where a command line application (such as [clasp](https://github.com/google/clasp)) may wish to provide a localhost webserver to process the credentials returned from the code exchange endpoint, where the uri will contain a port number that was not known at the time the corresponding OAuth2Client was constructed. One current workaround is to re-assign the OAuth2Client.redirectUri private property. Another workaround is to re-construct the corresponding OAuth2Client object. 

While I know of no similar explicit use case for overriding the client_id, this change also addresses whatever the original use case was for overriding the client_id in the corresponding call to generateAuthUrl().

### DIFF
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -34,6 +34,19 @@ export enum CodeChallengeMethod {
 export interface GetTokenOptions {
   code: string;
   codeVerifier?: string;
+  /**
+   * The client ID for your application. The value passed into the constructor
+   * will be used if not provided. Must match any client_id option passed to
+   * a corresponding call to generateAuthUrl.
+   */
+  client_id?: string;
+  /**
+   * Determines where the API server redirects the user after the user
+   * completes the authorization flow. The value passed into the constructor
+   * will be used if not provided. Must match any redirect_uri option passed to
+   * a corresponding call to generateAuthUrl.
+   */
+  redirect_uri?: string;
 }
 
 export interface TokenInfo {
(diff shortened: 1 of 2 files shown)


src/auth/oauth2client.ts
```diff
@@ -34,6 +34,19 @@ export enum CodeChallengeMethod {
 export interface GetTokenOptions {
   code: string;
   codeVerifier?: string;
+  /**
+   * The client ID for your application. The value passed into the constructor
+   * will be used if not provided. Must match any client_id option passed to
+   * a corresponding call to generateAuthUrl.
+   */
+  client_id?: string;
```
JustinBeckwith: Let's use `clientId` here and `redirectUri` below, since we're dealing with property names not bound to the API.

JustinBeckwith:
@ofrobots would be great if you could take a look and keep me honest here

src/auth/oauth2client.ts
TimJohns: Thanks Justin - I had that thought originally as well, but then went with `redirect_uri` and `client_id` for consistency with the opts properties currently passed to `generateAuthUrl()`, where they similarly appear as-written within the URL query params. If you think it's unlikely they're being passed to `generateAuthUrl()` elsewhere in the wild, I'd feel better about changing them in BOTH places (i.e. [https://github.com/google/google-auth-library-nodejs/blob/master/src/auth/oauth2client.ts#L415-L416](here) as well) for the same reason. Consistency across the two would allow the calling code to be a little cleaner.

That said, there's some risk -- I'd hazard a guess they're probably not in wide use -- but no data to back that up, and if they are, could be a breaking change for somebody. I did just go looking and found references to`redirect_uri` in the comments in a couple of places in google-api-nodejs-client, but no executable code.  

I saw this PR was reviewed but not yet merged - I'd be inclined to suggest it as-is with the above in mind, but if the naming change is still needed, happy to do it; please let me know what you think about doing so only here, or in both places.

PR #350, opened 2018-04-13
test: fix possible EXDEV in fs.rename

alexander-fenster:
Making the same fix as in https://github.com/google/google-api-nodejs-client/pull/1030 to make the kitchen sync test run in case if `/tmp` is on a separate filesystem (which is quite common).


alexander-fenster:
Kitchen _sync_, I like this mistake

test/test.kitchen.ts
```diff
@@ -16,10 +16,12 @@
 
 import * as cp from 'child_process';
 import * as fs from 'fs';
+import * as mv from 'mv';
 import {ncp} from 'ncp';
 import * as pify from 'pify';
 import * as tmp from 'tmp';
 
+const mvp = pify(mv);
 const rename = pify(fs.rename);
```
JustinBeckwith: do we still need this `rename` reference?

test/test.kitchen.ts
```diff
@@ -54,7 +56,7 @@ it('should be able to use the d.ts', async () => {
   console.log(`${__filename} staging area: ${stagingPath}`);
   await spawnp('npm', ['pack']);
   const tarball = `${pkg.name}-${pkg.version}.tgz`;
-  await rename(tarball, `${stagingPath}/google-auth-library.tgz`);
+  await mvp(tarball, `${stagingPath}/google-auth-library.tgz`);
```
JustinBeckwith: Please add a brief comment explaining why we brought a module in to do this when fs.rename is right there :)

PR #353, opened 2018-04-24
chore: add nodejs 10 to test matrix

JustinBeckwith:
It found a bug!


.circleci/config.yml
```diff
@@ -81,7 +84,12 @@ jobs:
     <<: *unit_tests
   node9:
     docker:
-      - image: node:9
+      - image: node:8
```
jinwoo: This seems wrong?

JustinBeckwith: ooooops

PR #354, opened 2018-04-25
fix: add generic type to request

JustinBeckwith:


src/auth/googleauth.ts
```diff
@@ -708,9 +708,10 @@ export class GoogleAuth {
    * HTTP request using the given options.
    * @param opts Axios request options for the HTTP request.
    */
-  async request(opts: AxiosRequestConfig): Promise<AxiosResponse> {
+  // tslint:disable-next-line no-any
+  async request<T = any>(opts: AxiosRequestConfig): Promise<AxiosResponse<T>> {
     const client = await this.getClient();
-    return client.request(opts);
+    return client.request<T>(opts);
```
jinwoo: Don't we have a set of types for credential objects that we support?

JustinBeckwith: This is a type for a credential, it's a generic request API.  The `T` here defines the return type of the data coming from a user defined endpoint.  This truly could be anything.

jinwoo: Got it.

PR #361, opened 2018-05-14
fix: read credentials field before keyFilename

kjin:
In `google-auto-auth`, [credentials were being read before keyFilename](https://github.com/stephenplusplus/google-auto-auth/blob/master/index.js#L103).

I think in porting over features, that order was switched, so this PR switches it back.

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -657,12 +657,12 @@ export class GoogleAuth {
       this.jsonContent = options.credentials || this.jsonContent;
     }
     if (!this.cachedCredential) {
-      if (this.keyFilename) {
+      if (this.jsonContent) {
+        this.cachedCredential = await this.fromJSON(this.jsonContent);
+      } else if (this.keyFilename) {
         const filePath = path.resolve(this.keyFilename);
         const stream = fs.createReadStream(filePath);
         this.cachedCredential = await this.fromStreamAsync(stream);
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
I'm curious - how did this present itself as a bug?  If it was a breaking change to move one way, then it would probably be a breaking change the other direction.

ofrobots:
This needs a test to make sure the precedence is preserved going forward.

kjin:
@JustinBeckwith This behavior is specified on the [Trace Agent API](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/src/config.ts#L167) and tested [here](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/test/test-config-credentials.ts#L109). I wouldn't call it a breaking change since it is fixing a bug (that the behavior in this library deviates from the library it replaces), rather than reversing a previous breaking change.

@ofrobots Done

ofrobots:
If this is a regression, then the fix isn't semver major - depending on how much time has elapsed. Do we know which commit regressed and when it was released?

kjin:
@ofrobots It is #281, which was released on Feb 20.

However, users of `@google-cloud/common` don't see this change until 0.18, because 0.18 drops the dependency on `google-auto-auth`.

JustinBeckwith:
SGTM :)

PR #371, opened 2018-06-01
DO NOT MERGE: make it webpackable

alexander-fenster:
This is the first attempt to make the auth library webpack'able into the bundle of the normal size (159K in this case). This was achieved by removing all crypto stuff (that blows up the bundle to 400K) and replacing it with calls to browser's SubtleCrypto interface (thanks @JustinBeckwith for your help!)

TODO: tests. I only checked it in my local html file. We need to add some tests for browser functionality.


src/auth/oauth2client.ts
```diff
@@ -442,18 +457,19 @@ export class OAuth2Client extends AuthClient {
    * resulting SHA256. If used, this must be paired with a S256
    * code_challenge_method.
    */
-  generateCodeVerifier() {
+  async generateCodeVerifier() {
```
JustinBeckwith: Crikey! That's a semver major change 😱

src/auth/oauth2client.ts
```diff
@@ -705,6 +721,21 @@ export class OAuth2Client extends AuthClient {
...
+   * @param jsonp If set, append callback= parameter to request JSONP response.
+   */
+  static getRevokeTokenUrl(token: string, jsonp?: string): string {
+    let parameters = '';
+    if (jsonp) {
+      parameters = querystring.stringify({token, 'callback': jsonp});
+    } else {
+      parameters = querystring.stringify({token});
+    }
+    return OAuth2Client.GOOGLE_OAUTH2_REVOKE_URL_ + '?' + parameters;
```
JustinBeckwith: Lets use string interpolation here if we can.

src/auth/oauth2client.ts
```diff
@@ -705,6 +721,21 @@ export class OAuth2Client extends AuthClient {
     return {headers, res: r.res};
   }
 
+  /**
+   * Generates an URL to revoke the given token.
+   * @param token The existing token to be revoked.
+   * @param jsonp If set, append callback= parameter to request JSONP response.
+   */
+  static getRevokeTokenUrl(token: string, jsonp?: string): string {
```
JustinBeckwith: Does this really need to be public?

src/auth/oauth2client.ts
```diff
@@ -857,9 +885,14 @@ export class OAuth2Client extends AuthClient {
       throw new Error('The verifyIdToken method requires an ID Token');
     }
 
-    const response = await this.getFederatedSignonCertsAsync();
-    const login = this.verifySignedJwtWithCerts(
-        options.idToken, response.certs, options.audience,
+    let format: CertificateFormat = CertificateFormat.PEM;
```
JustinBeckwith: use a const and a ternary

src/auth/oauth2client.ts
```diff
@@ -892,32 +925,44 @@ export class OAuth2Client extends AuthClient {
...
    * Gets federated sign-on certificates to use for verifying identity tokens.
    * Returns certs as array structure, where keys are key ids, and values
-   * are PEM encoded certificates.
+   * are certificates in either PEM or JWK format.
    * @param callback Callback supplying the certificates
    */
   getFederatedSignonCerts(): Promise<FederatedSignonCertsResponse>;
-  getFederatedSignonCerts(callback: GetFederatedSignonCertsCallback): void;
-  getFederatedSignonCerts(callback?: GetFederatedSignonCertsCallback):
+  getFederatedSignonCerts(
```
JustinBeckwith: Another semver major change

src/auth/oauth2client.ts
```diff
@@ -892,32 +925,44 @@ export class OAuth2Client extends AuthClient {
...
+      format: CertificateFormat,
+      callback: GetFederatedSignonCertsCallback): void;
+  getFederatedSignonCerts(
+      format?: CertificateFormat, callback?: GetFederatedSignonCertsCallback):
       Promise<FederatedSignonCertsResponse>|void {
+    if (!format) {
```
JustinBeckwith: You can just set this in the function signature instead of making it optional and doing this check below:

`getFederatedSignonCerts(format: CertificateFormat = CertificateFormat.PEM ...`

src/auth/oauth2client.ts
```diff
@@ -1069,7 +1130,7 @@ export class OAuth2Client extends AuthClient {
    * @return The decoded string
    */
   decodeBase64(b64String: string) {
-    const buffer = new Buffer(b64String, 'base64');
+    const buffer = Buffer.from(b64String, 'base64');
```
JustinBeckwith: Oh yuck, we should clean this up in a CL like now.  Good fix.

src/crypto/browser/crypto.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import * as base64js from 'base64-js';
+import * as TextEncoding from 'text-encoding-shim';
+
+import {Crypto} from '../crypto';
+
+export class BrowserCrypto implements Crypto {
+  sha256DigestBase64(str: string): Promise<string> {
+    return new Promise(resolve => {
+      window.crypto.subtle
```
JustinBeckwith: A few comments would go a long way here :)

src/crypto/browser/jwkverifier.ts
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+
+import * as base64js from 'base64-js';
+import * as TextEncoding from 'text-encoding-shim';
+
+import {CertVerifier, JwkCertificate} from '../certverifier';
+
+export class JwkVerifier implements CertVerifier {
+  verify(pubkey: JwkCertificate, data: string, signature: string):
```
JustinBeckwith: Lets make this an async function, and use `await` instead of creating a new Promise and juggling `then` statements

webpack.config.js
```diff
@@ -0,0 +1,41 @@
+const path = require('path');
+const webpack = require('webpack');
+// const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
```
JustinBeckwith: just a nit to remove

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./node_modules/gts/tsconfig-google.json",
   "compilerOptions": {
+    "lib": ["es2015", "dom"],
```
JustinBeckwith: 🎉

webpack.config.js
```diff
@@ -0,0 +1,41 @@
...
+    ]
+  },
+  mode: 'production',
+  plugins: [
+    new webpack.DefinePlugin({
+      'process.env': {
+        'WEBPACK': 'true',
+      }
+    })
+//  , new BundleAnalyzerPlugin()
```
JustinBeckwith: nit for cleanup

JustinBeckwith:
I know you're on it, but I think we should make sure to include browser tests :)

src/auth/oauth2client.ts
alexander-fenster: I would better make it public to make it possible to assign the URL to some button, etc. Can be discussed though.

src/auth/oauth2client.ts
alexander-fenster: still learning typescript :)

src/auth/oauth2client.ts
alexander-fenster: Well, it must be public. The server does not send CORS headers for revoking the token, so the only way to call it from browser is via JSONP (by adding a `<script>` tag that will point to that URL), so the URL must be exposed by the library.
image

Other calls (such as getting a token) return CORS headers so we don't need to expose their URLs. If I find more URLs without CORS I'll need to expose them as well.

alexander-fenster: An alternative is to make a function that will append `<script>` tag and make all JSONP magic happen under the hood (like JQuery's `$.ajax({ jsonp: functionName })` call does), but I don't like this idea since I feel it's very wrong for this auth library to make any assumptions about HTML document structure and spam with extra `<script>` tags.

MrEmelianenko:
+1

You did a great job, thank you! Looking forward to seeing this in `master`.

alexander-fenster:
Thanks @MrEmelianenko, I paused it for a while to get some other stuff done but expect to go back to it within one or two weeks, add some Karma tests and finally get it to master (which will make [googleapis](https://github.com/google/google-api-nodejs-client) work in browser almost for free).

bobtrahan:
Stoked to see this hit master. I have been on a quest to get this to work in an application for a few days and finally realized what was going on.

alexander-fenster:
For those who are interested in this - some updates and merging will happen very soon! Please stay on the line :)

SPY:
@alexander-fenster sorry for my obtrusiveness, but is there any news about merge?

alexander-fenster:
@SPY It's my first priority for this/next week. Going to bring it up to date, add browser tests, and send it out for the code review.

chengjia0807:
Any good news?

alexander-fenster:
@chengjia0807 All tests pass now!  Just need to add some tests for browser functionality - working on it now.

JustinBeckwith:
👋 @alexander-fenster FYI - since this has breaking changes, I re-targeted it to the `next` branch.  I want to try and track all 3.x stuff over there until we feel ready to go.  Is this still on track for this year?

alexander-fenster:
@JustinBeckwith It is targeted for this year. The code is complete, I'm playing with Karma locally and will add it to this PR soon. Actually, you can go ahead and re-review the code since I will only add tests to this.

alexander-fenster:
One question here, can I still have Kokoro running tests on this PR?

JustinBeckwith:
Eugh.  I didn't know kokoro wouldn't run against non-master branches.  Uh... @theacodes @kinwa91?

alexander-fenster:
I just realized that adding `kokoro:force-run` helps here so I'm fine now :)

ammaarpatel99:
Just wondering when this is likely to be workable and merged for general use. I am making angular application which uses Angular Universal server side rendering (with an express server) and therefore requires webpack. However I can't use this on GCP app engine because when using client libraries I always get errors when compiling. If it won't be ready anytime soon but there is another way to make it work I would greatly appreciate been told. Thanks

alexander-fenster:
@ammaarpatel99 Which client library are you using? We do want to make it workable, if you give examples of what you're trying to do I can take a look what's going wrong before it gets merged to master and released.

alexander-fenster:
Bundle composition:

image

Total bundle size: 181K.

src/isbrowser.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2014 Google Inc. All Rights Reserved.
+ * Copyright 2017 Google Inc. All Rights Reserved.
```
JustinBeckwith: 2017 seems somewhat random :)

system-test/test.kitchen.ts
```diff
@@ -43,6 +48,19 @@ describe('pack and install', () => {
     await execa('npm', ['install'], {cwd: `${stagingPath}/`, stdio: 'inherit'});
   });
 
+  it('should be able to webpack the library', async function() {
+    if (process.version.match(/^v[46]\./)) {
```
JustinBeckwith: We only run system tests on 8 or LTS right now, so you shouldn't need this

test/test.oauth2.ts
```diff
@@ -95,18 +95,19 @@ it('should throw an error if generateAuthUrl is called with invalid parameters',
...
-  const match = codes.codeVerifier.match(/[a-zA-Z0-9\-\.~_]*/);
-  assert(match);
-  if (!match) return;
-  assert(match.length > 0 && match[0] === codes.codeVerifier);
-});
+it('should generate a valid code verifier and resulting challenge',
+   async () => {
+     const codes = await client.generateCodeVerifierAsync();
+     // ensure the code_verifier matches all requirements
+     assert.strictEqual(codes.codeVerifier.length, 128);
```
JustinBeckwith: use `assert.match` from chai

test/test.oauth2.ts
```diff
@@ -130,15 +131,15 @@ it('should verifyIdToken properly', async () => {
...
-        assert.strictEqual(JSON.stringify(certs), JSON.stringify(fakeCerts));
-        assert.strictEqual(requiredAudience, audience);
-        assert.strictEqual(theMaxExpiry, maxExpiry);
-        return new LoginTicket('c', payload);
-      };
+  client.verifySignedJwtWithCertsAsync = async (
+      jwt: string, certs: {}, requiredAudience: string|string[],
+      issuers?: string[], theMaxExpiry?: number) => {
+    assert.strictEqual(jwt, idToken);
+    assert.strictEqual(JSON.stringify(certs), JSON.stringify(fakeCerts));
```
JustinBeckwith: can we just `assert.deepStrictEqual` here instead of `JSON.stringify`?

tsconfig.json
```diff
@@ -1,13 +1,23 @@
...
     "rootDir": ".",
-    "outDir": "build"
+    "outDir": "build",
+    "target": "es5",
```
JustinBeckwith: I don't think we should set `es5` here.  Lets just let that float with the default?

tsconfig.json
```diff
@@ -1,13 +1,23 @@
...
+    "sourceMap": true
```
JustinBeckwith: That's emitted by default.  You should be able to drop this.

tsconfig.json
```diff
@@ -1,13 +1,23 @@
...
     "src/**/*.ts",
     "test/*.ts",
-    "system-test/*.ts"
+    "test/**/*.ts",
+    "system-test/*.ts",
+    "browser-test/*.ts",
+    "browser-test/**/*.ts"
+  ],
+  "exclude": [
+    "node_modules",
```
JustinBeckwith: You shouldn't need to include node_modules

JustinBeckwith:
So that profile you posted is interested!  `semver` should *not* be bundled with the client, and it looks like a big chunk.  The code that uses `semver` is checking `process.version`, so I am surprised it's working at all.  We may need to do something explicit to prevent that?

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,222 @@
...
+    assert.strictEqual(query.response_type, 'code token');
+    assert.strictEqual(query.access_type, ACCESS_TYPE);
+    assert.strictEqual(query.scope, SCOPE);
+    assert.strictEqual(query.client_id, CLIENT_ID);
+    assert.strictEqual(query.redirect_uri, REDIRECT_URI);
+    done();
+  });
+
+  it('getToken should work', async () => {
+    const now = (new Date()).getTime();
```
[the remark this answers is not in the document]
alexander-fenster: This was just copied from unit tests :) Fixed here (but not there).

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,222 @@
...
+    const stub = sinon.stub(client.transporter, 'request');
+    // @ts-ignore TS2345
+    stub.returns(Promise.resolve(
+        {data: {access_token: 'abc', refresh_token: '123', expires_in: 10}}));
+    client.getToken('code here', (err, tokens) => {
+      assert(tokens!.expiry_date! >= now + (10 * 1000));
```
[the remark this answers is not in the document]
alexander-fenster: Okay!

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,222 @@
...
+    assert.deepStrictEqual(result1.certs, result2.certs);
+    assert.deepStrictEqual(result1.format, result2.format);
+  });
+
+  it('should generate a valid code verifier and resulting challenge',
+     async () => {
+       const codes = await client.generateCodeVerifierAsync();
+       // ensure the code_verifier matches all requirements
+       assert.strictEqual(codes.codeVerifier.length, 128);
+       const match = codes.codeVerifier.match(/[a-zA-Z0-9\-\.~_]*/);
```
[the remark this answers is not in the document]
alexander-fenster: Yes, we can even remove the length check and move it to the regex :)

tsconfig.json
alexander-fenster: Yes, that's a leftover from something else :)

alexander-fenster:
@JustinBeckwith One more look? :)

alexander-fenster:
Re: changelog

From what I understand, for the basic usage of auth library, there should be no changes. Two functions were removed from `OAuth2Client`: `generateCodeVerifier` and `verifySignedJwtWithCerts` (replaced with `generateCodeVerifierAsync` and `verifySignedJwtWithCertsAsync`).

I'm trying to prepare a webpacked version of google-api-nodejs-client with this latest state of auth library, I'll link it to this PR when it works.

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,223 @@
...
+    assert.strictEqual(query.redirect_uri, REDIRECT_URI);
+    done();
+  });
+
+  it('getToken should work', async () => {
+    const now = Date.now();
+    const stub =
+        sinon
+            .stub(client.transporter, 'request')
+            // @ts-ignore TS2345
```
JustinBeckwith: So what's up with this ignore?

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,223 @@
...
+            .resolves({
+              data:
+                  {access_token: 'abc', refresh_token: '123', expires_in: 10}
+            });
+    client.getToken('code here', (err, tokens) => {
```
JustinBeckwith: Looks like a test bug.  It's an async method that is using a callback.  Can you just use the async `getToken`?

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,223 @@
...
+      assert.isAbove(tokens!.expiry_date!, now + (10 * 1000));
+      assert.isBelow(tokens!.expiry_date!, now + (15 * 1000));
+    });
+  });
+
+  it('getFederatedSignonCerts talks to correct endpoint', async () => {
+    const stub = sinon
+                     .stub(client.transporter, 'request')
+                     // @ts-ignore TS2345
```
JustinBeckwith: Same question as above, this is weird.

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,223 @@
...
+                     .resolves(FEDERATED_SIGNON_JWK_CERTS_AXIOS_RESPONSE);
+    const result = await client.getFederatedSignonCertsAsync();
+    assert.strictEqual(result.format, CertificateFormat.JWK);
+    assert.strictEqual(
```
JustinBeckwith: I think this line and the next can be summed up with `assert.deepStrictEqual`, no?

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,223 @@
...
+        Object.keys(result.certs).length, FEDERATED_SIGNON_JWK_CERTS.length);
+    for (const cert of FEDERATED_SIGNON_JWK_CERTS) {
+      assert.deepStrictEqual(result.certs[cert.kid], cert);
+    }
+  });
+
+  it('getFederatedSignonCerts caches certificates', async () => {
+    const stub = sinon
+                     .stub(client.transporter, 'request')
+                     // @ts-ignore TS2345
```
JustinBeckwith: WEIRD

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,223 @@
...
+       assert.match(codes.codeVerifier, /^[a-zA-Z0-9\-\.~_]{128}$/);
+     });
+
+  it('should include code challenge and method in the url', async () => {
+    const codes = await client.generateCodeVerifierAsync();
+    const authUrl = client.generateAuthUrl({
+      code_challenge: codes.codeChallenge,
+      code_challenge_method: CodeChallengeMethod.S256
+    });
+    const parsed = url.parse(authUrl);
```
JustinBeckwith: `url.parse` is deprecated:
https://nodejs.org/docs/latest/api/deprecations.html#deprecations_dep0116_legacy_url_api

Please use the new WHATWG Url class

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,223 @@
...
+    if (typeof parsed.query !== 'string') {
```
JustinBeckwith: This should just be an `assert.strictEqual`

browser-test/test.oauth2.ts
```diff
@@ -0,0 +1,223 @@
...
+      throw new Error('Unable to parse querystring');
+    }
+    const props = qs.parse(parsed.query);
+    assert.strictEqual(props.code_challenge, codes.codeChallenge);
+    assert.strictEqual(props.code_challenge_method, CodeChallengeMethod.S256);
+  });
+
+  it('should verify a valid certificate against a jwt', async () => {
+    const maxLifetimeSecs = 86400;
+    const now = new Date().getTime() / 1000;
```
JustinBeckwith: Just use `Date.now()`

src/auth/oauth2client.ts
```diff
@@ -933,11 +979,33 @@ export class OAuth2Client extends AuthClient {
...
+      case CertificateFormat.PEM:
+        certificates = res.data;
+        break;
+      case CertificateFormat.JWK:
+        for (const key of res.data.keys) {
+          certificates[key.kid] = key;
+        }
+        break;
+      default:
+        throw new Error('Unsupported certificate format ' + format);
```
JustinBeckwith: use string interpolation

src/crypto/browser/crypto.ts
```diff
@@ -0,0 +1,97 @@
...
+// This file implements crypto functions we need using in-browser
+// SubtleCrypto interface `window.crypto.subtle`.
+
+import * as base64js from 'base64-js';
+import {CryptoSigner} from '../crypto';
+
+// Not all browsers support `TextEncoder`. The following `require` will
+// provide a fast UTF8-only replacement for those browsers that don't support
+// text encoding natively.
+require('fast-text-encoding');
```
JustinBeckwith: Can we only require this if `TextEncoder` is not available?  Don't want to pay the import cost if we don't have to, and IE/Edge are a.) a very small % of browsers and b.) dying :)

test/test.oauth2.ts
```diff
@@ -95,18 +95,19 @@ it('should throw an error if generateAuthUrl is called with invalid parameters',
...
-  if (!match) return;
-  assert(match.length > 0 && match[0] === codes.codeVerifier);
-});
+it('should generate a valid code verifier and resulting challenge',
+   async () => {
+     const codes = await client.generateCodeVerifierAsync();
+     // ensure the code_verifier matches all requirements
+     assert.strictEqual(codes.codeVerifier.length, 128);
+     const match = codes.codeVerifier.match(/[a-zA-Z0-9\-\.~_]*/);
+     assert(match);
```
JustinBeckwith: Use assert.match

browser-test/test.oauth2.ts
alexander-fenster: I don't know :(  It works for me locally but the Kokoro build fails:

image

I have no idea where it sees `undefined`. The type definition looks good to me (it's `any` there).

browser-test/test.oauth2.ts
alexander-fenster: Fixed, thanks!

browser-test/test.oauth2.ts
alexander-fenster: The format of two objects to compare is a little bit different but I changed the test to make it more readable, with `assert.deepStrictEqual` at the end.

browser-test/test.oauth2.ts
alexander-fenster: Changed. This was copied from `test/test.oauth.ts` without any changes.

src/crypto/browser/crypto.ts
alexander-fenster: The module has this checks and it does nothing if `TextEncoder`/`TextDecoder` are available.

```js
(function(scope) {
'use strict';

// fail early
if (scope['TextEncoder'] && scope['TextDecoder']) {
  return false;
}
...
scope['TextEncoder'] = FastTextEncoder;
scope['TextDecoder'] = FastTextDecoder;

}(typeof window !== 'undefined' ? window : (typeof global !== 'undefined' ? global : this)));
```

alexander-fenster: But if you're talking about the import costs, I can surely do that :)

test/test.oauth2.ts
JustinBeckwith: 👍

JustinBeckwith:
@alexander-fenster this is looking awesome.  After you add details on the breaking changes to the description, I can approve this and we can start cranking on v3 💃

browser-test/test.oauth2.ts
alexander-fenster: no `URL` class in webpack yet: https://github.com/webpack/node-libs-browser/issues/69

alexander-fenster:
@JustinBeckwith I updated the PR description above.

browser-test/test.oauth2.ts
```diff
@@ -111,25 +110,22 @@ describe('Browser OAuth2 tests', () => {
...
     };
 
     const generated = client.generateAuthUrl(opts);
-    const parsed = url.parse(generated);
-    if (typeof parsed.query !== 'string') {
-      throw new Error('Unable to parse querystring');
-    }
-    const query = qs.parse(parsed.query);
+    // can't use URL class in webpack, so parsing URL manually
+    const queryString = generated.replace(/^.*?\?(.*?)(?:#.*)?$/, "$1");
```
JustinBeckwith: wait wat - WHATWG Url is the new hot in node because it's a browser class.
https://developer.mozilla.org/en-US/docs/Web/API/URL

JustinBeckwith:
@alexander-fenster I'll let you do the honors of merging :)

alexander-fenster:
I. Cannot. Believe. It. Is. Merged.

Thanks @JustinBeckwith :)

dmytrobanasko:
Facing an [issue](https://github.com/JustinBeckwith/gaxios/issues/44) after switching from webbbpackkk branch to latest release (actually, after axios -> gaxios)
@alexander-fenster, did you try it with new client ?

alexander-fenster:
Let's discuss it there to keep the conversation in one place :)  Thanks for your report!

PR #377, opened 2018-06-04
chore: fix lint task in circle config

alexander-fenster:
It never worked (wrong syntax, `requires` can be in workflow, not in a job description), but Circle never complained. Actually, it complained only when I tried to rerun something with SSH:

image


JustinBeckwith:
Oooooops.  Thank you :)

PR #378, opened 2018-06-05
feat: enable custom service account with compute client

JustinBeckwith:
This adds the ability to specify which service account the developer wants to use when using a `Compute` auth client.  This came up in the context of a missing tutorial.

### DIFF
--- a/examples/compute.js
+++ b/examples/compute.js
@@ -1,0 +1,31 @@
+// Copyright 2018, Google, LLC.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


examples/compute.js
```diff
@@ -0,0 +1,31 @@
...
+'use strict';
+
+const { Compute } = require('google-auth-library');
+
+/**
+ * Acquire a client, and make a request to an API that's enabled by default.
+ */
+async function main() {
+  const client = new Compute({
+    serviceAccountEmail: 'beckwith@google.com'
```
theacodes: please don't hard-code real email addresses.

JustinBeckwith: But I love getting emails from my node.js on GCP friends 🤷‍♂️

test/test.compute.ts
```diff
@@ -260,3 +260,18 @@ it('should return a helpful message on token refresh response.statusCode 404',
      }
      throw new Error('Expected to throw');
    });
+
+it('should accept a custom service account', async () => {
+  const serviceAccountEmail = 'beckwith@google.com';
```
fhinkel: Do you really want your real email address here?

PR #397, opened 2018-06-28
test: run tests with hard-rejection

ofrobots:
Ref: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2839


test/mocha.opts
```diff
@@ -0,0 +1,3 @@
+--require hard-rejection/register
+--require source-map-support/register
+--timeout 10000
```
JustinBeckwith: are you sure that mocha lets you pass the require flag twice like this?  I honestly don't know.

ofrobots: Yeah, that works. The `--require` option is a pass-through to node.

PR #402, opened 2018-07-02
fix: deprecate the `getDefaultProjectId` method

JustinBeckwith:


src/messages.ts
```diff
@@ -16,8 +16,12 @@
...
-  https://cloud.google.com/docs/authentication/.`;
+Cloud SDK. We recommend that most server applications use service accounts
+instead. If your application continues to use end user credentials from Cloud
+SDK, you might receive a "quota exceeded" or "API not enabled" error. For
+more information about service accounts, see
+https://cloud.google.com/docs/authentication/.`;
+
+export const DEFAULT_PROJECT_ID_DEPRECATED =
+    `The 'getDefaultProjectId' method has been deprecated, and will be removed
+ in the 3.0 release of this library. Please use the 'getProjectId' method instead.`;
```
ofrobots: The warning will reach the end user and it may not be clear what 'this library' means. It would be just a disembodied warning message without any context. Maybe spell out the library name instead?

Secondly, apart from a small number users who are using this library directly, most users of this library get this as a dependency of another Google Cloud client library. For those use-cases, I think it is a fatal error if we client library uses a deprecated method, and we should try out best to ensure this warning doesn't leak to the end user.  Are we doing anything to ensure that the client libraries do not end up using any deprecated APIs?

src/auth/googleauth.ts
```diff
@@ -126,23 +123,36 @@ export class GoogleAuth {
...
-   * Obtains the default project ID for the application.
-   * @param callback Optional callback
-   * @returns Promise that resolves with project Id (if used without callback)
+   * THIS METHOD HAS BEEN DEPRECATED.
+   * It will be removed in 3.0.  Please use getProjectId instead.
    */
   getDefaultProjectId(): Promise<string>;
   getDefaultProjectId(callback: ProjectIdCallback): void;
   getDefaultProjectId(callback?: ProjectIdCallback): Promise<string|null>|void {
+    process.emitWarning(messages.DEFAULT_PROJECT_ID_DEPRECATED);
```
ofrobots: I would like this to be done like this:

```ts
const deprecation = {
  message: messages.DEFAULT_PROJECT_ID_DEPRECATED,
  code: 'google-auth-library:DEP001'
};
if (semver.satisfies(process.version, ">=8")) {
  process.emitWarning(deprecation.message, 'DeprecationWarning', deprecation.code);
} else {
  // This path can be removed once we drop support for Node 6.
  process.emitWarning(`${deprecation.code}: ${deprecation.message}`);
}
```

This addresses the comment that I added below when used with the `--throw-deprecations` flag in the client library tests.

JustinBeckwith: Awesome feedback - thank you!  I updated the code to (mostly) do this (the contract on `emitWarning` is a little different).

src/messages.ts
JustinBeckwith: I updated this to use `google-auth-library` instead :) On the second point - should we run the system tests with `--throw-deprecations` flag?

src/messages.ts
```diff
@@ -14,10 +14,48 @@
...
+
+export interface Warning {
+  code: string;
+  type: string;
+  message: string;
+}
+
+export const PROBLEMATIC_CREDENTIALS_WARNING = {
+  code: 'google-auth-library:DEP001',
+  type: 'DeprecationWarning',
```
ofrobots: This is not a deprecation, just a plain warning. We are not removing or changing functionality. We are just letting users know that they might be using problematic default credentials.

With this in mind, maybe we should drop the `DEP` infix in the code? `google-auth-library:0001`?

JustinBeckwith:
@ofrobots lets try this again :)

src/auth/googleauth.ts
```diff
@@ -126,23 +123,36 @@ export class GoogleAuth {
...
    */
   getDefaultProjectId(): Promise<string>;
   getDefaultProjectId(callback: ProjectIdCallback): void;
   getDefaultProjectId(callback?: ProjectIdCallback): Promise<string|null>|void {
+    messages.warn(messages.DEFAULT_PROJECT_ID_DEPRECATED);
     if (callback) {
-      this.getDefaultProjectIdAsync()
-          .then(r => callback(null, r))
-          .catch(callback);
+      this.getProjectIdAsync().then(r => callback(null, r)).catch(callback);
```
ofrobots: Egads 🙀!

While you're here, fix the `promise.then(f(cb)).catch(cb)` land-mine, or open a bug.

src/auth/googleauth.ts
```diff
@@ -126,23 +123,36 @@ export class GoogleAuth {
...
+  /**
+   * Obtains the default project ID for the application.
+   * @param callback Optional callback
+   * @returns Promise that resolves with project Id (if used without callback)
+   */
+  getProjectId(): Promise<string>;
+  getProjectId(callback: ProjectIdCallback): void;
+  getProjectId(callback?: ProjectIdCallback): Promise<string|null>|void {
+    if (callback) {
+      this.getProjectIdAsync().then(r => callback(null, r)).catch(callback);
```
ofrobots: and here.

src/auth/googleauth.ts
JustinBeckwith: Yeah, ooops :) Filed #403 to track.

PR #405, opened 2018-07-04
fix: support HTTPS proxies

### LINKED ISSUE #352: Add support for proxies
Axios does not properly support proxies (https://github.com/axios/axios/issues/658). Can Axios be replaced with a library that does?

JustinBeckwith:
Uses [https-proxy-agent](https://github.com/TooTallNate/node-https-proxy-agent) to provide HTTPS proxy support, making up for the fact that axios doesn't support it natively yet.  

I don't have an HTTPS proxy sitting around where I can test this, so I am really throwing stuff at the wall here. 

Also of note - there is no reason I can possibly come up with why we would ever support an `HTTP` proxy.  Every endpoint we're going to hit for a Google service will be `HTTPS`, so I want to explicitly remove that from the documentation. 

Fixes #352

### DIFF
--- a/src/transporters.ts
+++ b/src/transporters.ts
@@ -17,6 +17,9 @@
 import axios, {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios';
 import {validate} from './options';
 
+// tslint:disable-next-line variable-name
+const HttpsProxyAgent = require('https-proxy-agent');
+
 // tslint:disable-next-line no-var-requires
 const pkg = require('../../package.json');
 const PRODUCT_NAME = 'google-api-nodejs-client';
@@ -92,6 +95,13 @@ export class DefaultTransporter {
       } else {
         throw e;
       }
+    }
+
+    // If the user configured an `HTTPS_PROXY` environment variable, create
+    // a custom agent to proxy the request.
+    const proxy = process.env.HTTPS_PROXY || process.env.https_proxy;
+    if (proxy) {
+      opts.httpsAgent = new HttpsProxyAgent(proxy);
     }
 
     if (callback) {
--- a/src/transporters.ts
+++ b/src/transporters.ts
@@ -102,6 +102,7 @@ export class DefaultTransporter {
     const proxy = process.env.HTTPS_PROXY || process.env.https_proxy;
     if (proxy) {
       opts.httpsAgent = new HttpsProxyAgent(proxy);
+      opts.proxy = false;
     }
 
     if (callback) {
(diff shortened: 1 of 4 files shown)


src/transporters.ts
```diff
@@ -95,6 +98,13 @@ export class DefaultTransporter {
       }
     }
 
+    // If the user configured an `HTTPS_PROXY` environment variable, create
+    // a custom agent to proxy the request.
+    const proxy = process.env.HTTPS_PROXY || process.env.https_proxy;
+    if (proxy) {
+      opts.httpsAgent = new HttpsProxyAgent(proxy);
```
ofrobots: If/Once client libraries start using axios, they will need the same pattern. Should we make an axios wrapper that everyone can use instead of axios?

yonilerner: In my testing, this only works if you also pass `proxy: false` as an option

yonilerner: When I change the code to this, it works:
```javascript
        const proxy = process.env.HTTPS_PROXY || process.env.https_proxy;
        if (proxy) {
            opts.httpsAgent = new HttpsProxyAgent(proxy);
            opts.proxy = false
        }
```

yonilerner: Actually, I might be wrong. Retesting

yonilerner: OK I _was_ right.

JustinBeckwith: Can you share how you're testing this?

yonilerner: We have some proxy server running somewhere with access logs. Ive set `HTTPS_PROXY=http://localhost:31280` (Im using SSH port-forwarding to access it) in the environment for my code.

When I use my original hack from #352 , I can connect to Google and I see the access logged on the proxy server. That hack adds to the axios config: 
```typescript
{
        httpsAgent,
        proxy: false
    }
```

When I `npm install 'JustinBeckwith/google-auth-library-nodejs#proxy'` and comment out my hack, the proxy returns an error.

If I then go into `node_modules/google-auth-library-nodejs/build/src/transporters.js` and add `opts.proxy = false` after your initial `opts.httpsAgent = new HttpsProxyAgent(proxy);`, I can once again connect to Google and the proxy logs the access.

fhinkel:
@JustinBeckwith , why did we close this without merging? Did we realize this is not going to work? Looking through the comments this looks promising to me.

JustinBeckwith:
My thought at the time was that this wouldn't be effective since it *only* applies to the auth library, not every library we have.  That having been said, it's worth reconsidering!

fhinkel:
Which other libraries use `axios`? It's only a handful, right?

JustinBeckwith:
Fair point :)

PR #409, opened 2018-07-06
fix: use gcp-metadata for compute credentials

JustinBeckwith:
**BREAKING CHANGE**:  The `OAuth2Client`, `JWT`, and `Compute` classes no longer accept parameters to override the authBaseUrl and tokenUrl.  These properties were largely available for testing, but there are better ways to test.  

#### Old code

```js
client = new OAuth2Client(CLIENT_ID, CLIENT_SECRET, REDIRECT_URI, {authBaseUrl, tokenUrl});	
```

#### New code

```js
client = new OAuth2Client(CLIENT_ID, CLIENT_SECRET, REDIRECT_URI);	
```

### DIFF
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-import axios, {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios';
+import {AxiosError, AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios';
 import * as gcpMetadata from 'gcp-metadata';
-import * as rax from 'retry-axios';
+
 import {CredentialRequest, Credentials} from './credentials';
 import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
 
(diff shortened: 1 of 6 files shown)


src/auth/computeclient.ts
```diff
@@ -28,13 +28,12 @@ export interface ComputeOptions extends RefreshOptions {
...
 }
 
-// Create a scoped axios instance that will retry 3 times by default
-const ax = axios.create();
-rax.attach(ax);
-
 export class Compute extends OAuth2Client {
   private serviceAccountEmail: string;
 
+  // Google Compute Engine metadata server token endpoint.
```
kjin: Unused comment?

test/test.googleauth.ts
```diff
@@ -18,7 +18,7 @@ import assert from 'assert';
 import child_process from 'child_process';
 import crypto from 'crypto';
 import * as fs from 'fs';
-import {BASE_PATH, HEADER_NAME, HOST_ADDRESS} from 'gcp-metadata';
+import {BASE_PATH, HEADERS, HOST_ADDRESS} from 'gcp-metadata';
```
kjin: Are `HEADER_NAME` and `HEADERS` both available on the same version of `gcp-metadata`? (If not, I think you'd need to update that dependency?)

JustinBeckwith: They're both available at the same time :)

PR #410, opened 2018-07-06
fix: deprecate the `createScopedRequired` methods

JustinBeckwith:
BREAKING CHANGE:  The `createScopedRequired` method has been deprecated on multiple classes.  The `createScopedRequired` and `createScoped` methods on the `JWT` class were largely in place to help inform clients when scopes were required in an application default credential scenario.  Instead of checking if scopes are required after creating the client, instead scopes should just be passed either into the `GoogleAuth.getClient` method, or directly into the `JWT` constructor.  

### Old code
```js
auth.getApplicationDefault(function(err, authClient) {
   if (err) {
     return callback(err);
   }
  if (authClient.createScopedRequired && authClient.createScopedRequired()) {
    authClient = authClient.createScoped([
      'https://www.googleapis.com/auth/cloud-platform'
    ]);
  }
  callback(null, authClient);
});
```

### New code
```js
const client = await auth.getClient({
  scopes: ['https://www.googleapis.com/auth/cloud-platform']
});
```

### DIFF
--- a/src/auth/iam.ts
+++ b/src/auth/iam.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import * as messages from '../messages';
+
 export interface RequestMetadata {
   'x-goog-iam-authority-selector': string;
   'x-goog-iam-authorization-token': string;
@@ -40,6 +42,7 @@ export class IAMAuth {
    */
   createScopedRequired() {
     // IAM authorization does not use scopes.
+    messages.warn(messages.IAM_CREATE_SCOPED_DEPRECATED);
     return false;
   }
 
--- a/src/auth/iam.ts
+++ b/src/auth/iam.ts
@@ -37,7 +37,7 @@ export class IAMAuth {
   /**
    * Indicates whether the credential requires scopes to be created by calling
    * createdScoped before use.
-   *
+   * @deprecated
    * @return always false
    */
   createScopedRequired() {
--- a/src/auth/jwtclient.ts
+++ b/src/auth/jwtclient.ts
@@ -16,6 +16,9 @@
 
 import {GoogleToken} from 'gtoken';
 import * as stream from 'stream';
+
+import * as messages from '../messages';
+
 import {CredentialBody, Credentials, JWTInput} from './credentials';
 import {JWTAccess} from './jwtaccess';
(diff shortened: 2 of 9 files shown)


JustinBeckwith:
*gentle ping* @google/google-node-team @google/yoshi-nodejs

src/auth/jwtclient.ts
```diff
@@ -119,19 +122,27 @@ export class JWT extends OAuth2Client {
...
-      // For others, convert to a string and check the length.
-      return String(this.scopes).length === 0;
+    messages.warn(messages.JWT_CREATE_SCOPED_DEPRECATED);
+    return !this.hasScopes();
+  }
+
+  /**
+   * Determine if there are currently scopes available.
+   */
+  protected hasScopes() {
```
ofrobots: Can this be private?

src/auth/iam.ts
```diff
@@ -40,6 +42,7 @@ export class IAMAuth {
    */
   createScopedRequired() {
```
ofrobots: You missed the `@deprecated` annotation in the comment. Since you added it elsewhere, add it here too.

src/messages.ts
```diff
@@ -70,9 +70,45 @@ export const DEFAULT_PROJECT_ID_DEPRECATED = {
...
+export const COMPUTE_CREATE_SCOPED_DEPRECATED = {
   code: 'google-auth-library:DEP003',
   type: WarningTypes.DEPRECATION,
+  message: [
+    'The `createScopedRequired` method on the `Compute` class has been deprecated,',
+    'and will be removed in the 3.0 release of google-auth-library.'
+  ].join(' ')
+};
+
+export const JWT_CREATE_SCOPED_DEPRECATED = {
```
ofrobots: nit: perhaps a single deprecation was sufficient. `createScopeRequired` has been deprecated on _all_ classes.

JustinBeckwith: Ah - let me explain!  The rationale is that we only warn once for each deprecation.  I wanted to make sure that we warn once for each separate API the user is off and using.  I could see this one both ways, so I err'd on the side of over-communicating.

ofrobots: 👍

PR #414, opened 2018-07-07
chore: deprecate getRequestMetadata

JustinBeckwith:
BREAKING CHANGE:  The `getRequestMetadata` method has been deprecated on the `IAM`, `OAuth2`, `JWT`, and `JWTAccess` classes.  The `getRequestHeaders` method should be used instead.  The methods have a subtle difference:  the `getRequestMetadata` method returns an object with a headers property, which contains the authorization header.  The `getRequestHeaders` method simply returns the headers.  

### Old code
```js
const res = await client.getRequestMetadata();
const headers = res.headers;
```

### New code
```js
const headers = await client.getRequestHeaders();
```

### DIFF
--- a/src/auth/jwtaccess.ts
+++ b/src/auth/jwtaccess.ts
@@ -14,14 +14,17 @@
  * limitations under the License.
  */
 
+import {IncomingHttpHeaders} from 'http';
 import jws from 'jws';
 import LRU from 'lru-cache';
 import * as stream from 'stream';
-
 import * as messages from '../messages';
-
 import {JWTInput} from './credentials';
 import {RequestMetadataResponse} from './oauth2client';
+
+export type Claims = {
+  [index: string]: string
+};
 
 export class JWTAccess {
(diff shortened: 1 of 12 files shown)


src/auth/jwtaccess.ts
```diff
@@ -14,19 +14,25 @@
  * limitations under the License.
  */
 
+import {IncomingHttpHeaders} from 'http';
```
kjin: I'm wondering if we really need to use `IncomingHttpHeaders` if we already know that the headers are of type `type T = { [key: string]: string }`. If that's true I would suggest using that because `T` is a subtype of `IncomingHttpHeaders` and you don't have to cast values as strings in tests.

JustinBeckwith: That's an entirely fair point 👍

JustinBeckwith:
*gentle ping* @google/google-node-team @google/yoshi-nodejs

PR #416, opened 2018-07-09
fix: use forked axios 😧

JustinBeckwith:
This uses a forked version of axios that has all of the commits on master added

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,21 +5,21 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.49",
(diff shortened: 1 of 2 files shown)


ofrobots:
This is all sorts of worrisome. 🙀

ofrobots:
Hmm.. maybe this is for experimentation only.. before a release is cut on axios? In which case I rescind the worry-cat.

JustinBeckwith:
Your worry cat is for good reason.  The owners have not been responding to issues.  I have the ability to commit to master, but I *don't* have the keys to `npm publish`.  
https://github.com/axios/axios/issues

The library is stable enough? right now that I'm not freaking out, but yeah, this ain't a good look.

PR #398, opened 2018-06-29
chore: update gcp-metadata to pick up fix

ofrobots:
gcp-metadata no longer swallows unexpected errors.
Refs: stephenplusplus/gcp-metadata#57

PR #420, opened 2018-07-11
chore: update gcp-metadata for isAvailable fix

ofrobots:
The `isAvailable` function no longer swallows unexpected errors which
means that we can start reporting better errors back to users of this
library.

Previously #398.

### DIFF
--- a/test/test.googleauth.ts
+++ b/test/test.googleauth.ts
@@ -67,14 +67,10 @@ function nockIsGCE() {
 }
 
 function nockNotGCE() {
-  return nock(host).get(instancePath).replyWithError({code: 'ETIMEDOUT'});
-}
-
-function nockENOTFOUND() {
   return nock(host).get(instancePath).replyWithError({code: 'ENOTFOUND'});
 }
 
-function nockErrGCE() {
+function nock500GCE() {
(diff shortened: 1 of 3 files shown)


test/test.googleauth.ts
```diff
@@ -1021,30 +1017,30 @@ it('_checkIsGCE should not set the _isGCE flag when not running on GCE',
      scope.done();
    });
 
-it('_checkIsGCE should retry the check for isGCE if it fails the first time',
+it('_checkIsGCE should retry the check for isGCE on transient http errors',
```
JustinBeckwith: So honest question - do we think this test belongs here?  I could see the argument for deleting the test, and trusting the tests in `gcp-metadata`.

JustinBeckwith: Same question for the tests below.

ofrobots: Some of these tests do belong here – specifically the tests that ensure we cache the answer from the first time around. Others can be potentially removed, but I don't have strong feelings. I can remove them if you prefer.

JustinBeckwith: 🤷‍♂️ s'all good

PR #429, opened 2018-07-24
fix: update env var for detecting gcf

kjin:
`X_GOOGLE_FUNCTION_NAME` was introduced sometime earlier this year. It seems to contain the same value as `FUNCTION_NAME` and probably functions as a replacement; we should check this env var as well in case `FUNCTION_NAME` eventually disappears.

### DIFF
--- a/src/auth/envDetect.ts
+++ b/src/auth/envDetect.ts
@@ -52,7 +52,8 @@ function isAppEngine() {
 }
 
 function isCloudFunction() {
-  return !!process.env.FUNCTION_NAME;
+  // X_GOOGLE_FUNCTION_NAME is a new alias for FUNCTION_NAME.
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
@steren what's the word here?

JustinBeckwith:
The docs here say that `FUNCTION_NAME` is a perfectly valid way to check, so lets keep this as is for now :)

kjin:
Ok 👍

hdp617:
`FUNCTION_NAME` is no longer set in Google Cloud Functions for Node.js10 and subsequent runtimes. Could we update this?
https://cloud.google.com/functions/docs/env-var#environment_variables_set_automatically
It looks like `K_SERVICE` is the replacement.

ofrobots:
@hdp617 Thanks for reporting. As this is a closed bug, I have moved your comment to #747.

PR #434, opened 2018-07-31
fix: Use new auth URIs

theacodes:


theacodes:
It's what the spec requests of us, so 🤷‍♀️

PR #450, opened 2018-08-18
Set private_key_id in JWT access token header like other google auth …

anuraaga:
…libraries.

When trying to use NodeJS gRPC to authenticate to an API, I found that NodeJS doesn't set `kid` in the JWT header like other google auth libraries (I've used Java, Python, Go, and C++ fine). This field propagates `private_key_id` to `kid` when present.

Also go ahead and set `typ` to fully reproduce the behavior in the other libraries.

The API is backwards compatible as it only adds optional parameters to the end of constructors. Behavior is significantly different, but it only adds header data in a way that matches other auth libraries, so probably doesn't warrant a semver bump.

For reference:

Java implementation
https://github.com/google/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java#L335

Go implementation
https://github.com/golang/oauth2/blob/master/google/jwt.go#L54

Spec for `kid` parameter
https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-11#section-4.1.7


anuraaga:
@JustinBeckwith would you be able to review this PR? Thanks!

JustinBeckwith:
@theacodes I'm not 100% up to speed on what's happening in here :)  Could you please take a look and make sure this makes sense?

theacodes:
This matches the Python library's behavior, which is canonical.

PR #451, opened 2018-08-18
Remove mention of CHANGELOG from CONTRIBUTING.md since there doesn't …

anuraaga:
... seem to be a changelog in the repo.


JustinBeckwith:
Thanks for the contribution!  The good news here is that we're actually adding a changelog in the next release :) The github template you're editing is generated, and we're trying to keep all this stuff consistent.

There will be a changelog in the 2.0 release!

anuraaga:
Cool, thanks for the update!

PR #467, opened 2018-09-07
Update omitted parameter 'hd'

Beingbook:
reference: https://developers.google.com/identity/protocols/OpenIDConnect#authenticationuriparameters

I'm not sure there is need test and how to test.
At this moment, my clients are using the option through casting as `any`.

- [X] Tests and linter pass


Beingbook:
I signed it!

PR #479, opened 2018-09-27
fix: export additional typescript types

JustinBeckwith:
I noticed that you're having to reach into files for types.  I should export them :)
https://github.com/googleapis/cloud-trace-nodejs/pull/871/files

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,13 +16,13 @@
 import {GoogleAuth} from './auth/googleauth';
 
 export {Compute, ComputeOptions} from './auth/computeclient';
-export {Credentials} from './auth/credentials';
+export {CredentialBody, CredentialRequest, Credentials, JWTInput} from './auth/credentials';
 export {GoogleAuthOptions} from './auth/googleauth';
-export {IAMAuth} from './auth/iam';
-export {JWTAccess} from './auth/jwtaccess';
-export {JWT} from './auth/jwtclient';
-export {CodeChallengeMethod, OAuth2Client} from './auth/oauth2client';
-export {UserRefreshClient} from './auth/refreshclient';
+export {IAMAuth, RequestMetadata} from './auth/iam';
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -16,13 +16,13 @@
...
-export {IAMAuth} from './auth/iam';
-export {JWTAccess} from './auth/jwtaccess';
-export {JWT} from './auth/jwtclient';
-export {CodeChallengeMethod, OAuth2Client} from './auth/oauth2client';
-export {UserRefreshClient} from './auth/refreshclient';
+export {IAMAuth, RequestMetadata} from './auth/iam';
+export {Claims, JWTAccess} from './auth/jwtaccess';
+export {JWT, JWTOptions} from './auth/jwtclient';
+export {Certificates, CodeChallengeMethod, GenerateAuthUrlOpts, GetTokenOptions, OAuth2Client, OAuth2ClientOptions, RefreshOptions, TokenInfo, VerifyIdTokenOptions} from './auth/oauth2client';
+export {UserRefreshClient, UserRefreshClientOptions} from './auth/refreshclient';
```
ofrobots: I don't know if this is actionable at this point, but it seems that the module export contains a large number of unrelated types without much name-spacing. For example it would have been preferable if we exported `credential.Body` and `credential.Request` rather than `CredentialBody` and `CredentialRequest`. Unwieldy.

JustinBeckwith: Well I mean, we could change it 🤷‍♂️   Are you saying I should be exporting namespaces with the types bundled in?

ofrobots: Potentially, yes. It seems weird to have lots of unrelated classes being exported through the default module export. For example `RequestMetadata` as a name doesn't mean anything without some context. Another example `IAMAuth` is not necessarily a better name than `iam.Auth`.

This was an observation on the design of the overall API of this library. I don't expect you to make the change as part of this PR.

PR #536, opened 2018-11-19
fix(deps): upgrade to gcp-metadata v0.9.0

JustinBeckwith:
BREAKING CHANGE: In cases where the `Compute` client is used, the the `getRequestHeaders` method will no longer return an `AxiosResponse` object if a request is made to the metadata server.  The value of the object will return null. This is not likely to affect any downstream APIs, but is a breaking change nonetheless.


package.json
```diff
@@ -38,7 +38,7 @@
     "@types/node": "^10.5.1",
     "@types/pify": "^3.0.2",
     "@types/semver": "^5.5.0",
-    "@types/sinon": "^5.0.1",
```
JustinBeckwith: Note: the current @types/sinon package is broken.

JustinBeckwith:
@ofrobots *gentle ping*

src/auth/computeclient.ts
```diff
@@ -66,21 +66,20 @@ export class Compute extends OAuth2Client {
...
-      tokens.expiry_date =
-          ((new Date()).getTime() + (res.data.expires_in * 1000));
+    const tokens = data as Credentials;
+    if (data && data.expires_in) {
+      tokens.expiry_date = ((new Date()).getTime() + (data.expires_in * 1000));
       delete (tokens as CredentialRequest).expires_in;
     }
     this.emit('tokens', tokens);
-    return {tokens, res};
+    return {tokens, res: null};
```
ofrobots: This seems like a behavior change?

JustinBeckwith: Ah, good catch :) I actually hate the way we return the responses all over the place with this API.  There was a breaking change to `gcp-metadata` where we stripped that stuff out, so it makes it hard/impossible to show here.

Do you think this is enough of a change to merit a semver major release?

ofrobots: Semver major is about communicating changes, and IMO the size of the change shouldn't have an impact on the semver-ness. If this is publicly exposed API, it would make sense to make it semver major.

kinwa91: pingpingping @JustinBeckwith

jkwlui (approved):

JustinBeckwith:
OOOk, so here's what I decided to do.  There are a few breaking changes brewing for this lib, so I updated the `next` branch, changed the target of this PR, and figured we could track 3.x changes here.

PR #545, opened 2018-11-25
feat: use user credentials if available for adc

JustinBeckwith:
DO NOT MERGE.  This is here to start a conversation around a new approach to default credentials.


JustinBeckwith:
This was a bad idea.

PR #559, opened 2018-12-08
refactor: use execa for install tests, run eslint on samples

JustinBeckwith:
And other misc cleanup.


samples/refreshAccessToken.js
```diff
@@ -1,112 +0,0 @@
-// Copyright 2017, Google, Inc.
```
kinwa91: Why was this removed?

JustinBeckwith: The `refreshAccessToken` method was deprecated in 2.0 :) Users literally never need to call it.

PR #569, opened 2018-12-14
fix(types): export GCPEnv type

ofrobots:
It would be good if users could refer to `GCPEnv` enumeration directly.


ofrobots:
@JustinBeckwith if would be awfully nice if there was a button I could press here to get this into a release!

PR #593, opened 2019-01-15
refactor(deps): use `gaxios` for HTTP requests instead of `axios`

stephenplusplus:
This switches from Axios to Gaxios. Hopefully nobody notices.

### DIFF
--- a/test/test.oauth2.ts
+++ b/test/test.oauth2.ts
@@ -15,9 +15,9 @@
  */
 
 import * as assert from 'assert';
-import {GaxiosError} from 'gaxios';
 import * as crypto from 'crypto';
 import * as fs from 'fs';
+import {GaxiosError} from 'gaxios';
 import * as nock from 'nock';
 import * as path from 'path';
 import * as qs from 'querystring';
@@ -27,6 +27,7 @@ import * as url from 'url';
 import {CodeChallengeMethod, OAuth2Client} from '../src';
 import {LoginTicket} from '../src/auth/loginticket';
 import * as messages from '../src/messages';
+
 const assertRejects = require('assert-rejects');
 
 nock.disableNetConnect();
@@ -642,9 +643,7 @@ it('should pass due to valid issuer', async () => {
 it('should be able to retrieve a list of Google certificates', (done) => {
   const scope = nock('https://www.googleapis.com')
                     .get(certsPath)
-                    .replyWithFile(200, certsResPath, {
-                      'Content-Type': 'application/json',
-                    });
+                    .replyWithFile(200, certsResPath);
   client.getFederatedSignonCerts((err, certs) => {
     assert.strictEqual(err, null);
     assert.strictEqual(Object.keys(certs!).length, 2);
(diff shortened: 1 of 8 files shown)


test/test.oauth2.ts
```diff
@@ -642,7 +642,9 @@ it('should pass due to valid issuer', async () => {
 it('should be able to retrieve a list of Google certificates', (done) => {
   const scope = nock('https://www.googleapis.com')
                     .get(certsPath)
-                    .replyWithFile(200, certsResPath);
+                    .replyWithFile(200, certsResPath, {
+                      'Content-Type': 'application/json',
```
JustinBeckwith: There is a fix here for Gaxios that will fix this issue:
https://github.com/JustinBeckwith/gaxios/pull/21

Need to wait for this land before we release it:
https://github.com/googleapis/gcp-metadata/pull/152

test/test.oauth2.ts
```diff
@@ -999,8 +1002,8 @@ it('getToken should allow a code_verifier to be passed', async () => {
   scope.done();
   assert(res.res);
   if (!res.res) return;
-  const params = qs.parse(res.res.config.data);
-  assert(params.code_verifier === 'its_verified');
+  const params = qs.parse(res.res.config.body);
```
JustinBeckwith: This is very confusing.  What's up here?

test/test.oauth2.ts
```diff
@@ -1031,7 +1034,7 @@ it('getToken should set client_id if not provided in options', async () => {
   scope.done();
   assert(res.res);
   if (!res.res) return;
-  const params = qs.parse(res.res.config.data);
+  const params = qs.parse(res.res.config.body);
```
JustinBeckwith: These should all be data!

stephenplusplus: I'm not sure! `data` wasn't there, but `body` was!

JustinBeckwith: 1.1.1 just shipped - can you try that?

stephenplusplus: It's there now! PR updated.

test/test.oauth2.ts
stephenplusplus: Got rid of it!

test/test.oauth2.ts
```diff
@@ -950,11 +951,12 @@ it('should not retry requests with streaming data', done => {
   const scope = nock('http://example.com').post('/').reply(401);
   client.credentials = {
     access_token: 'initial-access-token',
-    refresh_token: 'refresh-token-placeholder'
+    refresh_token: 'refresh-token-placeholder',
+    expiry_date: (new Date()).getTime() + 500000
```
JustinBeckwith: I'm curious why this change was needed

stephenplusplus: Without this, the test seemed to first try to fetch a new token, which would have required a different nock / more logic.

stephenplusplus:
Tested with several of the client libs as well. BigQuery, Bigtable, Compute, DNS, Logging, Storage to name a fewveral.

JustinBeckwith:
And it worked!?

stephenplusplus:
Yep! And the proxy test script I've been using.

PR #605, opened 2019-01-27
fix: throw if there is no refresh token despite the necessity of refreshing

yuku:
If `OAuth2Client` without refresh token has no access token or has an expired access token, `OAuth2Client.prototype.getAccessToken` should throw an error.

This commit also reduces the number of unreachable code.

- [x] Tests and linter pass


yuku:
I signed it!

PR #618, opened 2019-02-07
fix: use key file when fetching project id

### LINKED ISSUE #614: getProjectId ignores keyFilename
I'm not sure if this is a bug or it is intended, but it looks like if no env vars are set and one only provides a `keyFilename` to the `GoogleAuth` constructor it fails to retrieve a `projectId` unless `getClient` is called first.

```js
const auth = new GoogleAuth({
  keyFilename: '/path/to/my/file.json'
});

let projectId = await auth.getProjectIdAsync(); // will be null

await auth.getClient();
projectId = await auth.getProjectIdAsync(); // will be pulled from cached creds
```

It seems like we'd want to check for the `keyFilename` inside of the [`getFileProjectId`](https://github.com/googleapis/google-auth-library-nodejs/blob/master/src/auth/googleauth.ts#L600) method, but it is pretty unclear to me if this is intentional or not. I'd be happy to put together a PR if this is something we want.

callmehiphop:
Fixes #614 

- [x] Tests and linter pass


src/auth/googleauth.ts
```diff
@@ -603,6 +603,15 @@ export class GoogleAuth {
       return this.cachedCredential.projectId;
     }
 
+    // Ensure the projectId is loaded from the keyFile if available.
+    if (this.keyFilename) {
+      const creds = await this.getClient();
+
+      if (creds && creds.projectId) {
+        return creds.projectId;
```
callmehiphop: @JustinBeckwith made a small amendment to your commit, let me know if you hate it!

PR #627, opened 2019-02-13
fix: avoid loading fast-text-encoding if not in browser environment

### LINKED ISSUE #626: fast-text-encoding pollutes global namespace, breaks other libraries
#### Environment details

  - OS: Linux
  - Node.js version: 8.15.0
  - npm version: 6.7.0
  - `google-auth-library` version: 3.1.0

#### Steps to reproduce

This library relies on `fast-text-encoding@1.0.0`, which dumps `TextEncoder` and `TextDecoder` into the global namespace. Unfortunately, the polyfill implementations provided by `fast-text-encoding` are incomplete (they don't support UTF-8, for example) and this causes other libraries that rely on this functionality to break. For example, libraries built with Emscripten check for `TextDecoder` and will fail if `google-auth-library` is loaded.

Easy fix is to defer requiring `browser/crypto` and `node/crypto` until you know which environment you're in. Since `TextEncoder` and `TextDecoder` aren't needed for `Node`, that seems reasonable.

modeswitch:
Fixes #626

- [x] Tests and linter pass

### DIFF
--- a/src/crypto/browser/crypto.ts
+++ b/src/crypto/browser/crypto.ts
@@ -19,11 +19,12 @@
 
 import * as base64js from 'base64-js';
 import {CryptoSigner} from '../crypto';
+import {isBrowser} from '../../isbrowser';
 
 // Not all browsers support `TextEncoder`. The following `require` will
 // provide a fast UTF8-only replacement for those browsers that don't support
 // text encoding natively.
-if (typeof TextEncoder === 'undefined') {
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
Thanks for the fix!

modeswitch:
@JustinBeckwith I see where the coverage dropped, but I'm not sure how to simulate a browser environment without a bunch of extra work. For one line of code that seems excessive?

PR #642, opened 2019-03-17
feat: support scopes on compute credentials

JustinBeckwith:
The GCE metadata server is adding support for requesting tokens against user accounts with a predefined set of scopes.  This adds experimental support for passing the scopes as querystring parameters to the metadata service in compute credentials.

### DIFF
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -14,9 +14,12 @@
  * limitations under the License.
  */
 
+import * as arrify from 'arrify';
 import {GaxiosError, GaxiosOptions, GaxiosPromise} from 'gaxios';
 import * as gcpMetadata from 'gcp-metadata';
+
 import * as messages from '../messages';
+
 import {CredentialRequest, Credentials} from './credentials';
 import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
 
@@ -26,10 +29,17 @@ export interface ComputeOptions extends RefreshOptions {
    * may have multiple service accounts.
    */
   serviceAccountEmail?: string;
+  /**
+   * The scopes that will be requested when acquiring service account
+   * credentials. Only applicable to modern App Engine and Cloud Function
+   * runtimes as of March 2019.
+   */
+  scopes?: string|string[];
 }
 
 export class Compute extends OAuth2Client {
   private serviceAccountEmail: string;
+  scopes: string[];
 
   /**
    * Google Compute Engine service account credentials.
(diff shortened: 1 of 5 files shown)


src/auth/computeclient.ts
```diff
@@ -43,6 +53,7 @@ export class Compute extends OAuth2Client {
     // refreshed before the first API call is made.
     this.credentials = {expiry_date: 1, refresh_token: 'compute-placeholder'};
     this.serviceAccountEmail = options.serviceAccountEmail || 'default';
+    this.scopes = arrify(options.scopes);
```
mcdonamp: I admit I don't know how gaxios works handles params, so this may be correct. My thought was we needed:

```suggestion
    this.scopes = scopes.join(',');
```

As the metadata server takes `.../token?scopes=https://www.googleapis.com/auth/drive,...`

JustinBeckwith: Yup!  `gaxios` will automatically take collections there and join them in the way you laid out :)

JustinBeckwith:
@bcoe @alexander-fenster this is ready for a look :)

src/auth/computeclient.ts
```diff
@@ -68,7 +79,13 @@ export class Compute extends OAuth2Client {
     const tokenPath = `service-accounts/${this.serviceAccountEmail}/token`;
     let data: CredentialRequest;
     try {
-      data = await gcpMetadata.instance(tokenPath);
+      data = await gcpMetadata.instance({
+        property: tokenPath,
+        params: {
+          scopes: this.scopes
+          // TODO: clean up before submit, fix upstream type bug
```
bcoe: is this still a `TODO` prior to landing?

src/auth/computeclient.ts
```diff
@@ -26,10 +29,17 @@ export interface ComputeOptions extends RefreshOptions {
    * may have multiple service accounts.
    */
   serviceAccountEmail?: string;
+  /**
+   * The scopes that will be requested when acquiring service account
+   * credentials. Only applicable to modern App Engine and Cloud Function
+   * runtimes as of March 2019.
+   */
+  scopes?: string|string[];
```
alexander-fenster: Does anything break if we use just `string[]` here? Are there any usages of `string`?

src/auth/computeclient.ts
mcdonamp: Amazing 💫

src/auth/computeclient.ts
mcdonamp: I expect that users will want to just say `getTokenWithScope('https://...')` rather than having to wrap it `getTokenWithScope(['https://...'])`.

alexander-fenster: Ah, I didn't realize it's an exported interface. OK!

PR #649, opened 2019-03-22
Release google-auth-library v3.1.2

jkwlui:
This pull request was generated using releasetool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@
 
 [1]: https://www.npmjs.com/package/google-auth-library-nodejs?activeTab=versions
 
+## v3.1.2
+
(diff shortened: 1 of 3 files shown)


jkwlui:
@alexander-fenster I've a potentially stupid question about releasing here. Do we update all depending packages (@google-cloud/common, @google-cloud/storage) to use the latest version of `google-auth-library`? Users with `package-lock.json`'s could be potentially locked to a lower version right?

alexander-fenster:
@jkwlui We don't. Since it's a patch release it will be automatically pulled by everyone who uses the library, since `package-lock.json` only fixes versions for those who _develops_ the library, and it is never included in the library release (`npm pack`) which is published.

PR #652, opened 2019-03-26
test: harden tests to handle local logins

JustinBeckwith:
When running unit tests locally, I was getting failing tests if `GOOGLE_APPLICATION_CREDENTIALS` was set in my current session.  The unit test weren't properly faking those out.  This fixes that!

### DIFF
--- a/test/test.googleauth.ts
+++ b/test/test.googleauth.ts
@@ -1104,7 +1104,7 @@ describe('googleauth', () => {
      async () => {
        assert.notEqual(true, auth.isGCE);
        const scope = nockNotGCE();
-       const isGCE = await auth._checkIsGCE();
+       await auth._checkIsGCE();
        assert.strictEqual(false, auth.isGCE);
        scope.done();
      });
@@ -1116,7 +1116,7 @@ describe('googleauth', () => {
        const scope = nockIsGCE();
        await auth._checkIsGCE();
        assert.strictEqual(true, auth.isGCE);
-       const isGCE2 = await auth._checkIsGCE();
+       await auth._checkIsGCE();
        assert.strictEqual(true, auth.isGCE);
        scope.done();
      });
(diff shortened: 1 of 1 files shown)


test/test.googleauth.ts
```diff
@@ -1104,7 +1104,7 @@ describe('googleauth', () => {
      async () => {
        assert.notEqual(true, auth.isGCE);
        const scope = nockNotGCE();
-       const isGCE = await auth._checkIsGCE();
+       await auth._checkIsGCE();
```
bcoe: kind of a weird API method; you call `auth._checkIsGCE()` and as a side effect it populates `auth.isGCE`? I'd potentially instead memoize this, and always `await auth._checkIsGCE()`, but on future calls you'll just get a promise that resolves immediately.

----

this has nothing to do with your work :smile:

test/test.googleauth.ts
```diff
@@ -1134,42 +1134,51 @@ describe('googleauth', () => {
...
-       await assertRejects(auth.getCredentials());
+       const body = await auth.getCredentials();
+       assert.ok(body);
+       assert.strictEqual(
+           body.client_email, 'test-creds@test-creds.iam.gserviceaccount.com');
+       assert.strictEqual(body.private_key, undefined);
        scopes.forEach(s => s.done());
      });
 
+  it('getCredentials should error if metadata server is not reachable', async () => {
```
bcoe: I might move `getCredentials` into a describe block, then the test cases read a bit better:

"it should error if metadata server is not reachable"

test/test.googleauth.ts
```diff
@@ -1233,7 +1242,8 @@ describe('googleauth', () => {
   it('getCredentials should return error when env const is not set',
      async () => {
        // Set up a mock to return a null path string
-       mockEnvVar('GOOGLE_APPLICATION_CREDENTIALS');
+       blockGoogleApplicationCredentialEnvironmentVariable();
```
bcoe: did this helper already exist for other tests?

JustinBeckwith: Yup!

test/test.googleauth.ts
JustinBeckwith: I've been chipping at this thing for a while, but it's still gross 🤣

test/test.googleauth.ts
JustinBeckwith: So personally - I try to avoid nested `describe` blocks, because it screws up indentation and makes things harder to read.  That having been said - lets chat in depth as part of a clean up PR as you mentioned above :)

PR #658, opened 2019-03-29
fix: throw when adc cannot acquire a projectId

JustinBeckwith:
**BREAKING CHANGE** Currently, when a user is using application default credentials, and they call `auth.getProjectId`, we check a variety of sources, and return the first projectId we find.  In the event of none - we return `null`.  This PR changes the behavior, so instead of returning `null`, we throw an exception.


JustinBeckwith:
image

PR #668, opened 2019-04-12
fix: re-throw original exception and preserve message in compute client

### LINKED ISSUE #532: Cloud functions Export/Import - A Forbidden error was returned while attempting to retrieve an access token for the Compute Engine built-in service account.
I am trying to generate a snapshot of my Firestore database, while running the following code inside a Cloud Function, I receive a "Forbidden Error".
From my side, I can't see nothing else in my request that need to be added. 
```
const {auth} = require('google-auth-library');

async function main() {
    const client = await auth.getClient({
            scopes: [
                'https://www.googleapis.com/auth/datastore',
                'https://www.googleapis.com/auth/cloud-platform'
            ]
        }
    );
    const projectId = await auth.getProjectId();
    const url = `https://firestore.googleapis.com/v1beta1/projects/${projectId}/databases/(default):exportDocuments`;
    return client.request({
        url: url,
        method: 'post',
        data: {
            outputUriPrefix: 'gs://my-awesome-storage.appspot.com'
        }
    });
}
```

Could someone please clarify what it actually means?

```
Forbidden error was returned while attempting to retrieve an access token for the Compute Engine built-in service account. This may be because the Compute Engine instance does not have the correct permission scopes specified. The caller does not have permission
```

- [https://firebase.google.com/docs/firestore/reference/rest/v1beta1/projects.databases/exportDocuments#http-request](https://firebase.google.com/docs/firestore/reference/rest/v1beta1/projects.databases/exportDocuments#http-request)

#### Environment details
  - `google-auth-library` version: 2.0.1

JustinBeckwith:
Fixes #532.  The original error message returned from the GCE metadata service was getting swallowed.  This ensures the underlying exception is thrown, and the message is preserved.

### DIFF
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -87,27 +87,20 @@ export class Compute extends OAuth2Client {
     return super.requestAsync<T>(opts, retry).catch(e => {
       const res = (e as GaxiosError).response;
       if (res && res.status) {
-        let helpfulMessage = null;
-        if (res.status === 403) {
-          helpfulMessage =
-              'A Forbidden error was returned while attempting to retrieve an access ' +
-              'token for the Compute Engine built-in service account. This may be because the Compute ' +
-              'Engine instance does not have the correct permission scopes specified.';
-        } else if (res.status === 404) {
-          helpfulMessage =
-              'A Not Found error was returned while attempting to retrieve an access' +
-              'token for the Compute Engine built-in service account. This may be because the Compute ' +
-              'Engine instance does not have any permission scopes specified.';
-        }
-        if (helpfulMessage) {
-          if (e && e.message && !retry) {
-            helpfulMessage += ' ' + e.message;
-          }
-          if (e) {
-            e.message = helpfulMessage;
-          } else {
-            e = new Error(helpfulMessage);
-            (e as NodeJS.ErrnoException).code = res.status.toString();
+        (e as NodeJS.ErrnoException).code = res.status.toString();
+        if (!retry) {
+          if (res.status === 403) {
+            e.message =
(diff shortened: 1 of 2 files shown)


src/auth/computeclient.ts
```diff
@@ -87,27 +87,20 @@ export class Compute extends OAuth2Client {
...
+                'A Forbidden error was returned while attempting to retrieve an access ' +
+                'token for the Compute Engine built-in service account. This may be because the Compute ' +
+                'Engine instance does not have the correct permission scopes specified: ' +
+                e.message;
+          } else if (res.status === 404) {
+            e.message =
+                'A Not Found error was returned while attempting to retrieve an access' +
+                'token for the Compute Engine built-in service account. This may be because the Compute ' +
+                'Engine instance does not have any permission scopes specified: ' +
+                e.message;
```
alexander-fenster: I don't get it :) It does make the code look much better, but it does not add anything to the message: both old and new code append `e.message`, it has not change. In the users' example the message they got was

```
Forbidden error was returned while attempting to retrieve an access token for the Compute Engine built-in service account. This may be because the Compute Engine instance does not have the correct permission scopes specified. The caller does not have permission
```

which clearly has `e.message` appended: `The caller does not have permission`

JustinBeckwith: You ...  are totally right.  One of the commenters on the bug had a great hidden gem:

> It looks like https://github.com/googleapis/google-auth-library-nodejs/blob/master/src/auth/computeclient.ts#L92 is potentially firing on any 403/404 rather than just ones related to getting a token?

We were appending that message for any *request*, not just requests to get the access token.  Helpfully, the unit tests ensured we kept on doing the wrong thing.  

Keep making comments - every time you do it I get to delete more code!

PR #689, opened 2019-05-04
Remove deprecated

alexander-fenster:
We promised to remove these deprecated methods in 3.0.0 and did not. Let's remove them in 4.0.0!

- [x] Tests and linter pass


alexander-fenster:
The diffs are totally wrong, I'll recreate this PR correctly.

PR #693, opened 2019-05-04
test: make tests work on Windows and in GCE

alexander-fenster:
This PR applies the fix from https://github.com/googleapis/synthtool/pull/240 to this repo.

- [x] Tests and linter pass

### DIFF
--- a/.kokoro/test.bat
+++ b/.kokoro/test.bat
@@ -17,7 +17,15 @@
 cd /d %~dp0
 cd ..
 
-call npm install -g npm@latest || goto :error
+@rem The image we're currently running has a broken version of Node.js enabled
+@rem by nvm (v10.15.3), which has no npm bin. This hack uses the functional
+@rem Node v8.9.1 to install npm@latest, it then uses this version of npm to
+@rem install npm for v10.15.3.
+call nvm use v8.9.1 || goto :error
+call node C:\Users\kbuilder\AppData\Roaming\nvm-ps\versions\v8.9.1\node_modules\npm-bootstrap\bin\npm-cli.js i npm -g || goto :error
+call nvm use v10.15.3 || goto :error
+call node C:\Users\kbuilder\AppData\Roaming\nvm-ps\versions\v8.9.1\node_modules\npm\bin\npm-cli.js i npm -g || goto :error
+
 call npm install || goto :error
 call npm run test || goto :error
 
--- a/test/test.googleauth.ts
+++ b/test/test.googleauth.ts
@@ -122,7 +122,7 @@ describe('googleauth', () => {
     sandbox
       .stub(child_process, 'exec')
       .callThrough()
-      .withArgs('gcloud config config-helper --format json', undefined)
+      .withArgs('gcloud config config-helper --format json', sinon.match.func)
       .callsArgWith(1, null, '', null);
 
     const fakeStat = {isFile: () => true} as fs.Stats;
@@ -768,6 +768,7 @@ describe('googleauth', () => {
     const scope = createGetProjectIdNock(STUB_PROJECT);
     const projectId = await auth.getProjectId();
     const stub = (child_process.exec as unknown) as sinon.SinonStub;
+    stub.restore();
     assert(stub.calledOnce);
     assert.strictEqual(projectId, STUB_PROJECT);
     scope.done();
--- a/test/test.googleauth.ts
+++ b/test/test.googleauth.ts
@@ -51,7 +51,6 @@ const BASE_URL = [
 const privateJSON = require('../../test/fixtures/private.json');
 const private2JSON = require('../../test/fixtures/private2.json');
 const refreshJSON = require('../../test/fixtures/refresh.json');
-const fixedProjectId = 'my-awesome-project';
 const privateKey = fs.readFileSync('./test/fixtures/private.pem', 'utf-8');
 const wellKnownPathWindows = path.join(
   'C:',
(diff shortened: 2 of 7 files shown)


alexander-fenster:
Oh wow. 6 Windows tests are failing for real! I'll take a look and fix in a separate PR.

alexander-fenster:
(This one still can be merged, as it allows Windows builds to run on the right version of Node.js)

.kokoro/test.bat
```diff
@@ -17,7 +17,15 @@
 cd /d %~dp0
 cd ..
 
-call npm install -g npm@latest || goto :error
+@rem The image we're currently running has a broken version of Node.js enabled
+@rem by nvm (v10.15.3), which has no npm bin. This hack uses the functional
+@rem Node v8.9.1 to install npm@latest, it then uses this version of npm to
```
bcoe: I will try to work with @jkwlui to get this patched first thing next week; hate that we have to install npm twice.

alexander-fenster:
@bcoe @JustinBeckwith Can you please review the test changes? Looks like everything worked.

test/test.googleauth.ts
```diff
@@ -1221,6 +1264,10 @@ describe('googleauth', () => {
   it('should make the request', async () => {
     const url = 'http://example.com';
     const {auth, scopes} = mockGCE();
+
+    // tslint:disable-next-line no-any
+    sinon.stub(auth as any, 'getDefaultServiceProjectId').resolves();
```
JustinBeckwith: This should not be required.  This is exactly what `createGetProjectIdNock` is supposed to do.

test/test.googleauth.ts
```diff
@@ -402,18 +403,30 @@ describe('googleauth', () => {
...
+      // git does not create symlinks on Windows
+      return;
+    }
     await auth._getApplicationCredentialsFromFilePath(
       './test/fixtures/goodlink'
     );
   });
 
   it('getApplicationCredentialsFromFilePath should error on invalid symlink', async () => {
+    if (onWindows()) {
```
JustinBeckwith: I mean, if windows doesn't create the symlink, it shouldn't matter right?

test/test.googleauth.ts
```diff
@@ -810,6 +823,9 @@ describe('googleauth', () => {
   });
 
   it('getApplicationDefault should cache the credential when using GCE', async () => {
+    // tslint:disable-next-line no-any
+    sinon.stub(auth as any, 'getDefaultServiceProjectId').resolves();
```
JustinBeckwith: This is what `createGetProjectIdNock` is supposed to do.  Something is very wrong if we need this

JustinBeckwith: (here and below)

test/test.googleauth.ts
alexander-fenster: Right! Removed this

test/test.googleauth.ts
alexander-fenster: Changed - please take a look!

test/test.googleauth.ts
alexander-fenster: Changed - please take a look!

alexander-fenster:
It works! @JustinBeckwith can I get one more look here? Thank you!

test/test.googleauth.ts
```diff
@@ -117,6 +118,8 @@ describe('googleauth', () => {
       .withArgs(wellKnownPathWindows)
       .returnsArg(0);
 
+    sandbox.stub(child_process, 'exec').callsArgWith(1, null, '', null);
```
JustinBeckwith: Can we scope this to only mock it with known parameters?  I just want to make sure we're being super explicit about what we want to mock.

```js
sandbox.stub(child_process, 'exec')
  .callThrough()
  .withArgs('gcloud config config-helper --format json')
  .callsArgWith(1, null, '');
```

test/test.googleauth.ts
```diff
@@ -641,12 +654,12 @@ describe('googleauth', () => {
   });
 
   it('getProjectId should return a new projectId the first time and a cached projectId the second time', async () => {
-    mockEnvVar('GCLOUD_PROJECT', fixedProjectId);
+    mockEnvVar('GCLOUD_PROJECT', STUB_PROJECT);
```
JustinBeckwith: Where is `STUB_PROJECT` even defined?  What was wrong with just `fixedProjectId`?

test/util.ts
```diff
@@ -0,0 +1,3 @@
+export function onWindows(): boolean {
```
JustinBeckwith: Can we just add this as an `const isWindows = process.platform === 'win32';` at the top of the file?  No need to have this separate for now, and this file would need a lic header if we did

test/test.googleauth.ts
alexander-fenster: In the same file:

https://github.com/googleapis/google-auth-library-nodejs/blob/47905529d225ca76c020f2400a57fae6150e9536/test/test.googleauth.ts#L42-53

It just happened that we had two variables with same values and same meaning.

test/util.ts
alexander-fenster: Sure! I was not sure if we needed this in more than one file when I did that. Changing.

test/test.googleauth.ts
alexander-fenster: Done! `sinon` has some weird issues with types (it was forcing me to pass at least 2 parameters to `.withArgs` in this case), so I added this `undefined`. By looking at `.d.ts` I couldn't understand what went wrong.

JustinBeckwith:
LGTM after Windows build is passing :)

PR #719, opened 2019-05-28
feat: use X-Goog-Api-Key header

alexander-fenster:
We have this thing called gRPC Fallback, which supports API keys and requires them to be sent in this special header `X-Goog-Api-Key`.  Some details [here](https://cloud.google.com/apis/docs/system-parameters) and more details - with JavaScript code sample I wrote - [here](https://googleapis.github.io/HowToRPC.html).

Let's set this header so that the [JavaScript code](https://github.com/googleapis/googleapis.github.io/blob/5d7c512d44e80f1abecd4e0bafb9ed7eb40168a3/examples/rpc/javascript/src/index.js#L49) could use `getRequestHeaders` in both service account and API key use cases.

- [x] Tests and linter pass

### DIFF
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -756,7 +756,7 @@ export class OAuth2Client extends AuthClient {
     }
 
     if (this.apiKey) {
-      return {headers: {}};
+      return {headers: {'X-Goog-Api-Key': this.apiKey}};
     }
     let r: GetTokenResponse | null = null;
     let tokens: Credentials | null = null;
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -885,9 +885,9 @@ export class OAuth2Client extends AuthClient {
         opts.headers = opts.headers || {};
         opts.headers.Authorization = r.headers.Authorization;
       }
-
       if (this.apiKey) {
-        opts.params = Object.assign(opts.params || {}, {key: this.apiKey});
+        opts.headers = opts.headers || {};
+        opts.headers['X-Goog-Api-Key'] = this.apiKey;
       }
       r2 = await this.transporter.request<T>(opts);
     } catch (e) {
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@types/mocha": "^5.2.1",
     "@types/mv": "^2.1.0",
     "@types/ncp": "^2.0.1",
-    "@types/nock": "^10.0.0",
+    "@types/nock": "^10.0.3",
     "@types/node": "^10.5.1",
     "@types/semver": "^6.0.0",
     "@types/sinon": "^7.0.0",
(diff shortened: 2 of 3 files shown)


src/auth/oauth2client.ts
```diff
@@ -756,7 +756,7 @@ export class OAuth2Client extends AuthClient {
     }
 
     if (this.apiKey) {
-      return {headers: {}};
+      return {headers: {'X-Goog-Api-Key': this.apiKey}};
```
JustinBeckwith: So further down, in the `requestAsync` method, there's a bit of code that grabs the API key and stuffs it in a querystring parameter (I think).  Could I trouble you to modify that code to use the HTTP header instead as well?  Or alternatively, file an issue to come back around for it?

alexander-fenster: So I checked it and verified it works, but I'm still not sure if it's a safe change to remove that `&key=` and replace it with the API token.

But here are my examples that I checked.

Sample JSON request to Vision API:
```
$ cat request.json 
{
  "requests": [
    {
      "image":{
        "source": {
          "imageUri": "https://timedotcom.files.wordpress.com/2019/03/kitten-report.jpg"
        }
      },
      "features":[
        {
          "type":"LABEL_DETECTION",
          "maxResults":1
        }
      ]
    }
  ]
}
```

Passing API key in a GET parameter:
```
$ curl -H 'Content-Type: application/json' -X POST -d '@request.json' https://vision.googleapis.com/v1/images:annotate?key=$GOOGLE_API_KEY 
{
  "responses": [
    {
      "labelAnnotations": [
        {
          "mid": "/m/01yrx",
          "description": "Cat",
          "score": 0.99598557,
          "topicality": 0.99598557
        }
      ]
    }
  ]
}
```

No API key - does not work:
```
$ curl -H 'Content-Type: application/json' -X POST -d '@request.json' https://vision.googleapis.com/v1/images:annotate
{
  "error": {
    "code": 403,
    "message": "The request is missing a valid API key.",
    "status": "PERMISSION_DENIED"
  }
}
```

Passing API key in an HTTP header - it works again:
```
$ curl -H "X-Goog-Api-Key: $GOOGLE_API_KEY" -H 'Content-Type: application/json' -X POST -d '@request.json' https://vision.googleapis.com/v1/images:annotate
{
  "responses": [
    {
      "labelAnnotations": [
        {
          "mid": "/m/01yrx",
          "description": "Cat",
          "score": 0.99598557,
          "topicality": 0.99598557
        }
      ]
    }
  ]
}
```

What's your call @JustinBeckwith?

alexander-fenster: (I made the change, PTAL)

alexander-fenster: I double checked that those two (`&key=` query string parameter and `X-Goog-Api-Key` header) are interchangeable so it should be ok to remove the query string parameter in favod of the header.

Can I have one more look at the added changes @JustinBeckwith?

PR #735, opened 2019-06-15
fix(compute): correctly specify scopes when fetching token

stephenplusplus:
Fixes https://github.com/googleapis/nodejs-storage/issues/748

According to the requirements of the metadata server, this array of values was being parsed incorrectly by gaxios, which uses `querystring.stringify()`. The scopes query string parameter is expected as a comma-separated string value, as opposed to the `?scope=scope1&scopes=scope2` style that `querystring.stringify()` uses: https://cloud.google.com/functions/docs/securing/function-identity#access_tokens

### DIFF
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -80,13 +80,16 @@ export class Compute extends OAuth2Client {
     const tokenPath = `service-accounts/${this.serviceAccountEmail}/token`;
     let data: CredentialRequest;
     try {
-      data = await gcpMetadata.instance({
+      const instanceOptions: gcpMetadata.Options = {
         property: tokenPath,
-        params: {
+      };
+      if (this.scopes.length > 0) {
+        instanceOptions.params = {
           scopes: this.scopes.join(','),
-          // TODO: clean up before submit, fix upstream type bug
-        } as {},
-      });
+        };
+        // TODO: clean up before submit, fix upstream type bug
+      };
+      data = await gcpMetadata.instance(instanceOptions);
     } catch (e) {
       e.message = `Could not refresh access token: ${e.message}`;
       this.wrapError(e);
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -88,7 +88,7 @@ export class Compute extends OAuth2Client {
           scopes: this.scopes.join(','),
         };
         // TODO: clean up before submit, fix upstream type bug
-      };
+      }
       data = await gcpMetadata.instance(instanceOptions);
     } catch (e) {
       e.message = `Could not refresh access token: ${e.message}`;
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -83,7 +83,7 @@ export class Compute extends OAuth2Client {
       data = await gcpMetadata.instance({
         property: tokenPath,
         params: {
-          scopes: this.scopes,
+          scopes: this.scopes.join(','),
           // TODO: clean up before submit, fix upstream type bug
         } as {},
       });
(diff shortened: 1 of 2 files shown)


src/auth/computeclient.ts
```diff
@@ -83,7 +83,7 @@ export class Compute extends OAuth2Client {
       data = await gcpMetadata.instance({
         property: tokenPath,
         params: {
-          scopes: this.scopes,
+          scopes: this.scopes.join(','),
           // TODO: clean up before submit, fix upstream type bug
```
stephenplusplus: Sounds like this fix should just go in gaxios :)

src/auth/computeclient.ts
```diff
@@ -83,7 +83,7 @@ export class Compute extends OAuth2Client {
       data = await gcpMetadata.instance({
         property: tokenPath,
         params: {
-          scopes: this.scopes,
+          scopes: this.scopes.join(','),
```
stephenplusplus: The scopes are expected as a comma-separated string value, as opposed to the `?scope==scope1&scopes=scope2` style that `querystring.stringify()` uses: https://cloud.google.com/functions/docs/securing/function-identity#access_tokens

We could use `querystring.stringify(opts.params, {arrayFormat: 'comma'})`, but I'm not sure that every API endpoint has the same expected format when it comes to query string parameters with multiple values.

src/auth/computeclient.ts
```diff
@@ -80,13 +80,16 @@ export class Compute extends OAuth2Client {
...
-          scopes: this.scopes,
-          // TODO: clean up before submit, fix upstream type bug
-        } as {},
-      });
+      };
+      if (this.scopes.length > 0) {
+        instanceOptions.params = {
+          scopes: this.scopes.join(','),
+        };
+        // TODO: clean up before submit, fix upstream type bug
```
bcoe: do we need this TODO comment? don't quite follow it.

test/test.compute.ts
```diff
@@ -29,7 +29,7 @@ const tokenPath = `${BASE_PATH}/instance/service-accounts/default/token`;
 function mockToken(statusCode = 200, scopes?: string[]) {
   let path = tokenPath;
   if (scopes && scopes.length > 0) {
-    path += '?' + qs.stringify({scopes});
+    path += `?scopes=${encodeURIComponent(scopes.join(','))}`;
```
bcoe: interesting, so the root of the problem is that `storage` required multiple scopes?

src/auth/computeclient.ts
stephenplusplus: Carried over from @JustinBeckwith, who might have intended to fix the problem this PR solves before it shipped, but forgot 🤷‍♂. Interested to see if he has a better fix in mind!

test/test.compute.ts
stephenplusplus: Essentially, yes. Storage does, and I believe many other APIs do as well. The function that I modified is only active in GCE/GCF environments.

src/auth/computeclient.ts
ofrobots: You're change resolves the TODO (modulo the 'upstream bug' part). It can be removed.

stephenplusplus: What upstream type bug, though? Did Justin want to change the way params were parsed upstream? I guess I would like to make sure before removing the comment or merging this PR, as the clarity could nullify either idea.

stephenplusplus:
> Would it be possible to have a system test for this?

@JustinBeckwith are there tests that run on GC* environments?

@ofrobots I think the unit tests are an adequate way to cover this former leak. I just made a change to the unit test to make sure the intended behavior is described and covered better.

src/auth/computeclient.ts
JustinBeckwith: At one point there was a cast here, so that can be removed 🙃

PR #738, opened 2019-06-19
fix: feature detection to check for browser

JustinBeckwith:
Previously, we were using the precense of `window` to figure out if we were running in a browser.  Turns out, it causes a lot of trouble, like this:
https://github.com/googleapis/nodejs-storage/issues/694#issuecomment-503399549

We also see cases where `window` is injected in test environments for things like `jest`.

This uses a feature detection style approach, where we specifically check for subtle crypto availability.


src/transporters.ts
```diff
@@ -60,7 +58,7 @@ export class DefaultTransporter {
    */
   configure(opts: GaxiosOptions = {}): GaxiosOptions {
     opts.headers = opts.headers || {};
-    if (!isBrowser()) {
+    if (typeof window === 'undefined') {
```
bcoe: we're okay to keep this check the same?

alexander-fenster:
@bcoe re: certificate formats, that's all bad bad bad. We basically need to have the same "feature detection check" there as well (not just "oh it's a browser, let's do JWK and not PEM", but "if this environment supports PEM, do PEM, if it supports JWK, do JWK, and there might be other formats to consider"). I think it's OK to leave it as is for now, until someone comes and claims that the existing logic does not work for them.

src/transporters.ts
JustinBeckwith: Aye.  In this case, it doesn't have an impact on which browser feature we're trying to use.  It's just a naive check so we can try to send the right HTTP headers (I think).

PR #748, opened 2019-07-17
fix: use FUNCTION_TARGET to detect GCF 10 and above

### LINKED ISSUE #747: Environment detection does not work on GCF nodejs10
Originally reported by @hdp617 on https://github.com/googleapis/google-auth-library-nodejs/pull/429

> FUNCTION_NAME is no longer set in Google Cloud Functions for Node.js10 and subsequent runtimes. Could we update this?
https://cloud.google.com/functions/docs/env-var#environment_variables_set_automatically
It looks like K_SERVICE is the replacement.

JustinBeckwith:
Fixes #747 

@steren FYI

### DIFF
--- a/src/auth/envDetect.ts
+++ b/src/auth/envDetect.ts
@@ -52,7 +52,7 @@ function isAppEngine() {
 }
 
 function isCloudFunction() {
-  return !!process.env.FUNCTION_NAME;
+  return !!(process.env.FUNCTION_NAME || process.env.FUNCTION_TARGET);
 }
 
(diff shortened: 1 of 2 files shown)


ofrobots:
Would this work correctly on a function running on Cloud Run? See also my comment here: https://github.com/googleapis/google-auth-library-nodejs/issues/747#issuecomment-512607161

JustinBeckwith:
Chatted with @steren offline - this is fine as a tactical fix, but we do need to drive a better effort forward that uses the metadata server.

PR #749, opened 2019-07-19
refactor!: getOptions() no longer accepts GoogleAuthOptions

### LINKED ISSUE #390: GoogleAuth ignores credentials passed to it in getClient in favor of cached credentials
The [getClient](https://github.com/google/google-auth-library-nodejs/blob/master/src/auth/googleauth.ts#L656) method in GoogleAuth will ignore any credentials passed to it if it has cached credentials, regardless of differences between the credentials passed in the parameter and the credentials that have been cached:

https://github.com/google/google-auth-library-nodejs/blob/0ff24c7c4a1c5da604dd2019dfc225a922934e3e/src/auth/googleauth.ts#L656-L675

As far as I can find, there is no way to actually clear the cached credentials once it has been set.  (Aside from manually explicitly setting the property to null, which doesn't feel right to me).

I'd imagine ideally that if the credentials didn't match, the cached ones would be ignored.
I'd expect at least some way to bust the cache; possibly as a property in the options parameter (something like `useCachedCredentials: boolean`).

bcoe:
BREAKING CHANGE: `getOptions()` no longer accepts GoogleAuthOptions, preventing a variety of side effects caused by credentials that have already been stored in cache. `new GoogleAuth({})` should instead be used to customize options, or to instantiate multiple clients. `fromJSON()` has also been made idempotent (it no longer changes the cache). 

fixes #390

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -441,7 +441,6 @@ export class GoogleAuth {
         'Must pass in a JSON object containing the Google auth settings.'
       );
     }
-    this.jsonContent = json;
     options = options || {};
     if (json.type === 'authorized_user') {
       client = new UserRefreshClient(options);
@@ -451,6 +450,18 @@ export class GoogleAuth {
     }
     client.fromJSON(json);
     return client;
+  }
+
+  /**
+   * Create a credentials instance using the given input options.
+   * @param json The input object.
+   * @param options The JWT or UserRefresh options for the client
+   * @returns JWT or UserRefresh Client with data
+   */
+  private _cacheJWTInput(json: JWTInput, options?: RefreshOptions): JWT | UserRefreshClient {
+    let client: UserRefreshClient | JWT;
+    this.jsonContent = json;
+    return this.fromJSON(json, options);
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/auth/googleauth.ts
```diff
@@ -682,27 +708,14 @@ export class GoogleAuth {
    * Automatically obtain a client based on the provided configuration.  If no
    * options were passed, use Application Default Credentials.
    */
-  async getClient(options?: GoogleAuthOptions) {
```
alexander-fenster: I don't have a strong opinion about this, but I'm a little bit afraid that this kind of breaking change is breaking in a bad way: it won't actually make the old code (the code that uses `auth.getClient(opts)`) report any errors, but its behavior will be changed.

Do we want to deprecate the options differently by doing this:

```ts
interface DeprecatedOptions {}

async getClient(options?: DeprecatedOptions) {
  if (options) {
    throw new Error('Passing options to getClient is forbidden in v5.0.0. Use new GoogleAuth(opts) instead.');
  }
  ...
}
```

That way we'll deliberately break the code for users that rely on passing options to `getClient` and they will fix the code during the upgrade and not after some long debugging later.

The `DeprecatedOptions` can then be removed completely in v6.0.0.

PR #750, opened 2019-07-19
docs: use `open` instead of `opn` in the README

tashian:
The node module `opn` has been renamed to `open`, so I've changed the example to reflect that.


tashian:
I signed it!

PR #771, opened 2019-08-08
feat: allow upstream clients to opt out of User-Agent

bcoe:
~BREAKING CHANGE: User-Agent now defaults to google-auth-library; empty User-Agent now possible.~

I've been talked out of a breaking change.

### DIFF
--- a/src/transporters.ts
+++ b/src/transporters.ts
@@ -25,7 +25,7 @@ import {validate} from './options';
 
 // tslint:disable-next-line no-var-requires
 const pkg = require('../../package.json');
-const PRODUCT_NAME = 'google-auth-library';
+const PRODUCT_NAME = 'google-auth-library-nodejs';
 
 export interface Transporter {
   request<T>(opts: GaxiosOptions): GaxiosPromise<T>;
--- a/src/transporters.ts
+++ b/src/transporters.ts
@@ -25,7 +25,7 @@ import {validate} from './options';
 
 // tslint:disable-next-line no-var-requires
 const pkg = require('../../package.json');
-const PRODUCT_NAME = 'google-api-nodejs-client';
+const PRODUCT_NAME = 'google-auth-library';
 
 export interface Transporter {
   request<T>(opts: GaxiosOptions): GaxiosPromise<T>;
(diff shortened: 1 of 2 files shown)


alexander-fenster:
Why do you think it is a breaking change?

bcoe:
@alexander-fenster perhaps I'm being overly cautious, my concern is mainly that we're changing the default User-Agent on people, which could break folks that rely on this field, e.g., proxies, people writing tests with `nock`.

alexander-fenster:
I agree that it's a small change in behavior and folks using `nock` or similar things AND checking the headers might break. I highly doubt it still deserves a semver major :) 

Another way to look at it is that the users normally test their own code, e.g. if they set `User-Agent` they might want to see that it's actually sent out using `nock`. This use case still works. If they don't care about `User-Agent` they should not validate it in the test.

src/transporters.ts
```diff
@@ -57,18 +57,19 @@ export class DefaultTransporter {
...
     if (typeof window === 'undefined') {
       // set transporter user agent if not in browser
-      const uaValue: string = opts.headers['User-Agent'];
-      if (!uaValue) {
-        opts.headers['User-Agent'] = DefaultTransporter.USER_AGENT;
+      const uaValue: string = headers['User-Agent'];
+      if (!headers.hasOwnProperty('User-Agent')) {
+        headers['User-Agent'] = DefaultTransporter.USER_AGENT;
+      } else if (!uaValue) {
+        delete headers['User-Agent'];
```
JustinBeckwith: Who do we need to delete the property in this case?

src/transporters.ts
```diff
@@ -57,18 +57,19 @@ export class DefaultTransporter {
    * @return Configured options.
    */
   configure(opts: GaxiosOptions = {}): GaxiosOptions {
-    opts.headers = opts.headers || {};
+    const headers = Object.assign({}, opts.headers || {});
```
JustinBeckwith: For my learning - what's the value of doing the clone here as opposed to just modifying the headers in place?  As far as I can tell, we just assign it back to the options object at the end anyway.

src/transporters.ts
bcoe: I think it's cleaner, I'd rather have the `User-Agent` explicitly not set, rather than it set as an empty string, or `null`, or `undefined`.

src/transporters.ts
bcoe: It's a bit defensive, but if someone had written code like this:

```
const headers = {'User-Agent': 'my-app'};
const options = {
  headers
}
auth.configure(options);
console.info(headers);
```

without the copy, we will delete the key on their original headers object, not just `options.headers`, because options.headers is populated as a reference.

src/transporters.ts
alexander-fenster: So... what happens when we delete the `User-Agent` header? Will the underlying transporter not send anything, or will it send some default value?

alexander-fenster: Yeah, that's what I figured: gaxios will send this by default (with no `User-Agent` set):

```
User-Agent: node-fetch/1.0 (+https://github.com/bitinn/node-fetch)
```

Is that what we want?

bcoe:
@dazuma indicated that `User-Agent` field should be ignored, based on a conversation, I think what we're going to instead do is introduce an additional field which is auth library version, and not worry about mucking with user-agent.

PR #779, opened 2019-08-29
feat(impersonated): add impersonated credentials auth

### LINKED ISSUE #535: Implement ImpersonatedCredentials
Allows one user or service account to impersonate another using `iamcredentials api`

its already implemented or pending for several other languages:

* python: https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/impersonated_credentials.py#L17

* java: https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java

* golang:  https://go-review.googlesource.com/c/oauth2/+/143957

--

here's what i've done so far:

- [impersonatedclient.ts](https://gist.github.com/salrashid123/2d4b4143e3bdb0d3934a961068ba6626)
- [usage](https://gist.github.com/salrashid123/15ba1b10ac55192234d9533e100209a2)

salrashid123:
Fixes https://github.com/googleapis/google-auth-library-nodejs/issues/535

Adds `ImpersonatedCredentials`.  This credential type basically takes one source credential  and exchanges it for another, different service account.

for reference, it exists currently in:

- `google-auth-python`: [google.auth.impersonated_credentials.html](https://google-auth.readthedocs.io/en/latest/reference/google.auth.impersonated_credentials.html)

- `google-auth-java`: [ImpersonatedCredentials.java](https://googleapis.dev/java/google-auth-library/latest/com/google/auth/oauth2/ImpersonatedCredentials.html)

---

I do NOT have testcases that cover this (any pointers or sample on how to do the exchange (i.,e two round trips for the token would be appreciated)

---

- Usage:

```typescript



const {Logging} = require('@google-cloud/logging');
const { GoogleAuth, JWT, Impersonated, ImpersonatedOptions  } = require('google-auth-library');
const { Storage } = require('@google-cloud/storage');
const {Logging} = require('@google-cloud/logging');

var svcAccountFile = 'svc_account.json';
var project_id = 'yourprojectID';


const keys = require(svcAccountFile);

// create a sourceCredential
let saclient = new JWT(
  keys.client_email,
  null,
  keys.private_key,
  ['https://www.googleapis.com/auth/cloud-platform', 'https://www.googleapis.com/auth/iam'],
);


// Use that to impersonate the targetPrincipal
  sourceClient: saclient,
  targetPrincipal: "target@anotherproject.iam.gserviceaccount.com",
  lifetime: 30,
  delegates: [],
  targetScopes: ["https://www.googleapis.com/auth/cloud-platform"]
});

// At this point you can use the client as any other:

// 1. Acquire Headers
const authHeaders = await targetClient.getRequestHeaders();
console.log(authHeaders);

// 2. Use client in authorized session 
targetClient.getAccessToken().then(res => {
  let url = 'https://www.googleapis.com/storage/v1/b?project=' + project_id
  targetClient.requestAsync({ url }).then(resp => {
    console.log(resp.data.items[0]);
  }).catch(function (error) {
    console.error('Unable to list buckets: ' + error);
  });
});

// 3. Use client with GCP Service 
const { Logging } = require('@google-cloud/logging');

const log = logging.log('mylog');
const text = 'Hello, world!';
const metadata = {
  resource: {type: 'global'},
};
const entry = log.entry(metadata, text);
await log.write(entry);
console.log(`Logged: ${text}`);

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@
 
 [1]: https://www.npmjs.com/package/google-auth-library-nodejs?activeTab=versions
 
+## [7.3.0](https://www.github.com/googleapis/google-auth-library-nodejs/compare/v7.2.0...v7.3.0) (2021-07-02)
+
+
+### Features
+
+* add useJWTAccessAlways and defaultServicePath variable ([#1204](https://www.github.com/googleapis/google-auth-library-nodejs/issues/1204)) ([79e100e](https://www.github.com/googleapis/google-auth-library-nodejs/commit/79e100e9ddc64f34e34d0e91c8188f1818e33a1c))
+
 ## [7.2.0](https://www.github.com/googleapis/google-auth-library-nodejs/compare/v7.1.2...v7.2.0) (2021-06-30)
 
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "google-auth-library",
-  "version": "7.2.0",
+  "version": "7.3.0",
   "author": "Google Inc.",
   "description": "Google APIs Authentication Client Library for Node.js",
   "engines": {
--- a/samples/package.json
+++ b/samples/package.json
@@ -14,7 +14,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@googleapis/iam": "^0.2.0",
-    "google-auth-library": "^7.2.0",
+    "google-auth-library": "^7.3.0",
     "node-fetch": "^2.3.0",
     "opn": "^5.3.0",
     "server-destroy": "^1.0.1"


bcoe:
@salrashid123 @JustinBeckwith I will make an effort to provide code review on this tomorrow 👍

src/auth/impersonated.ts
```diff
@@ -0,0 +1,213 @@
...
+   *  final access_token. If set, the sequence of identities must have
+   * "Service Account Token Creator" capability granted to the preceding
+   * identity. For example, if set to [serviceAccountB, serviceAccountC],
+   * the sourceCredential must have the Token Creator role on serviceAccountB.
+   * serviceAccountB must have the Token Creator on serviceAccountC.
+   * Finally, C must have Token Creator on target_principal.
+   * If left unset, sourceCredential must have that role on targetPrincipal.
+   * @param targetScopes scopes to request during the authorization grant.
+   * @param lifetime number of seconds the delegated credential should be
+   * valid for (up to 3600).
```
bcoe: If I need to hold the actual credentials to generate the impersonated client, how does this provide improved security. I think I just don't fully understand the threat model this protects me from (not be a security person).

src/auth/impersonated.ts
```diff
@@ -0,0 +1,213 @@
...
+  ): Promise<GetTokenResponse> {
+    const iat = Math.floor(new Date().getTime() / 1000);
+
+    if (this.credentials.expiry_date) {
+      if (this.credentials.expiry_date <= iat) {
+        const token = await this.sourceClient
+          .getAccessToken()
+          .then(res => {
+            const name = 'projects/-/serviceAccounts/' + this.targetPrincipal;
+            const u = `https://iamcredentials.googleapis.com/v1/${name}:generateAccessToken`;
```
bcoe: it might be worth pulling this endpoint into a variable, and perhaps even making it configurable with `ImpersonatedOptions`.

test/test.impersonated.ts
```diff
@@ -0,0 +1,82 @@
...
+afterEach(() => {
+  nock.cleanAll();
+  sandbox.restore();
+});
+
+it('should refresh if access token has expired', async () => {
+  const scopes = [mockExample()];
+  impersonated.credentials.access_token = 'initial-access-token';
+  impersonated.credentials.expiry_date = new Date().getTime() - 10000;
+  await impersonated.request({url});
```
bcoe: is the idea that it will immediately refresh the token on the first request, because the impersonated account hasn't yet been created?

bcoe (changes requested):
this is looking really solid, mainly a few initial questions.

src/auth/impersonated.ts
salrashid123: You'll always need some credentials to bootstrap with but this capability basically allows you to impersonate _another_ credential

you can do something like this:  You own serivceAccoutnA, I own  serviceAccountB and I give B permissions on GCS BucketG that i own; A does not have permission on G.

I can create an [IAM condition](https://cloud.google.com/iam/docs/conditions-overview) that says "only allow A access to impersonate B between 1am and 2am".  So what serviceAccountA has is temp access to bucketG.  
You can basically continue this chain of delegation.  The full flow is described here

- https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials

test/test.impersonated.ts
salrashid123: yeah, i derived that idea from how `google-auth-python` does its refresh of creds.  Should i init creds on creation (i was just blindly copying the existing compute implementation here)

src/auth/impersonated.ts
salrashid123: ok added in an option

bcoe (approved):

bcoe (changes requested):
I'm going to switch some of the promise logic to use `await`, but otherwise this is honestly looking very solid; I believe we'll want to add documentation, which I will create a tracking ticket for.

bcoe:
this is blocked while we have a few internal discussions about how to make auth more extensible.

bcoe:
@bshaffer is this functionality partially addressed by your work around [id tokens](https://github.com/googleapis/google-auth-library-nodejs/pull/867).

salrashid123:
hi-
They're two different things though:  @bshaffer 's work is about generating and using a serviceAccount's OpenIDConnect token (OIDC)...the PR here is just about impersonating a service account (i.,e you impersonate serviceAcountA and get its `access_token` (not OIDC token)

bshaffer:
To echo @salrashid123, as far as I know this work is not related to the ID token work I've done.

rh389:
Is there anything we can do to help move this PR along? I'm not sure what's still outstanding.

(IIUC this functionality would be very useful in local development, to allow an individual developer's local credentials to impersonate a service account - some GCP APIs only support access by service accounts - without the developer needing a private SA key file on their machine)

bcoe:
@rh389 I apologize for the slow reply, I would love to get this work moving again. I'm going to bring it up with a few folks internally and see what we can make happen.

Mistic92:
Hi, we'd use this feature too :) For example for vision api we need service account from exact project as adding from other one is not working. Fortunatelly impersonation is working and we wanted to use it but without this feature we are blocked

bcoe:
@salrashid123 not quite sure what happened, had a merge conflict updating your PR, changes incoming.

PR #783, opened 2019-09-05
fix(deps): nock@next has types that work with our libraries

bcoe:
moves to version of nock with correct types.


callmehiphop:
LGTM assuming tests pass

PR #787, opened 2019-09-17
fix(deps): update to gcp-metadata and address envDetect performance issues

bcoe:
This pulls in an updated version of `gcp-metadata` which checks both IP and HOST; failing fast as soon as one of the endpoints fails to respond (this addresses issues with cloud run, vs., local environment).

Since only `gcpMetadata.isAvailable()` is currently fast pathed, this also performs a slight refactor on environment detection, we first detect `isComputeEngine()` hitting `isAvailable`, if we detect that we are in a compute engine environment, we next check if we're in kubernetes.

This results in one additional check, but improves cold-start performance in non-GCP environments drastically.

TODO: confirm with @JustinBeckwith, @stephenplusplus, etc., that `gcpMetadata.isAvailable()` is true for kubernetes.

### DIFF
--- a/src/auth/envDetect.ts
+++ b/src/auth/envDetect.ts
@@ -36,10 +36,12 @@ export async function getEnv() {
       env = GCPEnv.APP_ENGINE;
     } else if (isCloudFunction()) {
       env = GCPEnv.CLOUD_FUNCTIONS;
-    } else if (await isKubernetesEngine()) {
-      env = GCPEnv.KUBERNETES_ENGINE;
     } else if (await isComputeEngine()) {
-      env = GCPEnv.COMPUTE_ENGINE;
+      if (await isKubernetesEngine()) {
+        env = GCPEnv.KUBERNETES_ENGINE;
+      } else {
+        env = GCPEnv.COMPUTE_ENGINE;
+      }
     } else {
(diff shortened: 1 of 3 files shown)


src/auth/envDetect.ts
```diff
@@ -36,10 +36,12 @@ export async function getEnv() {
       env = GCPEnv.APP_ENGINE;
     } else if (isCloudFunction()) {
       env = GCPEnv.CLOUD_FUNCTIONS;
-    } else if (await isKubernetesEngine()) {
-      env = GCPEnv.KUBERNETES_ENGINE;
     } else if (await isComputeEngine()) {
-      env = GCPEnv.COMPUTE_ENGINE;
+      if (await isKubernetesEngine()) {
```
bcoe: we end up performing one more check for kubernetes, but we use the fast path `gcpMetadata.isAvailable` rather than the slow `gcpMetadata.instance('attributes/cluster-name')` for the initial check, correcting issues with `@nodejs/logging` performance.

bcoe:
related to:

https://github.com/googleapis/gcp-metadata/issues/238
https://github.com/googleapis/nodejs-logging-winston/issues/389
https://github.com/googleapis/nodejs-logging-bunyan/issues/354

PR #790, opened 2019-09-18
feat: adds forceRefreshOnFailure flag, for refreshing token on error

skeggse:
It's highly unlikely, but the `expiry_date` could be wrong, or the comparison in `isTokenExpiring` could fail due to an inaccurate system clock. As a further fallback behavior, some developers may wish to opt-in to attempting a refresh if Google's servers respond with a status code that may indicate the token has expired.

- [x] Tests and linter pass

### DIFF
--- a/src/auth/jwtclient.ts
+++ b/src/auth/jwtclient.ts
@@ -83,7 +83,10 @@ export class JWT extends OAuth2Client {
       optionsOrEmail && typeof optionsOrEmail === 'object'
         ? optionsOrEmail
         : {email: optionsOrEmail, keyFile, key, keyId, scopes, subject};
-    super({eagerRefreshThresholdMillis: opts.eagerRefreshThresholdMillis});
+    super({
+      eagerRefreshThresholdMillis: opts.eagerRefreshThresholdMillis,
+      forceRefreshOnFailure: opts.forceRefreshOnFailure,
+    });
     this.email = opts.email;
     this.keyFile = opts.keyFile;
     this.key = opts.key;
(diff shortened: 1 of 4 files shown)


bcoe:
:wave: @skeggse thanks for the contribution :+1: I have it on my plate this week to dig into why the heck refresh token logic sometimes fails, I'll get back to reviewing this once I've done a bit of a deep dive on the functionality.

bcoe:
@skeggse I addressed a major issue with our refresh logic here:

https://github.com/googleapis/google-auth-library-nodejs/pull/794

Do you think that this functionality is still something folks might need? If so I might just call the feature `alwaysRefresh`, and we could use it to set `forceReferesh` to `true` when getting a token?

skeggse:
I think this is still useful, and would like to see a solution similar to this one that doesn't depend on local state (like the system clock).

bcoe:
@skeggse how would the locking be facilitated in this case, you'd only have one of your clients configured to refresh, and several with no eager refresh or `alwaysLazyRefresh`?

skeggse:
@bcoe we extend the [`OAuth2Client`](https://github.com/googleapis/google-auth-library-nodejs/blob/08d038af73589aed460c01acd21f8f58fef01f5d/src/auth/oauth2client.ts#L359) to override the [`refreshTokenNoCache`](https://github.com/googleapis/google-auth-library-nodejs/blob/08d038af73589aed460c01acd21f8f58fef01f5d/src/auth/oauth2client.ts#L600) method and wrap the functionality in a pessimistic lock. Any additional logic to attempt refreshes in other contexts simply folds into that extended functionality.

bcoe:
@skeggse, I'm open to this feature, but I don't quite understand the flag `lazyRefresh`, I might call it something like `forceRefreshOnFailure`?

skeggse:
Sounds good to me! The flag just means that we should refresh the token in lazy cases, but your flag is far clearer.

bcoe:
@skeggse I want to use this later today as an example of how our release process works, it's not actually blocked by anything on your end :+1:

skeggse:
@bcoe awesome, love it!

PR #803, opened 2019-10-08
feat: GoogleAuth implements AuthClient

proppy:
See googleapis/nodejs-googleapis-common#175.


bcoe:
@proppy I will need to dig into this request before I can give any opinion ... my main concern is that auth is already fairly complex, and I don't love the idea of exporting more of an API surface.

Would this change also require updating types, so that you can then pass the auth instance in to `google-apis-nodejs-client`?

proppy:
@bcoe let's discuss this in https://github.com/googleapis/nodejs-googleapis-common/issues/175 as I already started describing the usecases I'm after in the issue comments.

src/auth/googleauth.ts
```diff
@@ -96,9 +97,7 @@ export interface GoogleAuthOptions {
 export const CLOUD_SDK_CLIENT_ID =
   '764086051850-6qr4p6gpi6hn506pt8ejuq83di341hur.apps.googleusercontent.com';
 
-export class GoogleAuth {
-  transporter?: Transporter;
-
+export class GoogleAuth extends AuthClient {
```
broady: this adds setCredentials, inherited by AuthClient, which wouldn't be used.

I think its implementation would need to be
```
this.getClient().setCredentials(...)
```

bcoe:
It might be good to try to get @JustinBeckwith's feedback on this as well, for historical context. I think of `GoogleAuth` more as a factory function, which I believe is why it itself is not an instance of Auth; I'm a little concerned this could create more confusion in the codebase.

bcoe:
@proppy I'm doing a little bit of cleanup, bother you to address @broady's feedback? and to add a test?

my tendency is to be a bit cautious about turning GoogleAuth, which has so far been an abstraction for fetching instances of other credential types, into an AuthClient itself ... but given that it meets the contract, perhaps this is okay. A couple tests would be great though, perhaps one that specifically addresses Chris' feedback.

proppy:
Sorry for letting that slip, would you be happier with the union type solution discussed in https://github.com/googleapis/nodejs-googleapis-common/issues/175 (`OAuth2Client | GoogleAuth`) instead of modifying the public interface?

bcoe:
@proppy I think that sounds like a lower touch solution to me, and I'd be happy with it. Want to make that PR, you can just ping me in chat and I'll land.

JustinBeckwith:
I think we got this covered in a different way in another PR :) Closing this out!

PR #829, opened 2019-11-20
feat: set x-goog-user-project header, with quota_project from default credentials

bcoe:
If the `quota_project` parameter is set, kicks it along in the request header `x-goog-user-project`.

### DIFF
--- a/src/auth/credentials.ts
+++ b/src/auth/credentials.ts
@@ -39,6 +39,7 @@ export interface JWTInput {
   client_id?: string;
   client_secret?: string;
   refresh_token?: string;
+  quota_project?: string;
 }
 
 export interface CredentialBody {
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -749,10 +749,15 @@ export class GoogleAuth {
     // quota_project, stored in application_default_credentials.json, is set in
     // the x-goog-user-project header, to indicate an alternate account for
     // billing and quota:
-    if (this.jsonContent && this.jsonContent.quota_project) {
-      headers = Object.assign({}, headers, {
-        'x-goog-user-project': this.jsonContent.quota_project,
-      });
+    if (this.jsonContent?.quota_project) {
+      // If x-goog-user-project has explicitly been set in the client headers,
+      // it takes precedence.
+      headers = Object.assign(
+        {
+          'x-goog-user-project': this.jsonContent.quota_project,
+        },
+        headers
+      );
     }
     return headers;
   }
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -745,7 +745,16 @@ export class GoogleAuth {
    */
   async getRequestHeaders(url?: string) {
     const client = await this.getClient();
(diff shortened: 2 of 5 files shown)


src/auth/credentials.ts
```diff
@@ -39,6 +39,7 @@ export interface JWTInput {
   client_id?: string;
   client_secret?: string;
   refresh_token?: string;
+  quota_project?: string;
```
broady: eeeee, this looks correct but I think "JWTInput" is the wrong name. It looks more like "JSONCredential"

client_id, client_secret, refresh_token have nothing to do with JWTs.

src/auth/googleauth.ts
```diff
@@ -745,7 +745,16 @@ export class GoogleAuth {
    */
   async getRequestHeaders(url?: string) {
     const client = await this.getClient();
-    return client.getRequestHeaders(url);
+    let headers = client.getRequestHeaders(url);
+    // quota_project, stored in application_default_credentials.json, is set in
+    // the x-goog-user-project header, to indicate an alternate account for
+    // billing and quota:
+    if (this.jsonContent && this.jsonContent.quota_project) {
```
broady: is minification a problem ever? i.e. `["quota_project"]`

src/auth/googleauth.ts
```diff
@@ -745,7 +745,16 @@ export class GoogleAuth {
...
+      headers = Object.assign({}, headers, {
+        'x-goog-user-project': this.jsonContent.quota_project,
+      });
```
broady: A comment about why it's not the other way around, i.e.
`Object.assign({}, ..., headers)`

src/auth/googleauth.ts
broady: oh, no, I guess because jsonContent is a JWTInput

src/auth/googleauth.ts
alexander-fenster: If `getRequestHeaders` always returns an object, the whole `Object.assign` can be removed and replaced with a simple assignment :)

src/auth/credentials.ts
bcoe: I agree this is an oddly named interface, having not spent a huge amount of time in this codebase, I find the naming a bit difficult in general :smile: ...

I think it's partially just that auth is a complex topic, with all the approaches we support.

tldr; as we put more resources on authentication, might be worth overhauling this library in the future.

src/auth/googleauth.ts
bcoe: I tend to agree that `headers` should take precedence if set explicitly, I've updated the PR and added a comment.

bcoe: I've noticed a few bugs in our auth library in my time working on it, caused by passing an object by reference and than modifying the object, e.g., deleting a key.

I've been being defensive by creating shallow copies with `Object.assign({})`, something worth thinking about in API design would be starting to `Object.freeze` objects that we don't wish for users to modify, at which point it forces people to make copies, but you can pass the object around if you do not need to mutate it.

PR #831, opened 2019-11-21
fix: expand on x-goog-user-project to handle auth.getClient()

bcoe:
## What is this?

This expands on the initial implementation of `x-goog-user-project` header population, such that headers are appropriately populated when `auth.getClient().getRequestHeaders()` is called, the initial approach only supported `auth.getRequestHeaders()`.

Thanks to @JustinBeckwith for pointing out this code-path which would have bitten us.

## Implementation Details

This implementation is closer to @broady's initial recommendation of having the logic in `UserRefreshClient` rather than my suggestion of `GoogleAuth`, however, I've moved the logic for header population into `OAuth2Client`, this approach seemed elegant because:

* `OAuth2Client` already implements `getRequestHeaders` (and several clients, like `UserRefreshClient` rely on its implementation.
* It puts us in a position to support `quota_project` across clients.
* It seemed the most elegant of a few approaches I tried.


broady:
LGTM

PR #857, opened 2019-12-23
feat: cache results of getEnv()

bcoe:
This fixes failing tests caused by introducing caching to [gcp-metadata](https://github.com/googleapis/gcp-metadata/pull/274).

Taking a page fro `gcp-metadata`, this now also caches the promise returned by `getEnv` rather than its value, such that multiple libraries instantiated at the same time will not result in multiple enqueued promises.

### DIFF
--- a/src/auth/envDetect.ts
+++ b/src/auth/envDetect.ts
@@ -37,7 +37,7 @@ export async function getEnv() {
 }
 
 async function getEnvMemoized(): Promise<GCPEnv> {
-  let env: GCPEnv | undefined;
+  let env = GCPEnv.NONE;
   if (isAppEngine()) {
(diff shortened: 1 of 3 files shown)


src/auth/envDetect.ts
```diff
@@ -22,27 +22,34 @@ export enum GCPEnv {
...
-      }
+  if (envPromise) {
+    return envPromise;
+  }
+  envPromise = getEnvMemoized();
+  return envPromise;
+}
+
+async function getEnvMemoized(): Promise<GCPEnv> {
+  let env: GCPEnv | undefined;
```
alexander-fenster: Maybe just `let env = GCPEnv.NONE;` and remove the last `else`?

PR #859, opened 2019-12-24
feat: id token clients

bshaffer:
This is an RFC for fetching ID tokens using JWT and Compute Metadata.

### DIFF
--- a/samples/idtokens.js
+++ b/samples/idtokens.js
@@ -1,0 +1,31 @@
+// Copyright 2019 Google LLC
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+/**
+ * Instead of specifying the type of client you'd like to use (JWT, OAuth2, etc)
+ * this library will automatically choose the right client based on the environment.
+ */
+const {GoogleAuth} = require('google-auth-library');
+
+async function main() {
+  const targetAudience = 'iap-client-id';
+  const iapUrl = 'https://some.iap.url';
+  const auth = new GoogleAuth();
+  const client = await auth.getIdTokenClient(targetAudience);
+  const res = await client.request({iapUrl});
+  console.log(res.data);
+}
+
+main().catch(console.error);
--- a/samples/idtokens.js
+++ b/samples/idtokens.js
@@ -21,10 +21,10 @@ const {GoogleAuth} = require('google-auth-library');
 
 async function main() {
   const targetAudience = 'iap-client-id';
-  const iapUrl = 'https://some.iap.url';
+  const url = 'https://some.iap.url';
   const auth = new GoogleAuth();
   const client = await auth.getIdTokenClient(targetAudience);
-  const res = await client.request({iapUrl});
+  const res = await client.request({url});
   console.log(res.data);
 }
 
--- a/src/auth/jwtclient.ts
+++ b/src/auth/jwtclient.ts
@@ -151,6 +151,21 @@ export class JWT extends OAuth2Client {
   }
 
   /**
+   * @param targetAudience the audience for the fetched ID token.
+   * @return URL to consent page.
+   */
+  getIdTokenClient(targetAudience: string): OAuth2Client {
+    return new JWT({
+      additionalClaims: {target_audience: targetAudience},
+      iss: this.email,
+      sub: this.subject,
+      scope: this.scopes,
+      keyFile: this.keyFile,
+      key: this.key,
+    } as JWTOptions);
+  }
+
+  /**
    * Indicates whether the credential requires scopes to be created by calling
    * createScoped before use.
    * @deprecated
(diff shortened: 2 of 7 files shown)


samples/idtokens.js
```diff
@@ -0,0 +1,31 @@
...
+'use strict';
+
+/**
+ * Instead of specifying the type of client you'd like to use (JWT, OAuth2, etc)
+ * this library will automatically choose the right client based on the environment.
+ */
+const {GoogleAuth} = require('google-auth-library');
+
+async function main() {
+  const targetAudience = 'iap-client-id';
```
bcoe: Should we also consider adding a region tag to the sample? seems like it could be valuable to reference samples in documentation we create on `cloud.google.com` for this feature?

src/auth/jwtclient.ts
```diff
@@ -150,6 +150,21 @@ export class JWT extends OAuth2Client {
     }
   }
 
+  /**
+   * @param targetAudience the audience for the fetched ID token.
+   * @return URL to consent page.
+   */
+  getIdTokenClient(targetAudience: string): OAuth2Client {
+    return new JWT({
```
bcoe: would we need to implement `fetchIdToken` in `jwtclient` as well?

samples/idtokens.js
```diff
@@ -0,0 +1,31 @@
...
+  const url = 'https://some.iap.url';
+  const auth = new GoogleAuth();
+  const client = await auth.getIdTokenClient(targetAudience);
+  const res = await client.request({url});
```
bcoe: Oh, I think I understand, the idea with the `getIdTokenClient` is you now have a client instantiated that can make the authenticated request with this new type of token? Feels a bit weird to me still.

bshaffer: The idea is that it's explicit when you're using ID tokens as opposed to access tokens. It's not a necessary requirement, we could have the client be inferred like this:

```js
  const auth = new GoogleAuth({targetAudience});
  const res = await client.request({url});
```

src/auth/jwtclient.ts
bshaffer: We could! this is already happening on line 123:

```js
if (
  this.additionalClaims &&
  (this.additionalClaims as {
    target_audience: string;
  }).target_audience
) {
  const {tokens} = await this.refreshToken();
  return {
    headers: this.addSharedMetadataHeaders({
      Authorization: `Bearer ${tokens.id_token}`,
    }),
  };
}
```
I could peel that part out, but since it's basically the same I didn't see much of a point.

test/test.googleauth.ts
```diff
@@ -1523,6 +1523,38 @@ describe('googleauth', () => {
...
+  it('should call getClient for getIdTokenClient', async () => {
+    // Set up a mock to return path to a valid credentials file.
+    mockEnvVar(
+      'GOOGLE_APPLICATION_CREDENTIALS',
+      './test/fixtures/private.json'
+    );
+
+    const spy = sinon.spy(auth, 'getClient');
+    const client = await auth.getIdTokenClient('a-target-audience');
+    assert(client instanceof JWT);
```
bcoe: it might be good to use `nock` to further test that the `Authorization` header was set.

Would the target audience tend to be something like `foo@appspot.gserviceaccount.com`, i.e., members in IAM? If so might be worth making the example `a-target-audience@example.com` (just so folks reading tests get this distinction).

src/auth/jwtclient.ts
bcoe: Oh interesting, I just noticed that `fetchIdToken` is private on ComputeClient, ignore me. If I'm following correctly:

1. a user instantiates their auth client with `targetAudience`, or for `JWT`, with `{additionalClaims: target_audience}`.
1. requests made will have additional headers set, signed with keys corresponding to public keys.
1. an upstream service can use `verify` to ensure that the request being made was from the appropriate audience, based on these same public keys?

---

I think for consistency we should probably make `JWT` accept `targetAudience` in the constructor, and just have it be an alias for `{target_audience: targetAudience}`, maybe? so that the constructor is the same as the `Compute` instance?

bshaffer:
Other language implementations:

 ### Python

- [`IDTokenCredentials`](https://github.com/googleapis/google-auth-library-python/blob/master/docs/user-guide.rst#identity-tokens): The method `IDTokenCredentials` is called on the `compute_engine` or `service_account` modules, as opposed to calling `Credentials`, which is done when using access tokens:

```python
from google.oauth2 import service_account

target_audience = 'https://example.com'

creds = service_account.IDTokenCredentials.from_service_account_file(
        '/path/to/svc.json',
        target_audience=target_audience)

# make authenticated request 
authed_session = AuthorizedSession(creds)
resp = authed_session.get(url)
```

### Java

- [`IdTokenCredentials`](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/IdTokenCredentials.java#L51): The `IdTokenCredentials` class extends a base `OAuth2Credentials` class, and requires an `IdTokenProvider` interface object, which is implemented by the `ServiceAccountCredentials` and `ComputeEngineCredentials` classes.

```java
GoogleCredentials adcCreds = GoogleCredentials.getApplicationDefault();
IdTokenCredentials tokenCredential = IdTokenCredentials.newBuilder()
    .setIdTokenProvider(adcCreds)
    .setTargetAudience(targetAudience).build();

// Make the HTTP request
GenericUrl genericUrl = new GenericUrl("https://example.com");
HttpCredentialsAdapter adapter = new HttpCredentialsAdapter(tokenCredential);
HttpTransport transport = new NetHttpTransport();
HttpRequest request = transport.createRequestFactory(adapter).buildGetRequest(genericUrl);
HttpResponse response = request.execute();
```

PR #870, opened 2020-01-14
feat: export LoginTicket and TokenPayload

### LINKED ISSUE #869: LoginTicket / TokenPayload types are not exported
I am trying to create a function that returns the `LoginTicket` or `TokenPayload`. These types are not exported. I see that `LoginTicket` is a class and maybe there is a reason to keep that private. If I have to settle for `TokenPayload`, that will satisfy my use case.

Example:

```ts
import { OAuth2Client, LoginTicket } from 'google-auth-library';

async function verify(idToken: string): LoginTicket {
  const client = new OAuth2Client({
    clientId: CLIENT_ID,
  });

  const ticket = await client.verifyIdToken({
    idToken,
    audience: CLIENT_ID,
  });

  return ticket;
}
```

Will you accept a patch with one or both of these types exported?

hjr3:

Fixes #869


bcoe:
@hjr3 thank you for the contribution :tada: it's very much appreciated.

PR #891, opened 2020-02-14
docs(samples): update ID token samples to match others

bshaffer:
- Updates region tags in `idtokens-iap.js` to match other language samples
 - Adds "sample-metadata" and moves region tags in `verifyIdToken-iap.js` to match other node samples
 - Updates README with now-valid link to samples


samples/verifyIdToken-iap.js
```diff
@@ -49,14 +58,13 @@ async function main(
 
   // Print out the info contained in the IAP ID token
   console.log(ticket);
-
```
bcoe: since top level `await` is not yet supported in Node.js, we should pull the await bits into an `async` helper function, this is the documented format:

```
function main(arg1 = 'default_value_one', arg2 = 1234) {
  // [START your_region_tag]
  /**
   * TODO(developer): Uncomment these variables before running the sample.
   */
  // const arg1 = 'default_value_one';
  // const arg2 = 1234;

  // Imports the Google Cloud Some API library
  const {SomeApiClient} = require('@google-cloud/some-api');

  // Instantiates a client
  const someApiClient = new SomeApiClient();

  async function doSomething() {
    ...
  }

  doSomething();
  // [END your_region_tag]
}

main(...process.argv.slice(2));
```

bshaffer: Ok I think I got it...

PR #905, opened 2020-02-29
fix: do not warn for SDK creds

### LINKED ISSUE #833: Don't warn when using Cloud SDK creds if quota_project_id is set
Here: https://github.com/googleapis/google-auth-library-nodejs/blob/3646b7f9deb296aaff602dd2168ce93f014ce840/src/auth/googleauth.ts#L427:13

JustinBeckwith:
We used to warn users when relying on a token provided by `gcloud auth application-default login`.  We are headed down a road where this is the default behavior (again), and probably don't need to warn users anymore.  Especially since the docs don't ask you to do it anymore.  

Fixes #833 somewhat bluntly

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -379,7 +379,6 @@ export class GoogleAuth {
       location,
       options
     );
-    this.warnOnProblematicCredentials(client as JWT);
     return client;
   }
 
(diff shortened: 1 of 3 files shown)


broady:
LGTM, server sends a good message upon failure.

bshaffer:
Oof, I guess we need to fix this in [PHP auth](https://github.com/googleapis/google-auth-library-php/blob/master/src/Credentials/UserRefreshCredentials.php#L88) as well..

PR #908, opened 2020-03-01
fix: use iamcredentials API to sign blobs

### LINKED ISSUE #866: Update signBlob API
From @frankyn on an issue about cold start performance: https://github.com/googleapis/nodejs-storage/issues/951#issuecomment-566416370

> The auth library is using an [older version of SignBlob](https://github.com/googleapis/google-auth-library-nodejs/blob/master/src/auth/googleauth.ts#L811) and should be using the [newer version of SignBlob](https://cloud.google.com/iam/credentials/reference/rest/v1/projects.serviceAccounts/signBlob). A similar issue was hit in [Java](https://github.com/googleapis/google-cloud-java/issues/4557).

JustinBeckwith:
Fixes #866

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -824,13 +824,16 @@ export class GoogleAuth {
     const id = `projects/${projectId}/serviceAccounts/${creds.client_email}`;
     const res = await this.request<SignBlobResponse>({
       method: 'POST',
-      url: `https://iam.googleapis.com/v1/${id}:signBlob`,
-      data: {bytesToSign: crypto.encodeBase64StringUtf8(data)},
+      url: `https://iamcredentials.googleapis.com/v1/{name=${id}}`,
+      data: {
+        payload: crypto.encodeBase64StringUtf8(data),
+      },
     });
-    return res.data.signature;
+    return res.data.signedBlob;
   }
 }
 
 export interface SignBlobResponse {
-  signature: string;
+  keyId: string;
+  signedBlob: string;
 }
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -809,10 +809,10 @@ export class GoogleAuth {
       throw new Error('Cannot sign data without `client_email`.');
     }
 
-    const id = `projects/${projectId}/serviceAccounts/${creds.client_email}`;
+    const url = `https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/${creds.client_email}:signBlob`;
     const res = await this.request<SignBlobResponse>({
       method: 'POST',
-      url: `https://iamcredentials.googleapis.com/v1/{name=${id}}`,
+      url,
       data: {
         payload: crypto.encodeBase64StringUtf8(data),
       },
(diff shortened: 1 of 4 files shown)


test/test.googleauth.ts
```diff
@@ -1390,21 +1390,21 @@ describe('googleauth', () => {
       const {auth, scopes} = mockGCE();
       mockEnvVar('GCLOUD_PROJECT', STUB_PROJECT);
       const email = 'google@auth.library';
-      const iamUri = `https://iam.googleapis.com`;
-      const iamPath = `/v1/projects/${STUB_PROJECT}/serviceAccounts/${email}:signBlob`;
-      const signature = 'erutangis';
+      const iamUri = `https://iamcredentials.googleapis.com`;
+      const iamPath = `/v1/%7Bname=projects/${STUB_PROJECT}/serviceAccounts/${email}%7D`;
```
frankyn: In Java we used the following pattern, (without  `{name=}`)

https://github.com/frankyn/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L76

```java
  static final String SIGN_BLOB_URL_FORMAT = "https://iam.googleapis.com/v1/projects/-/serviceAccounts/%s:signBlob?alt=json";
```

JustinBeckwith: Happy to go change it. Just so we're doing this with eyes wide open, a few things:
- I chose this pattern based on the docs (here)[https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signBlob#http-request]. Is the doc correct?  
- I did write a sample, and made sure with an end to end test that the endpoint works.  So what's there now is effective (again happy to change)

frankyn: I followed the same document in Java and I think I mistook `{name=...}` as being an expression. Additionally, I find it easier to parse when reading it. 

For `STUB_PROJECT`, given the project id is in the service account, stating it explicitly seems unnecessary for me.

src/auth/googleauth.ts
```diff
@@ -812,13 +812,16 @@ export class GoogleAuth {
     const id = `projects/${projectId}/serviceAccounts/${creds.client_email}`;
     const res = await this.request<SignBlobResponse>({
       method: 'POST',
-      url: `https://iam.googleapis.com/v1/${id}:signBlob`,
-      data: {bytesToSign: crypto.encodeBase64StringUtf8(data)},
+      url: `https://iamcredentials.googleapis.com/v1/{name=${id}}`,
```
bcoe: If I understand correctly, we're moving from the `iam` API to sign, rather than the `iamcredentials` API.

Are we sure these API surfaces are identical, or should be consider making this a breaking change? CC: @frankyn

frankyn: iamcredentials is the v2 of the iam API. I synced with the IAM team on which API libraries should use. iamcredentials works in the same way for this specific scenario and isn't impacted by the requests having to be routed back to US region (which is the case with the existing iam API).

bshaffer:
I agree with @bcoe about not understanding the implications of the change WRT backwards compatibility. Looks like `SignBlobResponse` changes, but it's only used internally so that *might* be OK. Are there other differences between making calls to the new Endpoint, e.g. different error responses or credential requirements?

JustinBeckwith:
@bshaffer really relying on @frankyn here :)

frankyn:
@bshaffer I sent an email cc'd you on it and raised the question to the SWE PoC that I spoke with last on this specific question.

PR #984, opened 2020-06-29
Deduce projectId from credentials

### LINKED ISSUE #983: Project Id detection fails when manually providing "credentials" option
#### Environment details

  - OS: Darwin Kernel Version 19.5.0
  - Node.js version: v14.4.0
  - npm version: 6.14.4
  - `google-auth-library` version: 6.0.1

#### Steps to reproduce

```js
import { Datastore } from '@google-cloud/datastore'

const datastore = new Datastore({
  credentials: JSON.parse(process.env.GOOGLE_APPLICATION_KEY) // content of the JSON file
})

datastore.get(datastore.key(["kind", "name"])).catch(err => {
  console.error(err) // Unable to detect a Project Id in the current environment.
})
```

Current workaround:
```js
const datastore = new Datastore({
  keyFilename: 'foo', // Any truthy value here
  credentials: { ... },
})
```

matthieusieben:

Fixes #983 🦕

## Details

When instantiating `GoogleAuth` with `options.credentials = { ... }`, there is no need for `options.keyFilename` to be set to be able to determine the `projectId`.


src/auth/googleauth.ts
```diff
@@ -594,7 +594,7 @@ export class GoogleAuth {
     }
 
     // Ensure the projectId is loaded from the keyFile if available.
-    if (this.keyFilename) {
+    if (this.keyFilename || this.jsonContent) {
```
JustinBeckwith, line 597: So I'm a little concerned here on a few fronts.  

The method is described as `getFileProjectId`, and I'm worried about concerning the scope of the method to also attempt reading from jsonContent outside the context of a file.  

The other thing I'm wondering - perhaps more importantly - will `this.jsonContent` have a `projectId` defined?  The code below will only return a `projectId` if it gets defined on the underlying `UserRefreshClient` or `JWT` objects.  If those objects are instantiated via `getClient`, it won't have a `projectId` (or at least I don't think it will).  

I don't think this is going to do what you want it to do :)

JustinBeckwith (changes requested):
Thank you for the commit!  I kind of suspect this approach won't work, but it's possible I'm missing something.  I think I'd need to see this passing in a system test.  Just to double check this does what we want - could I trouble you for a unit test?  Thank you so much for giving this a shot!

src/auth/googleauth.ts
matthieusieben: The thing is, this method and getClient don't have the same method to determine if data can be read from jsonContent.

You don't get the same result if you call this method after calling getClient once before... That is the only think this PR tries to fix

matthieusieben: But maybe we should indeed just remove the if() all together since we check for the return value of getClient anyway

bcoe:
@matthieusieben I think I'd like to take a step back and start with some specific use cases/unit tests.

If there are some situations where we could return the `projectId`, and are not, it would be good to support them. But I would want tests that demonstrate the end to end flow:

1. providing configuration to the auth client.
2. plucking the projecId out from that auth client and returning it.

Would you be open to closing this PR, and opening and starting with an issue? that outlines some of the workflows where we think we could return a `projectId` and are not?

matthieusieben:
There is already an issue opened [here](https://github.com/googleapis/google-auth-library-nodejs/issues/983).

I do the following:

```js
const datastore = new Datastore({ credentials: JSON.parse(process.env.GOOGLE_APPLICATION_KEY) })
```

matthieusieben:
This PR is most certainly not the best way to fix this issue. IMHO this class needs a refactor to consistently set itself up, as calling `getFileProjectId()` **after** `getClient()` does not yield a result consistent with calling `getFileProjectId()` **before** `getClient()`.

This PR only fixes that inconsistency. Nothing more nothing less.

bcoe:
> calling getFileProjectId() after getClient() does not yield a result consistent with calling getFileProjectId() before getClient().

@matthieusieben I hear you loud and clear. What if we add a unit test for this specific case. This will give us a little bit of a testing harness in place when we move forward with a refactor.

If you don't have time to write a test, its completely understood, but we'll need to add one before landing this.

sofisl:
I will be closing this bug for now until we can clarify the auth strategy a bit better. @matthieusieben, thank you for submitting and sparking this discussion!

PR #990, opened 2020-07-07
chore: address linting issues

JustinBeckwith:


sofisl (approved):

src/auth/jwtaccess.ts
```diff
@@ -150,7 +149,7 @@ export class JWTAccess {
     callback?: (err?: Error) => void
   ): void | Promise<void> {
     if (callback) {
-      this.fromStreamAsync(inputStream).then(r => callback(), callback);
+      this.fromStreamAsync(inputStream).then(() => callback(), callback);
```
stephenplusplus, line 152: Just to be cool, would ".then(callback, callback)" work, too?

src/auth/jwtclient.ts
```diff
@@ -300,7 +300,7 @@ export class JWT extends OAuth2Client implements IdTokenProvider {
     callback?: (err?: Error | null) => void
   ): void | Promise<void> {
     if (callback) {
-      this.fromStreamAsync(inputStream).then(r => callback(), callback);
+      this.fromStreamAsync(inputStream).then(() => callback(), callback);
```
stephenplusplus, line 303: Callback comment to my (callback, callback) comment.

src/auth/refreshclient.ts
```diff
@@ -135,7 +131,7 @@ export class UserRefreshClient extends OAuth2Client {
     callback?: (err?: Error) => void
   ): void | Promise<void> {
     if (callback) {
-      this.fromStreamAsync(inputStream).then(r => callback(), callback);
+      this.fromStreamAsync(inputStream).then(() => callback(), callback);
```
stephenplusplus, line 134: (📲◀️, 📲◀️)

stephenplusplus (approved):
Gotta love a good ol' lint PR.

src/auth/jwtaccess.ts
JustinBeckwith: Theoretically yes.  For readability, I like to expand this out for folks.

stephenplusplus: Probably the trickier thing here (for non-Promise users and non-JS users [who would also be non-Promise users]), is the "then(successHandler, errorHandler)" convention. If we're not all-in on the shortcut by using `callback` consistently, I think we'd be most readable these ways:

- `.then(callback).catch(callback)`
- `.then(() => callback()).catch(err => callback(err))`
- `.then(() => callback(), err => callback(err))`
- `.then(callback, callback)`

JustinBeckwith: That is entirely fair 😆 I have some weird historical aversion to just passing the function reference, because it kicked me in the face once.  Chances are I'm being weird, and you're right that I should just pass the function reference (example #4).

PR #991, opened 2020-07-07
fix: migrate token info API to not pass token in query string

bojeil-google:
Google APIs will stop accepting requests that pass OAuth tokens on the query string from June 1, 2021.
To align with security best practices, we should not pass the token in the query string when calling tokeninfo endpoint.
This also follows the gcloud samples code: https://cloud.google.com/sdk/gcloud/reference/auth/application-default/print-access-token?hl=en

`curl -H "Content-Type: application/x-www-form-urlencoded" -d "access_token=$(gcloud auth application-default print-access-token)" https://www.googleapis.com/oauth2/v1/tokeninfo`


src/auth/oauth2client.ts
```diff
@@ -459,7 +459,7 @@ export class OAuth2Client extends AuthClient {
   }
 
   protected static readonly GOOGLE_TOKEN_INFO_URL =
-    'https://oauth2.googleapis.com/tokeninfo';
```
JustinBeckwith, line 462: At some point, I was asked to move *to* this endpoint.  Could I trouble you to start a thread with me, Chris Broadfoot, and Hong on this specific point internally?

src/auth/oauth2client.ts
```diff
@@ -1015,9 +1015,12 @@ export class OAuth2Client extends AuthClient {
...
   async getTokenInfo(accessToken: string): Promise<TokenInfo> {
     const {data} = await this.transporter.request<TokenInfoRequest>({
-      method: 'GET',
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded',
+      },
       url: OAuth2Client.GOOGLE_TOKEN_INFO_URL,
-      params: {access_token: accessToken},
+      data: querystring.stringify({access_token: accessToken}),
```
JustinBeckwith, line 1023: You don't need to stringify here, `gaxios` does that for you :)  Just passing the object should suffice.

JustinBeckwith (commented):
Thanks for doing this!  I want to double check on the correct token Url internally, and make sure we have a public doc somewhere that has the "true" answer 😆

src/auth/oauth2client.ts
bojeil-google: It seems to stringify it as JSON based on the nock expectations:

```
"headers": {
    "content-type": [
      "application/json"
    ],
    "user-agent": [
      "google-api-nodejs-client/6.0.3"
    ],
    "x-goog-api-client": [
      "gl-node/10.15.3 auth/6.0.3"
    ],
    "accept": [
      "application/json"
    ],
    "content-length": [
      "22"
    ],
    "accept-encoding": [
      "gzip,deflate"
    ],
    "connection": [
      "close"
    ]
  },
  "body": "{\"access_token\":\"abc\"}"
}
```
But it does not form url-encode it to `"access_token=abc"`

A same technique seems to be used here to for form url-encoding:
https://github.com/googleapis/google-auth-library-nodejs/blob/b82db030935204c85ec7acc77542c21152aec074/src/auth/oauth2client.ts#L620

src/auth/oauth2client.ts
bojeil-google: Yeah, sure thing. I think I was involved in that earlier thread. That is how I got familiar with this issue.

bojeil-google: After offline discussion, the old endpoint now supports POST requests. I have reverted to the old endpoint.

JustinBeckwith (approved):

bcoe (approved):
👍 I think we should be careful about the rollout, but this looks good to me.

PR #999, opened 2020-07-12
feat: Added option for processing POST requests

ghost:
Option for POST request added alongwith some readability improvements.


JustinBeckwith:
Greetings @Accelerator-One and thank you for the pull request!  We really do appreciate the contribution, but I don't think these are a set of changes we want to accept.  These samples do a very particular set of things (specific to a GET call) so they are consistent across 7 different languages, and used in context on cloud.google.com.  I think we're going to avoid adding a specific POST example for now.  

Thanks again!

PR #1022, opened 2020-08-03
feat: defines internal utils for OAuth client Auth and error handling

bojeil-google:
This defines internal utilities to handle:
- OAuth client authentication
- Standard OAuth response parsing into native Javascript errors.

### DIFF
--- a/src/auth/oauth2common.ts
+++ b/src/auth/oauth2common.ts
@@ -1,0 +1,212 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {GaxiosOptions} from 'gaxios';
+import * as querystring from 'querystring';
+
+import {Crypto, createCrypto} from '../crypto/crypto';
+
+/** List of HTTP methods that accept request bodies. */
+const METHODS_SUPPORTING_REQUEST_BODY = ['PUT', 'POST', 'PATCH'];
+
+/**
+ * OAuth error codes.
+ * https://tools.ietf.org/html/rfc6749#section-5.2
+ */
+export type OAuthErrorCode =
+  | 'invalid_request'
+  | 'invalid_client'
+  | 'invalid_grant'
+  | 'unauthorized_client'
+  | 'unsupported_grant_type'
+  | 'invalid_scope'
+  | string;
+
+/**
+ * The standard OAuth error response.
+ * https://tools.ietf.org/html/rfc6749#section-5.2
+ */
+export interface OAuthErrorResponse {
+  error: OAuthErrorCode;
+  error_description?: string;
+  error_uri?: string;
+}
+
+/**
+ * OAuth client authentication types.
+ * https://tools.ietf.org/html/rfc6749#section-2.3
+ */
+export type ConfidentialClientType = 'basic' | 'request-body';
+
+/**
+ * Defines the client authentication credentials for basic and request-body
+ * credentials.
+ * https://tools.ietf.org/html/rfc6749#section-2.3.1
+ */
+export interface ClientAuthentication {
+  confidentialClientType: ConfidentialClientType;
+  clientId: string;
+  clientSecret?: string;
+}
+
+/**
+ * Abstract class for handling client authentication in OAuth-based
+ * operations.
+ * When request-body client authentication is used, only application/json and
+ * application/x-www-form-urlencoded content types for HTTP methods that support
+ * request bodies are supported.
+ */
+export abstract class OAuthClientAuthHandler {
+  private crypto: Crypto;
+
+  /**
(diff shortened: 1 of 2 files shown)


src/auth/oauth2common.ts
```diff
@@ -0,0 +1,212 @@
...
+    bearerToken?: string
+  ) {
+    // Bearer token prioritized higher than basic Auth.
+    if (bearerToken) {
+      opts.headers = opts.headers || {};
+      Object.assign(opts.headers, {
+        Authorization: `Bearer ${bearerToken}}`,
+      });
+    } else if (
+      typeof this.clientAuthentication !== 'undefined' &&
```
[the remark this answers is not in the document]
bcoe: I would just use optional chaining here:

`else if (this?. clientAuthentication?. confidentialClientType === 'basic) {` **wipes hands**.

src/auth/oauth2common.ts
```diff
@@ -0,0 +1,212 @@
...
+  /**
+   * Applies client authentication on the request's body if request-body
+   * client authentication is selected.
+   *
+   * @param opts The GaxiosOptions whose headers or data are to be modified
+   *   depending on the client authentication mechanism to be used.
+   */
+  private injectAuthenticatedRequestBody(opts: GaxiosOptions) {
+    if (
+      typeof this.clientAuthentication !== 'undefined' &&
```
bcoe, line 145: I think we could  use the optional chaining where ever we use this pattern.

test/test.oauth2common.ts
```diff
@@ -0,0 +1,415 @@
...
+    };
+    const actualOptions = Object.assign({}, originalOptions);
+    const expectedOptions = Object.assign({}, originalOptions);
+    (expectedOptions.headers as Headers).Authorization = `Bearer ${bearerToken}`;
+
+    handler.testApplyClientAuthenticationOptions(actualOptions, bearerToken);
+    assert.deepStrictEqual(expectedOptions, actualOptions);
+  });
+
+  it('should prioritize bearer token over basic auth', () => {
```
bcoe, line 334: really happy with the tests that have been added.

src/auth/oauth2common.ts
```diff
@@ -0,0 +1,212 @@
...
+import {Crypto, createCrypto} from '../crypto/crypto';
+
+/** List of HTTP methods that accept request bodies. */
+const METHODS_SUPPORTING_REQUEST_BODY = ['PUT', 'POST', 'PATCH'];
+
+/**
+ * OAuth error codes.
+ * https://tools.ietf.org/html/rfc6749#section-5.2
+ */
+export type OAuthErrorCode =
```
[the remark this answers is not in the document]
bcoe: My guess is they're exported for the benefit  of code reuse within the library itself?

bcoe (commented):
Left a few initial notes and have some more time slated to dig deeper tomorrow.

In general this structure is looking great.

src/auth/oauth2common.ts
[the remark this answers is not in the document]
bojeil-google: In general, the exported types and interfaces here will be re-used within the library but not externally as @bcoe mentioned below.
This type in particular, does not need to be exported. I have removed the export here. Thanks for catching.

src/auth/oauth2common.ts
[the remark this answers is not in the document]
bojeil-google: Good point. I ended up using optional chaining as @bcoe suggested. Turned out to be much cleaner :)

src/auth/oauth2common.ts
bojeil-google: Done. I agree this is much cleaner and simpler.

test/test.oauth2common.ts
bojeil-google: 👍
Code coverage for this file:
```
File                                   | % Stmts | % Branch | % Funcs | % Lines |
  oauth2common.ts                      |     100 |    90.63 |     100 |     100 |
```

test/test.oauth2common.ts
```diff
@@ -0,0 +1,415 @@
...
+    it('should inject creds in non-empty x-www-form-urlencoded content', () => {
+      const handler = new TestOAuthClientAuthHandler(reqBodyAuth);
+      const originalOptions: GaxiosOptions = {
+        url: 'https://www.example.com/path/to/api',
+        method: 'POST',
+        headers: {
+          // Handling of headers should be case insensitive.
+          'content-Type': 'application/x-www-form-urlencoded',
+        },
+        data: 'key1=value1&key2=value2',
```
bcoe, line 258: nit, might consider using:

```
const querystring = require('querystring')
querystring.encode({key1: 'value, k2: 'value2'})
```

Not a requirement.

test/test.oauth2common.ts
```diff
@@ -0,0 +1,415 @@
...
+      };
+      const actualOptions = Object.assign({}, originalOptions);
+      const expectedOptions = Object.assign({}, originalOptions);
+      const credentials =
+        `client_id=${encodeURIComponent(reqBodyAuth.clientId)}` +
```
bcoe, line 263: likewise, might be worth considering using querystring if it makes the code look cleaner, I'm open to template strings if it doesn't seem to look much cleaner.

bcoe (approved):
This  isa really solid contribution @bojeil-google thank you  👍

test/test.oauth2common.ts
bojeil-google: I agree. This is much cleaner.
Done.

test/test.oauth2common.ts
bojeil-google: Done here and elsewhere.

PR #1026, opened 2020-08-07
feat: implements the OAuth token exchange spec based on rfc8693

bojeil-google:
Implements an internal utility for exchanging OAuth tokens using the rfc/8693 spec.

### DIFF
--- a/src/auth/stscredentials.ts
+++ b/src/auth/stscredentials.ts
@@ -35,15 +35,59 @@ import {
  * https://tools.ietf.org/html/rfc8693#section-2.1
  */
 export interface StsCredentialsOptions {
+  /**
+   * REQUIRED. The value "urn:ietf:params:oauth:grant-type:token-exchange"
+   * indicates that a token exchange is being performed.
+   */
   grantType: string;
+  /**
+   * OPTIONAL. A URI that indicates the target service or resource where the
+   * client intends to use the requested security token.
+   */
   resource?: string;
+  /**
+   * OPTIONAL. The logical name of the target service where the client
+   * intends to use the requested security token.  This serves a purpose
+   * similar to the "resource" parameter but with the client providing a
+   * logical name for the target service.
+   */
   audience?: string;
+  /**
+   * OPTIONAL. A list of space-delimited, case-sensitive strings, as defined
+   * in Section 3.3 of [RFC6749], that allow the client to specify the desired
+   * scope of the requested security token in the context of the service or
+   * resource where the token will be used.
+   */
   scope?: string[];
+  /**
+   * OPTIONAL. An identifier, as described in Section 3 of [RFC8693], eg.
+   * "urn:ietf:params:oauth:token-type:access_token" for the type of the
+   * requested security token.
+   */
   requestedTokenType?: string;
+  /**
+   * REQUIRED. A security token that represents the identity of the party on
+   * behalf of whom the request is being made.
+   */
   subjectToken: string;
+  /**
+   * REQUIRED. An identifier, as described in Section 3 of [RFC8693], that
+   * indicates the type of the security token in the "subject_token" parameter.
+   */
   subjectTokenType: string;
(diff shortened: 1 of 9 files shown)


src/auth/stscredentials.ts
```diff
@@ -0,0 +1,179 @@
...
+ * The interface does not directly map to the spec and instead is converted
+ * to be compliant with the JavaScript style guide. This is because this is
+ * instantiated internally.
+ * StsCredentials implement the OAuth 2.0 token exchange based on
+ * https://tools.ietf.org/html/rfc8693.
+ * Request options are defined in
+ * https://tools.ietf.org/html/rfc8693#section-2.1
+ */
+export interface StsCredentialsOptions {
+  grantType: string;
```
JustinBeckwith, line 38: I don't know if the spec lays these out clearly, but I like to pull whatever text they have (if of appropriate value and length) to include in comments above the interface properties.  Like:

```js
/**
 * REQUIRED.  The value "urn:ietf:params:oauth:grant-type:token-exchange" indicates that a token exchange is being performed.
*/
grantType: string;

File: src/auth/stscredentials.ts [hunk 2]
```diff
@@ -0,0 +1,179 @@
...
+  audience?: string;
+  scope?: string;
+  requested_token_type?: string;
+  subject_token: string;
+  subject_token_type: string;
+  actor_token?: string;
+  actor_token_type?: string;
+  client_id?: string;
+  client_secret?: string;
+  [key: string]: string | undefined;
```
JustinBeckwith, line 67: That is less than ideal :/  It effectively means any string key is value here with a string value.  Is this really an open property bag?

src/auth/stscredentials.ts
```diff
@@ -0,0 +1,179 @@
...
+    };
+    // Apply OAuth client authentication.
+    this.applyClientAuthenticationOptions(opts);
+
+    try {
+      const response = await this.transporter.request<StsSuccessfulResponse>(
+        opts
+      );
+      // Successful response.
+      const stsSuccessfulResponse = response.data;
```
JustinBeckwith, line 165: I'm curious - why return the `response` object along with the data?

src/auth/stscredentials.ts
```diff
@@ -0,0 +1,179 @@
...
+      stsSuccessfulResponse.res = response;
+      return stsSuccessfulResponse;
+    } catch (error) {
+      // Translate error to OAuthError.
+      if (error.response) {
+        throw getErrorFromOAuthErrorResponse(
```
JustinBeckwith, line 171: Just to double check - does this method create a `new Error`, or modify the message on the existing?  I want to be about not creating a new error, because we want to preserve the original stack trace.

JustinBeckwith (approved):
LGTM with a few nits

src/auth/stscredentials.ts
bojeil-google: Done

src/auth/stscredentials.ts
bojeil-google: Ok. Currently there is only one non-standard field here. I manually added it.

src/auth/stscredentials.ts
bojeil-google: Mostly following the existing pattern established [here](https://github.com/bojeil-google/google-auth-library-nodejs/blob/a7e5701a8394d79fe93d28794467747a23cf9ff4/src/auth/oauth2client.ts#L309) and elsewhere in that file.

src/auth/stscredentials.ts
bojeil-google: It was creating a new error and not preserving the original error data. I have extended this to preserve the original error data (including stack) but to modify the error message.

bcoe:
@bojeil-google if you rebase with `master`  I believe it will address the docs failures you're seeing.

PR #1030, opened 2020-08-11
feat: defines ExternalAccountClient abstract class for external_account credentials

bojeil-google:
Base external account client is used to instantiate `AuthClients` for exchanging external account credentials for GCP access token and authorizing requests to GCP APIs.

The base class implements common logic for exchanging various type of external credentials via GCP STS token exchange endpoint for GCP access tokens. The logic of determining and retrieving the external credential based on the environment and `credential_source` in the credentials JSON file will be left for the subclasses.

### DIFF
--- a/src/auth/externalclient.ts
+++ b/src/auth/externalclient.ts
@@ -98,6 +98,9 @@ export abstract class ExternalAccountClient extends AuthClient {
    * object loaded from an external account credentials file.
    * @param options The external account options object typically loaded
    *   from the external account JSON credential file.
+   * @param additionalOptions Optional additional behavior customization
+   *   options. These currently customize expiration threshold time and
+   *   whether to retry on 401/403 API request errors.
    */
   constructor(
     options: ExternalAccountClientOptions,
@@ -110,11 +113,11 @@ export abstract class ExternalAccountClient extends AuthClient {
           `received "${options.type}"`
       );
     }
-    const clientAuth = options!.client_id
+    const clientAuth = options.client_id
       ? ({
           confidentialClientType: 'basic',
-          clientId: options!.client_id,
-          clientSecret: options!.client_secret,
+          clientId: options.client_id,
+          clientSecret: options.client_secret,
         } as ClientAuthentication)
       : undefined;
     this.stsCredential = new sts.StsCredentials(options.token_url, clientAuth);
--- a/src/auth/externalclient.ts
+++ b/src/auth/externalclient.ts
@@ -23,7 +23,7 @@ import * as stream from 'stream';
 import {Credentials} from './credentials';
 import {AuthClient} from './authclient';
 import {BodyResponseCallback} from '../transporters';
-import {GetAccessTokenResponse, Headers} from './oauth2client';
+import {GetAccessTokenResponse, Headers, RefreshOptions} from './oauth2client';
 import * as sts from './stscredentials';
 import {ClientAuthentication} from './oauth2common';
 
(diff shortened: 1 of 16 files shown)


src/auth/externalclient.ts
```diff
@@ -0,0 +1,313 @@
...
+
+import {Credentials} from './credentials';
+import {AuthClient} from './authclient';
+import {BodyResponseCallback} from '../transporters';
+import {GetAccessTokenResponse, Headers} from './oauth2client';
+import * as sts from './stscredentials';
+import {ClientAuthentication} from './oauth2common';
+
+/**
+ * The required token exchange grant_type: rfc8693#section-2.1
```
bcoe, line 31: I appreciate the links to RFCs, this will be useful for future generations.

src/auth/externalclient.ts
```diff
@@ -0,0 +1,313 @@
...
+/**
+ * The requested token exchange requested_token_type: rfc8693#section-2.1
+ */
+const STS_REQUEST_TOKEN_TYPE = 'urn:ietf:params:oauth:token-type:access_token';
+/** The default OAuth scope to request when none is provided. */
+const DEFAULT_OAUTH_SCOPE = 'https://www.googleapis.com/auth/cloud-platform';
+/**
+ * Offset to take into account network delays and server clock skews.
+ */
+export const EXPIRATION_TIME_OFFSET = 5 * 60 * 1000;
```
bcoe, line 43: Where does the 5 minute value come from? wondering if this is similar to where other clients landed.

src/auth/externalclient.ts
```diff
@@ -0,0 +1,313 @@
...
+  audience: string;
+  subject_token_type: string;
+  service_account_impersonation_url?: string;
+  token_url: string;
+  token_info_url?: string;
+  client_id?: string;
+  client_secret?: string;
+  credential_source: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    [key: string]: any;
```
bcoe, line 61: can we be more specific than `any`?

src/auth/externalclient.ts
```diff
@@ -0,0 +1,313 @@
...
+  private cachedAccessToken: CredentialsWithResponse | null;
+  private readonly stsCredential: sts.StsCredentials;
+
+  /**
+   * Instantiate an ExternalAccountClient instance using the provided JSON
+   * object loaded from an external account credentials file.
+   * @param options The external account options object typically loaded
+   *   from the external account JSON credential file.
+   */
+  constructor(private readonly options: ExternalAccountClientOptions) {
```
bcoe, line 98: do you know if the `readonly` modifier applies to accessing properties on the object? I fixed a variety of bugs in this library related to people modifying an object, not realizing it's passed by reference.

src/auth/externalclient.ts
```diff
@@ -0,0 +1,313 @@
...
+    } catch (e) {
+      const res = (e as GaxiosError).response;
+      if (res) {
+        const statusCode = res.status;
+        // Retry the request for metadata if the following criteria are true:
+        // - We haven't already retried.  It only makes sense to retry once.
+        // - The response was a 401 or a 403
+        // - The request didn't send a readableStream
+        const isReadableStream = res.config.data instanceof stream.Readable;
+        const isAuthErr = statusCode === 401 || statusCode === 403;
```
bcoe, line 233: Do we use this retry logic in a few places, it seems like something we should abstract into a helper.

bcoe (changes requested):
Left a few nits, this is looking solid to me.

src/auth/externalclient.ts
bcoe: I'd even be happy with a tracking issue, just seems like an area for some code cleanup.

bcoe (approved):

bcoe (approved):

bcoe (approved):
LGTM.

PR #1042, opened 2020-08-19
feat: defines `IdentityPoolClient` used for K8s and Azure workloads

bojeil-google:
This will subclass the abstract class `BaseExternalAccountClient` and will retrieve subject tokens from URLs (eg Azure Instance Metadata Service) or a file location (K8s KSA tokens).

Renamed `ExternalAccountClient` to `BaseExternalAccountClient`. The latter will not be exported. Instead the former will be created as an empty class with a static `fromJSON()` method to initialize the right type of external client (`IdentityPoolClient` or `AwsClient`).


test/test.identitypoolclient.ts
```diff
@@ -0,0 +1,530 @@
...
+
+describe('IdentityPoolClient', () => {
+  const fileSubjectToken = fs.readFileSync(
+    './test/fixtures/external-subject-token.txt',
+    'utf-8'
+  );
+  const projectNumber = '123456';
+  const poolId = 'POOL_ID';
+  const providerId = 'PROVIDER_ID';
+  const audience =
```
bcoe, line 128: If all these variables are used by the `audience` URL, and not elsewhere in test, I would be tempted to encapsulate them in a test helper, like `getAudienceUrl`; I read this and my immediate reaction was, that's a lot of variables, before I realized they're not all used in test cases.

src/auth/identitypoolclient.ts
```diff
@@ -0,0 +1,151 @@
...
+
+      if (!fs.lstatSync(filePath).isFile()) {
+        throw new Error();
+      }
+    } catch (err) {
+      err.message = `The file at ${filePath} does not exist, or it is not a file. ${err.message}`;
+      throw err;
+    }
+
+    const stream = fs.createReadStream(filePath);
```
bcoe, line 112: an alternate approach to `createReadStream`, I believe would be using the promises API:

```
const {readFile} = require('fs').promises

async function main () {
  return readFile(filePath)
}
```

bcoe (changes requested):
Left a few nits, but this is looking pretty solid to me.

src/auth/identitypoolclient.ts
bojeil-google: I ended up using `utils.promisify(fs.readFile)` since Node 10 does not seem to play well with `fs.promises`.

test/test.identitypoolclient.ts
bojeil-google: Ok, I made some refactor and moved a lot of these constants to common file and exported helper functions which are used by 2 test files.

bcoe (approved):

PR #1047, opened 2020-08-24
feat: implements AWS signature version 4 for signing requests

bojeil-google:
This is based on:
https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html

It will be used to generate signed requests to AWS GetCallerIdentity API.
https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

This API is used to return details about the IAM user or role whose
credentials are used to call the operation.


src/auth/awsrequestsigner.ts
```diff
@@ -0,0 +1,280 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {GaxiosOptions} from 'gaxios';
+
+import {Headers} from './oauth2client';
+import {Crypto, createCrypto, fromArrayBufferToHex} from '../crypto/crypto';
+
+type HttpMethod =
```
bcoe, line 20: Curious if we use a similar type anywhere else in the codebase, or if `gaxios` exposes it? If not, no objection to having it here.

src/auth/awsrequestsigner.ts
```diff
@@ -0,0 +1,280 @@
...
+  // Format: '%Y%m%dT%H%M%SZ'.
+  const amzDate = now
+    .toISOString()
+    .replace(/[-:]/g, '')
+    .replace(/\.[0-9]+/, '');
+  // Format: '%Y%m%d'.
+  const dateStamp = now.toISOString().replace(/[-]/g, '').replace(/T.*/, '');
+
+  // Change all additional headers to be lower case.
+  const reformattedAdditionalAmzHeaders: Headers = {};
```
bcoe, line 219: This logic is definitely the most complex I think I've bumped into during this review, I believe OAuth 1.0 worked similarly, with regards to  needing to sort and sign parameters?

test/test.awsrequestsigner.ts
```diff
@@ -0,0 +1,297 @@
...
+      ],
+      ProvisionedThroughput: {
+        WriteCapacityUnits: 5,
+        ReadCapacityUnits: 5,
+      },
+    };
+    // List of various requests and their expected signatures.
+    // Examples source:
+    // https://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html
+    const getRequestOptionsTests: AwsRequestSignerTest[] = [
```
bcoe, line 102: I like this DSL you've built out for exercising the the fairly complex logic in the AWS request signer.

bcoe (approved):
Reading through this, I definitely am left wishing there was less complexity 😆 

But I also think the implementation is sound, and doesn't introduce any unnecessary complexity.

src/auth/awsrequestsigner.ts
bojeil-google: It seems that `gaxios` [does not export](https://github.com/googleapis/gaxios/blob/5b586d52b1798b6edf2bec7e59402e4d4115c8ee/src/common.ts#L69) a dedicated type for this.

I couldn't find any similar type in the current repo.

src/auth/awsrequestsigner.ts
bojeil-google: I completely agree. The crypto they use is not trivial and it is not standard. You are right on the similarities with OAuth 1.0 but at least the latter is a standard and there is sufficient documentation on it. 
We had to follow the [AWS documentation](https://docs.aws.amazon.com/general/latest/gr/sigv4-calculate-signature.html) in order to generate these (step by step). They do provide reference input/outputs for each step which helped. 

I think the bulk of the logic is documented in `generateAuthenticationHeaderMap`. I provided links to the 4 steps and some comments to clarify intermediate steps. I hope this makes it easier for contributors to follow. One possibility that was floated around in the early stages was to pull in AWS SDK as a dependency. However, I think it would be too expensive for serverless and low memory environments like GCF, IOT devices, etc. The auth library is a dependency for a lot of other libraries where this feature is not used so it could be very taxing to these users.

test/test.awsrequestsigner.ts
bojeil-google: Thank you. I think this simplifies adding and removing new tests for this logic. It also makes it easier to replace the existing canonical request/signatures we are using. So it should not be expensive to maintain and build on in the future.

test/test.awsrequestsigner.ts
```diff
@@ -100,15 +98,446 @@ describe('AwsRequestSigner', () => {
...
+        getSignedRequest: () => {
+          const signature =
+            'b7a95a52518abbca0964a999a880429ab734f35ebbf1235bd79a5de87756dc4a';
+          return {
+            url: 'https://host.foo.com/',
+            method: 'POST',
+            headers: {
+              Authorization:
+                'AWS4-HMAC-SHA256 Credential=AKIDEXAMPLE/20110909/us-east-1/host/' +
+                `aws4_request, SignedHeaders=date;host;zoo, Signature=${signature}`,
```
bcoe, line 371: thank you for pulling these tests over from boto 👍 sorry for being a pain.

bcoe (approved):

PR #1049, opened 2020-08-28
feat: defines `AwsClient` used for AWS workloads

bojeil-google:
This will subclass the abstract class `BaseExternalAccountClient` and will compute subject tokens as follows:
- Retrieve AWS region from availability-zone.
- Check AWS credentials in environment variables. If not found, get
  from security-credentials endpoint.
- Get AWS credentials from security-credentials endpoint. In order
  to retrieve this, the AWS role needs to be determined by calling
  security-credentials endpoint without any argument. Then the
  credentials can be retrieved via: security-credentials/role_name
- Generate the signed request to AWS STS GetCallerIdentity action.
- Inject x-goog-cloud-target-resource into reformatted header and serialize the
  signed request. This will be the subject-token to pass to GCP STS.


bcoe (approved):
With note that we should make sure we document the environment variables we accept in docs.

bojeil-google:
Thanks for the prompt review, @bcoe. We will add comprehensive documentation under `docs/external-account.md`.

PR #1052, opened 2020-09-03
feat!: integrates external_accounts with `GoogleAuth` and ADC

bojeil-google:
Exports `IdentityPoolClient`, `IdentityPoolClientOptions`, `AwsClient`, `AwsClientOptions`, `ExternalAccountClient` and `ExternalAccountClientOptions`.

Defines `CredentialsClient` interface as the root interface for all clients that generate credentials. Updates `AuthClient` to implement that interface. All existing clients extend `AuthClient` and already conform to `CredentialsClient`.

Exposes `eagerRefreshThresholdMillis` and `forceRefreshOnFailure` on `BaseExternalAccountClient` as they are exposed on  all clients which inherit from `OAuth2Client` and `AuthClient`.

Defines `JSONClient` in `GoogleAuth` for all credentials that can be initialized from JSON objects (as part of ADC): `JWT | UserRefreshClient | BaseExternalAccountClient`. This excludes `Compute` clients.


src/auth/externalclient.ts
```diff
@@ -15,7 +15,6 @@
 import {RefreshOptions} from './oauth2client';
 import {
   BaseExternalAccountClient,
-  BaseExternalAccountClientOptions,
   EXTERNAL_ACCOUNT_TYPE,
```
bcoe, line 18: because of the suffix `type` I thought this would be an interface or `type`, I would add a comment:

```
// this string constant indicates that external account should  be instantiated ?
```

bcoe (changes requested):
I believe this will break the upstream:

https://github.com/googleapis/nodejs-googleapis-common/blob/master/src/authplus.ts

Which wraps GoogleAuth, I suggest we release this work as a  breaking change, just so we can make the update  intentional for upstream consumers.

The upstream consumers need  not take  this as a breaking change.

src/auth/externalclient.ts
bojeil-google: Done. I added clarification here and in `baseexternalclient.ts` where it is declared.

bojeil-google:
Sounds good on the breaking change. I think we can help make the necessary changes in `nodejs-googleapis-common`. We want to make sure all clients are able benefit from this feature.

bcoe (approved):

PR #1054, opened 2020-09-10
feat: default self-signed JWTs

bcoe:
Start using self signed JWTs for client libraries when:

* A service account is being used 
* A scope has not been set by the user
* A target_audience has not been set by the user
* An api_endpoint has not been set by the user

This will require upstream work in the client libraries to begin populating `defaultScopes` rather than `scopes`.

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -27,7 +27,7 @@ import {Compute, ComputeOptions} from './computeclient';
 import {CredentialBody, JWTInput} from './credentials';
 import {IdTokenClient} from './idtokenclient';
 import {GCPEnv, getEnv} from './envDetect';
-import {JWT, JWTOptions} from './jwtclient';
+import {JWT, JWTOptions, DefaultScopesKey} from './jwtclient';
 import {
   Headers,
   OAuth2Client,
@@ -85,7 +85,7 @@ export interface GoogleAuthOptions {
    * Scopes populated by the client library by default. We differentiate between
    * these and user defined scopes when deciding whether to use a self-signed JWT.
    */
-  defaultScopes?: string | string[];
+  [DefaultScopesKey]?: string | string[];
 
   /**
    * Required scopes for the desired API request
@@ -127,7 +127,7 @@ export class GoogleAuth {
 
   private keyFilename?: string;
   private scopes?: string | string[];
-  private defaultScopes?: string | string[];
+  private [DefaultScopesKey]?: string | string[];
   private clientOptions?: RefreshOptions;
 
   /**
@@ -139,8 +139,8 @@ export class GoogleAuth {
     opts = opts || {};
     this._cachedProjectId = opts.projectId || null;
     this.keyFilename = opts.keyFilename || opts.keyFile;
-    this.defaultScopes = opts.defaultScopes;
     this.scopes = opts.scopes;
+    this[DefaultScopesKey] = opts[DefaultScopesKey];
     this.jsonContent = opts.credentials || null;
     this.clientOptions = opts.clientOptions;
   }
@@ -252,7 +252,7 @@ export class GoogleAuth {
     );
     if (credential) {
       if (credential instanceof JWT) {
-        credential.defaultScopes = this.scopes;
+        credential[DefaultScopesKey] = this[DefaultScopesKey];
         credential.scopes = this.scopes;
       }
       this.cachedCredential = credential;
@@ -291,7 +291,7 @@ export class GoogleAuth {
 
     // For GCE, just return a default ComputeClient. It will take care of
     // the rest.
-    (options as ComputeOptions).scopes = this.scopes || this.defaultScopes;
+    (options as ComputeOptions).scopes = this.scopes || this[DefaultScopesKey];
     this.cachedCredential = new Compute(options);
     projectId = await this.getProjectId();
     return {projectId, credential: this.cachedCredential};
@@ -429,7 +429,7 @@ export class GoogleAuth {
     if (json.type === 'authorized_user') {
       client = new UserRefreshClient(options);
     } else {
-      (options as JWTOptions).defaultScopes = this.defaultScopes;
+      (options as JWTOptions)[DefaultScopesKey] = this[DefaultScopesKey];
       (options as JWTOptions).scopes = this.scopes;
       client = new JWT(options);
     }
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -82,6 +82,12 @@ export interface GoogleAuthOptions {
   clientOptions?: JWTOptions | OAuth2ClientOptions | UserRefreshClientOptions;
 
   /**
+   * Scopes populated by a client library by default. We differentiate between
+   * these and user defined scopes when determining the type of auth to use
+   */
+  defaultScopes?: string | string[];
+
+  /**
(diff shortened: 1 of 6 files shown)


src/auth/googleauth.ts
```diff
@@ -81,6 +81,12 @@ export interface GoogleAuthOptions {
    */
   clientOptions?: JWTOptions | OAuth2ClientOptions | UserRefreshClientOptions;
 
+  /**
+   * Scopes populated by the client library by default. We differentiate between
+   * these and user defined scopes when deciding whether to use a self-signed JWT.
+   */
+  defaultScopes?: string | string[];
```
JustinBeckwith, line 88: Huh - how would I differentiate this with the normal `scopes` property?

bshaffer:
This looks great!

src/auth/googleauth.ts
bshaffer: `defaultScopes` are set so the auth library can fall back to using scopes if a self-signed JWT cannot be used. Situations where this happens: Using Metadata credentials or when specifying an API endpoint

bshaffer: I don't see a user ever needing to set these, unless they were using `googleauth` with a custom API.

JustinBeckwith: Ah - the way this is set up though, a user *could* directly modify them. Having defaults as private members is fine, but exposing this publicly on the options is confusing to me at least.  But I concede ignorance here.

bshaffer: Are you proposing `defaultScopes` be set in the constructor only? I agree if so!

bcoe: I'm open to moving  the `defaultScopes` to being only configurable within the constructor  👍 

The user could still define them here.

JustinBeckwith: I am confused :) Do we expect the user to ever set default scopes?  That name feels like something that's internal.

bshaffer: We do not expect users to ever set default scopes

JustinBeckwith: In that case, this property shouldn't exist on `GoogleAuthOptions`, and it should be flagged `private` everywhere I think

bcoe: @JustinBeckwith this doesn't work because we need the client libraries themselves to set the option. What we're differentiating between is the default value set by the client, which itself instantiates the auth client, and scopes passed  from someone doing this:

```js
const {SecretManagerClient} = require('@google-cloud/secret-manager)
new SecretManagerClient({
  scopes: ['some-scopes']
})
```

☝️ this cannot be self signed.

**In the client itself:**

```js
const auth = new GoogleAuth({
 defaultScopes: ['secret-manager-scopes']
})
```

☝️ this should be self signed .

bcoe: Basically, we need to have  passed a value for the default scopes, from the generated client, through to [here](https://github.com/googleapis/gax-nodejs/blob/f7ebfb6fb3f7f1707debc7274f0af57382107ebc/src/grpc.ts#L96), which then needs to be used in instantiating auth client.

bcoe (commented):
I believe this is ready for review, when it comes time to integrate, the plan is that:

* `gax-nodejs` will use the `kDefaultScopes` symbol to provide a default set of scopes. This allows us to differentiate between user-defined and gax-defined scopes, while discouraging users from setting the key themselves (_as you need to go out of your way to do so._).
* `gax-nodejs` will populate `scopes` rather than `kDefaultScopes` if an alternate `servicePath` has been set.

Note: there's an ongoing discussion about simpler audience strings (_this feature will be implemented in a future PR._).

_Update: I ended up simplifying the implementation and going back to a string, based on Alex's feedback._

alexander-fenster (approved):

alexander-fenster (approved):

src/auth/jwtaccess.ts
```diff
@@ -65,12 +69,15 @@ export class JWTAccess {
    * @returns An object that includes the authorization header.
    */
   getRequestHeaders(url: string, additionalClaims?: Claims): Headers {
+    // Return cached authorization headers, unless we are within
+    // EXPIRY_DELTA ms of them expiring:
     const cachedToken = this.cache.get(url);
-    if (cachedToken) {
-      return cachedToken;
+    const now = Date.now();
+    if (cachedToken && cachedToken.expiration - now > EXPIRATION_DELTA) {
```
bshaffer, line 76: question: why is this different from `oauth2client.isTokenExpiring` ? This uses the value `eagerRefreshThresholdMillis` and defaults to 300000 (5 minutes).

test/test.jwt.ts
```diff
@@ -783,4 +784,100 @@ describe('jwt', () => {
...
+        scopes: [],
+        subject: 'bar@subjectaccount.com',
+      });
+      jwt.credentials = {refresh_token: 'jwt-placeholder'};
+      await jwt.getRequestHeaders('https//beepboop.googleapis.com');
+      // The second time we fetch headers should not cause getExpirationTime
+      // to be invoked a second time:
+      await jwt.getRequestHeaders('https//beepboop.googleapis.com');
+      sandbox.assert.calledTwice(getExpirationTime);
+      sandbox.assert.calledTwice(sign);
```
bshaffer, line 880: nit: might be worth having a NOOP test somewhere in the `test.jwt.ts`:

```js
it('returns no headers when no scopes or audiences are provided', async () => {
  const jwt = new JWT({
    email: 'foo@serviceaccount.com',
    key: fs.readFileSync(PEM_PATH, 'utf8'),
    scopes: [],
    subject: 'bar@subjectaccount.com',
  });
  const headers = await jwt.getRequestHeaders();
  assert.equals([], headers);
});

```

src/auth/jwtaccess.ts
```diff
@@ -103,10 +110,24 @@ export class JWTAccess {
...
     return headers;
   }
 
+  /**
+   * Returns an expiration time for the JWT token.
+   *
+   * @param iat The issued at time for the JWT.
+   * @returns An expiration time for the JWT.
+   */
+  static getExpirationTime(iat: number): number {
```
bshaffer, line 126: question: should this be private?

bshaffer (commented):
First pass. Looks great! Mostly just questions.

bshaffer:
Should `|| this.defaultScopes` be added at [this line](https://github.com/googleapis/google-auth-library-nodejs/pull/1054/files#diff-846a54d0fdb67700df505ff12b3a0386R267) as well?

src/auth/jwtaccess.ts
bcoe: No great reason, just thought 5 minutes is perhaps more than we need? perhaps I should be consistent.

bshaffer: totally your call! Just a callout to make sure you were aware the logic is different

bcoe: we're now using `eagerRefreshThresholdMillis` for consistency.

bcoe:
> Should || this.defaultScopes be added at this line as well?

I had missed setting `defaultScopes` in a couple places -- I'm hoping I caught them all now.

src/auth/jwtaccess.ts
bcoe: The `getExpirationTime` method is now `private`.

PR #1058, opened 2020-09-23
chore: sync to master

bojeil-google:
The sync pulls in latest changes in master having to do with self signed JWTs.

In addition, updates external_account clients to also handle `defaultScopes`.


src/auth/googleauth.ts
```diff
@@ -439,10 +445,11 @@ export class GoogleAuth {
         json as ExternalAccountClientOptions,
         options
       )!;
-      client.scopes = this.scopes;
+      client.scopes = this.scopes || this.defaultScopes;
```
bcoe, line 448: do you think we do this enough now that we should consider pulling this into a private helper, like:

```js
private getAnyScopes() {
  return this.scopes || this.defaultScopes
}
```

trying to decide if that makes us less likely to forget the `|| this.defaultScopes` as we refactor in the future.

bojeil-google: Good idea, I created the helper function. We use this in 5 places. I also think this makes it less likely to forget the `|| this.defaultScopes` by using the helper function.

bcoe (approved):

PR #1067, opened 2020-10-07
feat: get AWS region from environment variable

### LINKED ISSUE #1066: Get AWS region for AwsClient from environment variable
I see @bojeil-google is doing great work in the `byoid` branch to bring Workload Identity Federation to google-auth-library. I've been experimenting with it in a AWS Lambda function and ran into the issue that AwsClient is trying to invoke the AWS metadata service to get the region.

But when running in AWS Lambda, there is no metadata service. For the credentials itself the AwsClient first looks at the environment variables AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY and AWS_SESSION_TOKEN. If these env vars are available (as they are in a Lambda) those will be used instead of calling the metadata service for credentials.

It would be great if the AwsClient also looks at AWS_REGION first before making any call to the metadata service. This env var is available when running in AWS Lambda as can be seen at https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html

I'll submit a PR with the suggested change.

As a quick hacky workaround I created my own subclass:
```
class RegionalGoogleAuth extends GoogleAuth {
    async getClient() {
        let client = await super.getClient();
        if (client instanceof AwsClient) {
            const c = client as any;
            if (!c.__getAwsRegion) {
                c.__getAwsRegion = c.getAwsRegion.bind(c);
                c.getAwsRegion = async () => process.env.AWS_REGION ?? c.__getAwsRegion();
            }
            client = c;
        }
        return client;
    }
}
```

wvanderdeijl:
This gets the AWS region from the environment variable AWS_REGION as an alternative to invoking the AWS metadata service which is not available on all AWS products (such as Lambda)

- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-auth-library-nodejs/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Appropriate docs were updated (if necessary): there doesn't seem to be any documentation (yet?) about using AWS_* env vars

Fixes #1066 🦕

### DIFF
--- a/src/auth/awsclient.ts
+++ b/src/auth/awsclient.ts
@@ -199,6 +199,9 @@ export class AwsClient extends BaseExternalAccountClient {
    * @return A promise that resolves with the current AWS region.
    */
   private async getAwsRegion(): Promise<string> {
+    if (process.env['AWS_REGION']) {
+      return process.env['AWS_REGION'];
+    }
     const opts: GaxiosOptions = {
       url: this.regionUrl,
       method: 'GET',
--- a/test/test.awsclient.ts
+++ b/test/test.awsclient.ts
@@ -504,16 +504,19 @@ describe('AwsClient', () => {
     let envAwsAccessKeyId: string | undefined;
     let envAwsSecretAccessKey: string | undefined;
     let envAwsSessionToken: string | undefined;
+    let envAwsRegion: string | undefined;
 
     beforeEach(() => {
       // Store external state.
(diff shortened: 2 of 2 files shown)


bcoe:
@wvanderdeijl thank you for the contribution 👏 , I'm going to let @bojeil-google sign off on this patch, as they've been technical lead on BYOID.

bojeil-google (approved):
Thanks again for contributing with the fix @wvanderdeijl!

PR #1071, opened 2020-10-07
[DO NOT MERGE]: feat(crypto): play with webcrypto API

bcoe:
play with web-crypto API, so that we can provide feedback to Node.js project.

### DIFF
--- a/src/crypto/node/crypto.ts
+++ b/src/crypto/node/crypto.ts
@@ -14,32 +14,100 @@
 
 import * as crypto from 'crypto';
 import {Crypto} from '../crypto';
+import * as base64js from 'base64-js';
+
+interface WebCryptoInterface {
+  subtle: {
+    digest: Function;
+    importKey: Function;
+    verify: Function;
+    sign: Function;
+  };
+  getRandomValues: Function;
+}
+
+const webcrypto = (crypto as any).webcrypto as WebCryptoInterface;
 
 export class NodeCrypto implements Crypto {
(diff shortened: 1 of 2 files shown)


src/crypto/node/crypto.ts
```diff
@@ -14,32 +14,100 @@
...
+    const outputBuffer = await webcrypto.subtle.digest('SHA-256', inputBuffer);
+
+    return base64js.fromByteArray(new Uint8Array(outputBuffer));
   }
 
   randomBytesBase64(count: number): string {
-    return crypto.randomBytes(count).toString('base64');
+    const array = new Uint8Array(count);
+    webcrypto.getRandomValues(array);
+    return base64js.fromByteArray(array);
```
alexander-fenster, line 44: Since we're in Node context, we still can use `Buffer.from` here, right?

bcoe: In playing I could seem to, seems like this might be a good improvement.

src/crypto/node/crypto.ts
```diff
@@ -14,32 +14,100 @@
 
 import * as crypto from 'crypto';
 import {Crypto} from '../crypto';
+import * as base64js from 'base64-js';
```
JustinBeckwith, line 17: So this is cool and all, but the real test would be to delete this file, and rely on the browser context one we've already written :) That is already using SubtleCrypto, so if node is API compatible it should slide right in.  This PR could just be a deleted file!

PR #1073, opened 2020-10-09
use default AWS urls when not set in AwsClientOptions

### LINKED ISSUE #1066: Get AWS region for AwsClient from environment variable

wvanderdeijl:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-auth-library-nodejs/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea: see #1066 

Fixes #1066 🦕

This makes the URLs in the AwsClientOptions.credential_source optional. Most people would never need to supply different values, but this keeps that option available. This is convenient for users, but also makes it less awkward that you have to specify these URL's when using the AWS_* environment variables. When using environment variables, the URL's are never used which makes it awkward that you would have to specify them.

I also added jsdoc comments to all config options which might make it easier for people to configure the large number of options.

PS. I also removed `token_info_url` from `BaseExternalAccountClientOptions` as I couldn't find that it is used anywhere.

### DIFF
--- a/src/auth/awsclient.ts
+++ b/src/auth/awsclient.ts
@@ -21,16 +21,39 @@ import {
 } from './baseexternalclient';
 import {RefreshOptions} from './oauth2client';
 
+const DEFAULT_REGION_URL =
+  'http://169.254.169.254/latest/meta-data/placement/availability-zone';
+const DEFAULT_CREDENTIALS_URL =
+  'http://169.254.169.254/latest/meta-data/iam/security-credentials';
+const DEFAULT_CRED_VERIFICATION_URL =
+  'https://sts.{region}.amazonaws.com?Action=GetCallerIdentity&Version=2011-06-15';
+
+export interface AwsClientCredentialSource {
+  /** indicator that these credentials are intended for an AwsClient, should be set to `aws1` */
+  environment_id: string;
+  /**
+   * URL where to request the current AWS region if it is not set as an environment variable.
+   * Defaults to `http://169.254.169.254/latest/meta-data/placement/availability-zone`
+   */
+  region_url?: string;
+  /**
+   * URL where to request the current AWS credentials if these are not set as environment variables.
+   * Defaults to `http://169.254.169.254/latest/meta-data/iam/security-credentials`
+   */
+  url?: string;
+  /**
+   * URL to AWS STS GetCallerIdentity.
+   * Defaults to `https://sts.{region}.amazonaws.com?Action=GetCallerIdentity&Version=2011-06-15` where `{region}`
+   * is replaced with the actual region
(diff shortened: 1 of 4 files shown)


wvanderdeijl:
This one is probably best reviewed by @bojeil-google as they've been technical lead on BYOID.

bojeil-google:
Thanks for your contribution, @wvanderdeijl. However, at the moment, we don't plan to hardcode URLs. This is a large effort that covers multiple libraries. The consensus was to use an auto-generated configuration file to do so. We also want to future proof the implementation so it could possibly support changes in AWS architecture, etc. This is true for our support for Azure and other future environments.

wvanderdeijl:
Ah, I didn’t consider all the other language libraries that you need to keep in sync. Seems like you plan on having a way to generate the confit file for people, which seems like a reasonable approach.

wvanderdeijl:
The only change that you might want to take a look at is the removal of token_info_url from BaseExternalAccountClientOptions as I couldn't find that it is used anywhere.

bojeil-google:
Yeah, I will do some clean up and make some existing fields optional. The feature is still under development :)

I completely agree with your assessment that developers should not need to manually configure these variables. We plan to generate the file for developers. We will re-use the same ADC env var to point to the file. The experience will be seamless to developers, especially when migrating from service account keys to this solution.

PR #1081, opened 2020-10-21
fix: makes AWS fields optional when determinable from envvars

bojeil-google:
- Makes AWS `credential_source` fields optional when determinable from envvars.
- Cleans up comments and javadocs for AWS request signer.
- Sync to master.


src/auth/awsrequestsigner.ts
```diff
@@ -187,9 +191,17 @@ async function getSigningKey(
...
+ *   operations.
+ * @param host The AWS service URL hostname.
+ * @param canonicalUri The AWS service URL path name.
+ * @param canonicalQuerystring The AWS service URL query string.
+ * @param method The HTTP method used to call this API.
+ * @param region The AWS region.
+ * @param securityCredentials The AWS security credentials.
+ * @param requestPayload The optional request payload if available.
+ * @param additionalAmzHeaders The optional additional headers needed for the
+ *   requested AWS API.
```
JustinBeckwith, line 204: I know this already shipped and isn't part of the review, but I really advise to use a 1-3 parameters in most cases with JavaScript functions.  For consumers, there's often no typing or tooling (for non-TypeScript users), so taking more than a few parameters can be very hard.  It also significantly impacts code refactorability in the future.  I would really prefer we take a single "GenerateAuthHeaderMapOptions" parameter defined as an interface, that can change and grow over time without worrying about positional parameters.

JustinBeckwith (approved):

src/auth/awsrequestsigner.ts
bojeil-google: Thanks for the review and feedback @JustinBeckwith. I agree with you. I made the requested change. This specific function is internal only (it is not exported). So it should not be used externally, but regardless, I updated it as this could grow in the future. Using an object makes it definitely more manageable.

JustinBeckwith (approved):
Thank you!

bcoe (approved):

PR #1122, opened 2021-01-23
test: add integration tests for AWS external account clients

bojeil-google:
Adds integration tests for AWS external account clients.

The setup script `/samples/scripts/externalclient-setup.js` has been extended to also create an AWS provider to the created workload identity pool. This will require an AWS developer account to be set up and an AWS role for [web identity federation](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-idp_oidc.html) to be configured.
The script only needs to be run once. It will output the following to the terminal:
- `AUDIENCE_OIDC`: generated OIDC provider audience.
- `AUDIENCE_AWS`: generated AWS provider audience.
- `AWS_ROLE_ARN`: This is the AWS role for `AssumeRoleWithWebIdentity`.

These constants will need to be copied to the `samples/test/externalclient.test.js` test file. This can be run via `npm run samples-setup` after setting the `GOOGLE_APPLICATION_CREDENTIALS` environment variable and populating the `config` object with the AWS account ID and the AWS role name.

The samples tests, run as part of the existing `samples-test` suite, will simulate a local OIDC resource and test file-sourced and url-sourced credentials. The test will also simulate an AWS environment. It will exchange a Google ID token for AWS security keys. These will be set in the expected AWS environment variable before running the AWS ADC test.

Detailed instructions on how to configure the service account, GCP project and AWS acocunt are provided in `samples/script/externalclient-setup.js`.
Detailed instructions on the external client tests are provided in `samples/test/externalclient.test.js`.


samples/test/externalclient.test.js
```diff
@@ -247,4 +349,41 @@ describe('samples for external-account', () => {
...
+        regional_cred_verification_url:
+          'https://sts.{region}.amazonaws.com?Action=GetCallerIdentity&Version=2011-06-15',
+      },
+    };
+    await writeFile(configFilePath, JSON.stringify(config));
+
+    // Run sample script with GOOGLE_APPLICATION_CREDENTIALS environment
+    // variable pointing to the temporarily created configuration file.
+    // Populate AWS environment variables to simulate an AWS VM.
+    const output = await execAsync('node adc', {
```
bcoe, line 374: nit, I would be tempted to change this to `${process.execPath} adc`, just in case a bin is used other than Node.

bcoe (changes requested):
small nit, this is looking good to me.

samples/test/externalclient.test.js
bojeil-google: Thanks for the tip. This works great! It now also works on my machine too. I also injected the `process.env` into the `env` object. I had to pass it first so it doesn't overwrite the new value of `GOOGLE_APPLICATION_CREDENTIALS`.

bcoe (approved):

PR #1142, opened 2021-02-10
fix: expose missing external account client types/classes

bojeil-google:
Expose `BaseExternalAccountClient` and `BaseExternalAccountClientOptions`

The above base class and interface need to be exposed to unblock downstream libraries dependency on v7.

https://github.com/googleapis/nodejs-googleapis-common/pull/363
https://github.com/googleapis/google-api-nodejs-client/pull/2514
https://github.com/googleapis/nodejs-translate/pull/622
https://github.com/googleapis/gax-nodejs/pull/959

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,6 +52,10 @@ export {
   ExternalAccountClient,
   ExternalAccountClientOptions,
 } from './auth/externalclient';
+export {
+  BaseExternalAccountClient,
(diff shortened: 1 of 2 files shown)


bojeil-google:
@bencoe more changes are needed in the broken client libraries but we can't resolve those issues without first exposing these types/classes.

bcoe (approved):

PR #1147, opened 2021-03-10
build: remove blunderbuss config

sofisl:
We don't really use the assignments, helps keep our bug queue clean!

### DIFF
--- a/system-test/test.kitchen.ts
+++ b/system-test/test.kitchen.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-undef */
 // Copyright 2019 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


bcoe (approved):

sofisl:
@bcoe, had to include a change for the mocha bug that we saw in google-api-nodejs-client as well

system-test/test.kitchen.ts
```diff
@@ -13,7 +14,7 @@
 // limitations under the License.
 
 import * as assert from 'assert';
-import {describe, it, before, after} from 'mocha';
+//import {describe, it, before, after} from 'mocha';
```
bcoe, line 17: The need for commenting this out was likely fixed by:

https://github.com/mochajs/mocha/pull/4599

```suggestion
import {describe, it, before, after} from 'mocha';
```

PR #1153, opened 2021-04-02
fix: do not suppress external project ID determination errors

bojeil-google:
Do not suppress the underlying error, as the error could contain helpful
information for debugging and fixing. This is especially true for
external account creds as in order to get the project ID, the following
operations have to succeed:
1. Valid credentials file should be supplied.
2. Ability to retrieve access tokens from STS token exchange API.
3. Ability to exchange for service account impersonated credentials (if
   enabled).
4. Ability to get project info using the access token from step 2 or 3.

Without surfacing the error, it is harder for developers to determine
which step went wrong.

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -671,11 +671,18 @@ export class GoogleAuth {
       return null;
     }
     const creds = await this.getClient();
-    try {
-      return await (creds as BaseExternalAccountClient).getProjectId();
-    } catch (e) {
-      return null;
-    }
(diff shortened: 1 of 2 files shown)


bojeil-google:
@bcoe I believe this should be a safe change to make. This is an unrecoverable error that is typically thrown during development and setup. It is now returning a clearer error message. An internal team member was having issues with it and surfaced it. Let me know if you have any questions.

bcoe (approved):
I'm comfortable with this change 👍

PR #1165, opened 2021-04-23
fix: update return type to be string vs. number

### LINKED ISSUE #1092: err.code type changes from 'string' to 'number'.
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

1) Is this a client library issue or a product issue?
This is the client library for . We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit the [ Support page]() to reach the most relevant engineers.

2) Did someone already solve this?
  - Search the issues already opened: https://github.com/googleapis/google-auth-library-nodejs/issues
  - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node
  - Search or ask on StackOverflow (engineers monitor these tags): http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js

3) Do you have a support contract?
Please create an issue in the [support console](https://cloud.google.com/support/) to ensure a timely response.

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - OS: Windows 10.0.18362
  - Node.js version: 12.13.0
  - npm version: 6.12.0
  - `google-auth-library` version: 6.1.1.

#### Steps to reproduce

  1. Call google.classroom("v1").courses.list(), or any related API (courses.teacher, courses.student, etc.)
  2. Keep calling until a quota threshold is reached (should get a 429 error code).
  3. Look at the typeof the error.code returned.  It will be a 'number', but should be a 'string'.

Looks like 
   err.code = body.error.code; 
around line 95 in processError() in transporters.js needs to ensure the assigned value is a string.

Line 112 handles this by calling toString().  Might be other lines that need to ensure the code field is a string type.

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

sofisl:
Fixes #1092 
`GaxiosError` returns [code as a string](https://github.com/googleapis/gaxios/blob/18a48fc298c8d0e707b7f6e5cd892cb7c239f1db/src/common.ts#L21), but this method in the auth library overwrites the `code` as a number. We're just updating it to return a string to be consistent.

### DIFF
--- a/src/transporters.ts
+++ b/src/transporters.ts
@@ -141,11 +141,11 @@ export class DefaultTransporter {
         err.message = body.error.errors
           .map((err2: Error) => err2.message)
           .join('\n');
-        err.code = body.error.code;
+        err.code = body.error.code.toString();
         err.errors = body.error.errors;
       } else {
         err.message = body.error.message;
-        err.code = body.error.code || res.status;
+        err.code = body.error.code.toString() || res.status.toString();
       }
     } else if (res && res.status >= 400) {
       // Consider all 4xx and 5xx responses errors.
--- a/test/test.transporters.ts
+++ b/test/test.transporters.ts
@@ -100,7 +100,7 @@ describe('transporters', () => {
     transporter.request({url}, error => {
       scope.done();
       assert.strictEqual(error!.message, 'Error 1\nError 2');
-      assert.strictEqual((error as RequestError).code, 500);
+      assert.strictEqual((error as RequestError).code, '500');
       assert.strictEqual((error as RequestError).errors.length, 2);
       done();
     });


bcoe (changes requested):
I'm a little worried about this one, because when I write error handling for libraries I tend to assume integer error codes, for instance I would expect RequestError.code to be `500` not `"500"`.

I think the appropriate fix would perhaps be switching the type to allow `string` | `number`, we should then make sure that a number error code is not mistakenly turned into a `string` anywhere in the flow of the application.

A string error code like `ENOTFOUND`, should remain a string.

A numeric error code like `500`, should remain a number.

From the time of initially being thrown during a request with gaxios, through to it being caught by a user.

sofisl:
@bcoe, just to clarify, you're not recommending changing any of the existing code, just adding number as a possible type [here](https://github.com/googleapis/gaxios/blob/18a48fc298c8d0e707b7f6e5cd892cb7c239f1db/src/common.ts#L21), right?

The problem is that there is functionality currently that attempts to turn the error into a string, but it gets overwritten by the places identified in this PR. I'm worried that if we change that functionality, we'll break other things that depend on the error code being a string when not being passed through `transporter`

sofisl:
After further discussion, we are going to close this for now but we will open an issue to change the type to accept either a string or number on the next major bump.

PR #1173, opened 2021-05-17
fix: use iam client library to setup test

xil222:
This PR is for implementing using IAM client library for setting up externalClient tests, without making any changes to 
current tests.
The bug could be tracked from b/185501294.
@bojeil-google


package.json
```diff
@@ -23,6 +23,8 @@
     "fast-text-encoding": "^1.0.0",
     "gaxios": "^4.0.0",
     "gcp-metadata": "^4.2.0",
+    "google-auth-library": "^7.0.4",
```
JustinBeckwith, line 26: If these are just being used in samples, you would need to add them to `samples/package.json`.  There should already be a reference to `google-auth-library` there, since it gets linked in as part of the tests.

samples/scripts/externalclient-setup.js
```diff
@@ -84,8 +84,10 @@
 const fs = require('fs');
 const {promisify} = require('util');
 const {GoogleAuth} = require('google-auth-library');
+const {google} = require('googleapis');
 
 const readFile = promisify(fs.readFile);
+const iam = google.iam('v1');
```
JustinBeckwith, line 90: It's unfortunate that we would need to use the `googleapis` npm module here.  @alexander-fenster I don't believe we have a `@google-cloud/nodejs-iam` - but is there supposed to be a way to do this with a GAPIC?

samples/scripts/externalclient-setup.js
```diff
@@ -141,23 +143,15 @@ async function main(config) {
   const auth = new GoogleAuth({
     scopes: 'https://www.googleapis.com/auth/cloud-platform',
   });
-
-  // TODO: switch to using IAM client SDK once v1 API has all the v1beta
-  // changes.
-  // https://cloud.google.com/iam/docs/reference/rest/v1beta/projects.locations.workloadIdentityPools
-  // https://github.com/googleapis/google-api-nodejs-client/tree/master/src/apis/iam
+  const authClient = await auth.getClient();
```
JustinBeckwith, line 146: You can skip this line entirely and say `google.options({auth});` and it will work just the same :)

samples/scripts/externalclient-setup.js
xil222: Hi, @JustinBeckwith, thanks for your feedback.
I am not sure I understand
> 'It's unfortunate that we would need to use the `googleapis` npm module here', 

I am importing the `googleapis` module and using IAM client library inside the module, which is following the requirement above. What are the restrictions here that the IAM client library isn't allowed in this repo, and what's the expectation for the solution?

bojeil-google: Hey @JustinBeckwith, unfortunately `@google-cloud` does not seem to contain the needed IAM APIs: https://github.com/googleapis/google-cloud-node#google-cloud-nodejs-client-libraries
There is `@google-cloud/iam-credentials` but this is for iamcredentials.

Currently, it seems only `googleapis` is documented for IAM: https://cloud.google.com/iam/docs/reference/libraries#iam-client-libraries-reference-nodejs

samples/scripts/externalclient-setup.js
bojeil-google: +1

samples/scripts/externalclient-setup.js
```diff
@@ -141,23 +143,15 @@ async function main(config) {
...
-      `https://iam.googleapis.com/v1beta/projects/${projectId}/` +
-      `locations/global/workloadIdentityPools?workloadIdentityPoolId=${poolId}`,
-    method: 'POST',
-    data: {
-      displayName: 'Test workload identity pool',
-      description: 'Test workload identity pool for Node.js',
-    },
+  response = await iam.projects.locations.workloadIdentityPools.create({
+    parent: `projects/${projectId}/locations/global`,
+    workloadIdentityPoolId: poolId,
```
bojeil-google, line 152: Can you also add a display name and description: https://github.com/googleapis/google-api-nodejs-client/blob/19bd611ca69f334b342565698a8ec8991e956580/src/apis/iam/v1.ts#L2412

This makes it easy to understand who created this pool (It will also work well in the Cloud Console UI).

bojeil-google (changes requested):

samples/scripts/externalclient-setup.js
JustinBeckwith: Agreed.  We talked this through and I think we decided not to publish an IAM GAPIC for now, so this is fine.  Sorry for the confusion.

bojeil-google (approved):
Looks good from my end. Leaving it to repo owners for the final say.

bojeil-google:
@bcoe this is one of the action items from when we added the integration tests before launch. Your input is highly appreciated.

samples/package.json
```diff
@@ -14,6 +14,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "google-auth-library": "^7.1.0",
+    "googleapis": "^73.0.0",
```
bcoe, line 17: Could I get you to try out https://www.npmjs.com/package/@googleapis/iam, I love to stress test our individual modules, and it should improve install size.

bcoe (changes requested):
Just a nit, regarding using the IAM library directly. I appreciate this refactor.

samples/package.json
xil222: Thanks for your feedback. In externalclient-setup.js, I still need to set auth with import from **googleapis**,
`google.options({auth});`
I am not able to setup the auth if only install `@googleapis/iam` for this case.

bcoe: won't this work?

```js
const Iam = require('@googleapis/iam')
const iam = await iam.iam({
    version: 'v1',
    auth
});
const response = await iam.projects.locations.workloadIdentityPools.create({});
```

You can pass auth to the client itself, it doesn't need to be configured globally (_perhaps I'm missing the issue though._).

xil222: yes, this is working. Thank you.

bcoe (approved):

JustinBeckwith (approved):

PR #1177, opened 2021-05-19
feat: add detection for Cloud Run

### LINKED ISSUE #1155: GCPEnv to detect more GCP services
GCPEnv detects cloud run as compute engine. 
Can the library look for unique env vars like K_CONFIGURATION to classify other services

JustinBeckwith:
Fixes #1155.  This adds detection for Cloud Run via the `K_CONFIGURATION` env var.  The order of checks is important in that it:
- First makes sure there's a GCE Metadata server available
- Makes sure we're not on k8s
- Then finally checks for the env var

This should help get rid of false positives via Knative on k8s or non-gcp environments.


alexander-fenster (approved):

src/auth/envDetect.ts
```diff
@@ -62,6 +65,15 @@ function isCloudFunction() {
...
 }
 
+/**
+ * This check only verifies that the environment is running knative.
+ * This must be run *after* checking for Kubernetes, otherwise it will
+ * return a false positive.
+ */
+function isCloudRun() {
+  return !!process.env.K_CONFIGURATION;
+}
```
simonz130, line 75: I guess it's too late now but we should be querying metadata endpoint first and use this env variable only in case metadata endpoint doesn't return the environment

PR #1196, opened 2021-06-23
feat: use self-signed JWTs if alwaysUseJWTAccessWithScope is true

sofisl:
Work for the self-signed JWT implementation
In conjunction with: https://github.com/googleapis/gapic-generator-typescript/pull/908


sofisl:
Still need to add tests

.gitignore
```diff
@@ -13,3 +13,4 @@ package-lock.json
 yarn.lock
 dist/
 __pycache__
+.DS_Store
```
JustinBeckwith, line 16: Please add a newline to make the (-) go away

src/auth/googleauth.ts
```diff
@@ -111,6 +111,8 @@ export class GoogleAuth {
    * @private
    */
   private checkIsGCE?: boolean = undefined;
+  useJWTAccessAlways?: boolean;
```
JustinBeckwith, line 114: Can we bike shed this name a little bit?  From a customer perspective, I'm not sure if I'd know what `useJWTAccessAlways` would mean.  Would it make more sense to use something like `useSelfSignedJWT`?

src/auth/googleauth.ts
```diff
@@ -111,6 +111,8 @@ export class GoogleAuth {
...
+  defaultServicePath: string | undefined;
```
JustinBeckwith, line 115: If something is `type | undefined`, we typically use the `?` like we do elsewhere

src/auth/jwtclient.ts
```diff
@@ -26,13 +26,15 @@ import {
 } from './oauth2client';
 
 export interface JWTOptions extends RefreshOptions {
+  useJWTAccessAlways?: boolean;
```
JustinBeckwith, line 29: It would be nice to add descriptions to these properties so they show up in the documentation.  I know we don't have any existing ones here yet :/

src/auth/jwtclient.ts
```diff
@@ -68,7 +71,9 @@ export class JWT extends OAuth2Client implements IdTokenProvider {
     key?: string,
     scopes?: string | string[],
     subject?: string,
-    keyId?: string
+    keyId?: string,
+    useJWTAccessAlways?: boolean,
```
JustinBeckwith, line 75: Let's just stop adding new variables to the constructor, and if they want this, force folks to use the object approach.  This was largely here for back-compat.

src/auth/googleauth.ts
```diff
@@ -456,6 +458,8 @@ export class GoogleAuth {
     } else {
       (options as JWTOptions).scopes = this.scopes;
       client = new JWT(options);
+      client.defaultServicePath = this.defaultServicePath;
+      client.useJWTAccessAlways = this.useJWTAccessAlways || false;
```
JustinBeckwith, line 462: You shouldn't need to include the ` || false` here.  The `||` checks if the variable has a falsey value, which would get evaluated to `false` anyways.

JustinBeckwith: here and below

src/auth/jwtclient.ts
```diff
@@ -121,44 +128,63 @@ export class JWT extends OAuth2Client implements IdTokenProvider {
...
+      if (!this.access) {
+        this.access = new JWTAccess(
+          this.email,
+          this.key,
+          this.keyId,
+          this.eagerRefreshThresholdMillis
+        );
+      }
+
+      const audience = 'https://' + this.defaultServicePath + '/';
```
JustinBeckwith, line 141: Please use string interpolation

src/auth/googleauth.ts
bcoe: @JustinBeckwith I believe this to be consistent with other language's implementation.

My recommendation was that we allow this as a class variable, but don't advertise it as an option on the options interface. The reason being that we want to be able to set this option in GAPIC libraries, but don't want to encourage customers to set the variable.

src/auth/jwtclient.ts
bcoe: I'd like to advocate that we don;'t advertise this option in any interfaces, and just set it on the class.

This is what we did for `defaultScopes`.

src/auth/googleauth.ts
JustinBeckwith: okies - my feedback ain't blocking, and consistency is good

test/test.googleauth.ts
```diff
@@ -402,6 +402,20 @@ describe('googleauth', () => {
       assert.strictEqual(json.private_key, (result as JWT).key);
     });
 
+    it.only('fromJSON should set default access variable with private key', () => {
```
JustinBeckwith, line 405: blocking so we don't merge with the `.only`

JustinBeckwith (changes requested):

test/test.googleauth.ts
sofisl: yep, sorry still haven't pushed the new changes!

sofisl:
@bcoe @JustinBeckwith, comments  have been addressed & tests added!

src/auth/googleauth.ts
```diff
@@ -580,6 +580,8 @@ export class GoogleAuth {
                 ...this.clientOptions,
                 keyFile: this.keyFilename,
               });
+              client.defaultServicePath = this.defaultServicePath;
```
bcoe, line 583: we should need this because we have the helper `client.setGapicJWTValues`.

src/auth/googleauth.ts
```diff
@@ -600,6 +602,8 @@ export class GoogleAuth {
   fromAPIKey(apiKey: string, options?: RefreshOptions): JWT {
     options = options || {};
     const client = new JWT(options);
+    client.defaultServicePath = this.defaultServicePath;
```
bcoe, line 605: I think here you just want to call `setGapicJWTValues()`.

bcoe (commented):
LGTM, thank you for doing this.

sofisl:
Waiting for: https://github.com/googleapis/gapic-generator-typescript/pull/908

bshaffer (commented):
Same feedback here as in the gapic generator: I'd advocate for all of these to be renamed from `useJWTAccessAlways` to `useJwtAccessWithScope` for consistency (see [AIP #4111](https://google.aip.dev/auth/4111#scope-vs-audience)), unless this was a conscious decision with justification.

bshaffer:
> Same feedback here as in the gapic generator: I'd advocate for all of these to be renamed from `useJWTAccessAlways` to `useJwtAccessWithScope` for consistency (see [AIP #4111](https://google.aip.dev/auth/4111#scope-vs-audience)), unless this was a conscious decision with justification.

As https://github.com/googleapis/google-auth-library-nodejs/pull/1204 is already released and uses the name `useJWTAccessAlways`, it's ok to keep that variable name here, and this should not block rolling the feature out!

src/auth/jwtclient.ts
```diff
@@ -122,44 +122,63 @@ export class JWT extends OAuth2Client implements IdTokenProvider {
   protected async getRequestMetadataAsync(
     url?: string | null
   ): Promise<RequestMetadataResponse> {
-    if (!this.apiKey && !this.hasUserScopes() && url) {
-      if (
-        this.additionalClaims &&
-        (
-          this.additionalClaims as {
-            target_audience: string;
+    if (this.useJWTAccessAlways) {
```
bshaffer, line 125: This is interesting logic here because it's saying that for clients which set `useJWTAccessAlways` to `true`, do not allow API Key auth or ID token auth. I'm not sure this is what we want to go so far as to say that. The more accurate way (I think) would be something like this:

```js
var canSelfSign = (this.hasUserScopes() && this.useJWTAccessAlways) // self-signed JWT with scopes
  || (this.hasDefaultScopes() && url) // self-signed JWT with audience
if (!this.apiKey && canSelfSign) {
```

Where it used to have:

```js
if (!this.apiKey && !this.hasUserScopes() && url) {
```

sofisl: @bshaffer, I gave this a try earlier today and tests fail, both the ones we've added and the current functionality.

I think this is because we check later if we should sign with audience or scopes, in `jwtaccess.ts`, line 99:

```
if (scopes) {
      defaultClaims = {
        iss: this.email,
        sub: this.email,
        scope: scopes,
        exp,
        iat,
      };
    } else {
      // If either scopes are not specified, or if audience is something unique,
      // just use default service endpoint.
      defaultClaims = {
        iss: this.email,
        sub: this.email,
        aud: url,
        exp,
        iat,
      };
    }
```

> This is interesting logic here because it's saying that for clients which set useJWTAccessAlways to true, do not allow API Key auth or ID token auth

I'm curious about this, isn't a part of this RFC to always use self-signed JWT if it's coming from a GAPIC client?

bcoe (approved):

JustinBeckwith (approved):

bcoe (changes requested):
I believe @bshaffer is correct about the `authToken` edge case.

bcoe (approved):
To me this is looking really good, and I think the tests are comprehensive.

It would be good to have Brent do one more sniff test.

bcoe (changes requested):
we have a final session planned Monday to get this over finish line.

test/test.jwt.ts
```diff
@@ -22,6 +22,7 @@ import * as sinon from 'sinon';
 import {GoogleAuth, JWT} from '../src';
 import {CredentialRequest, JWTInput} from '../src/auth/credentials';
 import * as jwtaccess from '../src/auth/jwtaccess';
+import {stub} from 'sinon';
```
bcoe, line 25:
```suggestion
```

bcoe (approved):

src/auth/jwtclient.ts
```diff
@@ -148,10 +152,22 @@ export class JWT extends OAuth2Client implements IdTokenProvider {
...
+        if (this.hasUserScopes()) {
+          scopes = this.scopes;
+        } else if (!url) {
+          scopes = this.defaultScopes;
+        }
+
         const headers = await this.access.getRequestHeaders(
-          url,
-          this.additionalClaims
+          url ?? undefined,
```
bshaffer, line 164: NodeJS question: why isn't it enough to just pass in the falsy value (e.g. `url` without the `?? undefined` check)? Is it for cases where it's `false` or an empty string?

bshaffer (approved):
One question, otherwise this looks amazing!

src/auth/jwtclient.ts
bcoe: `null` and `false` are considered valid arguments for an optional string parameter, which can only be a string or `undefined`.

PR #1201, opened 2021-06-30
feat: Implement DownscopedClient#getAccessToken() and unit test

xil222:
Implement DownscopedClient class getAccessToken() API and related unit tests.
Design Doc: [link](https://docs.google.com/document/d/1MJb9SH7cgGQYTUb6cShtMoOuaU3b05XtU4BVijGVok0/edit?resourcekey=0-K6R_URgisKyatWB6cXPK5g#heading=h.rjoi9m8cpb7k)


xil222:
@bojeil-google

SurferJeffAtGoogle (approved):

bcoe:
@xil222  thank you for the contribution 👏  excited for this.

src/auth/downscopedclient.ts
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  /**
+   * Provides a request implementation with OAuth 2.0 flow. In cases of
+   * HTTP 401 and 403 responses, it automatically asks for a new access token
+   * and replays the unsuccessful request.
+   * @param opts Request options.
+   * @param callback callback.
+   * @return A promise that resolves with the HTTP response when no callback
+   * is provided.
```
bojeil-google, line 181: Indent 2 more spaces.

src/auth/downscopedclient.ts
```diff
@@ -0,0 +1,254 @@
...
+    opts: GaxiosOptions,
+    callback?: BodyResponseCallback<T>
+  ): GaxiosPromise<T> | void {
+    throw new Error('Not implemented.');
+  }
+
+  /**
+   * Forces token refresh, even if unexpired tokens are currently cached.
+   * GCP access tokens are retrieved from authclient object/source credential.
+   * Thenm GCP access tokens are exchanged for downscoped access tokens via the
```
bojeil-google, line 195: Typo "Thenm"

src/auth/downscopedclient.ts
```diff
@@ -0,0 +1,254 @@
...
+   * token exchange endpoint.
+   * @return A promise that resolves with the fresh downscoped access token.
+   */
+  protected async refreshAccessTokenAsync(): Promise<CredentialsWithResponse> {
+    // Retrieve GCP access token from source credential.
+    const subjectToken = await (await this.authClient.getAccessToken()).token;
```
bojeil-google, line 201: Why do you need another `await` here? I think this should be sufficient:
`const subjectToken = (await this.authClient.getAccessToken()).token;`

src/auth/downscopedclient.ts
```diff
@@ -0,0 +1,254 @@
...
+   */
+  private cachedDownscopedAccessToken: CredentialsWithResponse | null;
+  private readonly stsCredential: sts.StsCredentials;
+  public readonly authClient: AuthClient;
+  public readonly credentialAccessBoundary: CredentialAccessBoundary;
+  public readonly eagerRefreshThresholdMillis: number;
+  public readonly forceRefreshOnFailure: boolean;
+
+  constructor(
+    private client: AuthClient,
```
bojeil-google, line 91: Why don't you call it `private authClient: AuthClient`
This way you don't need to call `this.authClient = client;` later.

src/auth/downscopedclient.ts
```diff
@@ -0,0 +1,254 @@
...
+    private cab: CredentialAccessBoundary,
```
bojeil-google, line 92: Why don't you call it `private credentialAccessBoundary: CredentialAccessBoundary`
This way you don't need to call `this.credentialAccessBoundary = cab;` later.

src/auth/downscopedclient.ts
```diff
@@ -0,0 +1,254 @@
...
+    const stsResponse = await this.stsCredential.exchangeToken(
+      stsCredentialsOptions,
+      undefined,
+      this.credentialAccessBoundary
+    );
+
+    // Save response in cached access token.
+    this.cachedDownscopedAccessToken = {
+      access_token: stsResponse.access_token,
+      expiry_date: new Date().getTime() + stsResponse.expires_in * 1000,
```
bojeil-google, line 220: Note, `expires_in` may not always be available. In that case you can use the authClient expiration time.

src/auth/downscopedclient.ts
```diff
@@ -0,0 +1,254 @@
...
+   * Provides a mechanism to inject Downscoped access tokens directly.
+   * When the provided credential expires, a new credential, using the
+   * external account options are retrieved.
+   * Notice DownscopedClient is the broker class mainly used for generate
+   * downscoped access tokens, it is unlikely we call this function in real
+   * use case.
+   * We implement to make this a helper function for testing all cases in getAccessToken().
+   * @param credentials The Credentials object to set on the current client.
+   */
+  setCredentials(credentials: Credentials) {
```
bojeil-google, line 140: To make things simple. Let's require the provided credentials to have an expiration time. Can you throw an error if they don't?

src/index.ts
```diff
@@ -22,6 +22,7 @@ export {
 } from './auth/credentials';
 export {GCPEnv} from './auth/envDetect';
 export {GoogleAuthOptions, ProjectIdCallback} from './auth/googleauth';
+export {DownscopedClient} from './auth/downscopedclient';
```
bojeil-google, line 25: You shouldn't have exported this yet. This is a feature under construction. Please revert this. This is the last step of the implementation.

test/externalclienthelper.ts
```diff
@@ -51,6 +51,7 @@ const poolId = 'POOL_ID';
 const providerId = 'PROVIDER_ID';
 const baseUrl = 'https://sts.googleapis.com';
 const path = '/v1/token';
+const betaPath = '/v1beta/token';
```
bojeil-google, line 54: We don't need this. Please use GA endpoint.

test/externalclienthelper.ts
```diff
@@ -75,6 +76,26 @@ export function mockStsTokenExchange(
   return scope;
 }
 
+export function mockStsBetaTokenExchange(
```
bojeil-google, line 79: Please use the GA endpoint.

test/test.downscopedclient.ts
```diff
@@ -0,0 +1,406 @@
...
+}
+
+describe('DownscopedClient', () => {
+  let clock: sinon.SinonFakeTimers;
+
+  const auth = new GoogleAuth({
+    keyFilename: './test/fixtures/private.json',
+    scopes: 'https://www.googleapis.com/auth/cloud-platform',
+  });
+  let client: AuthClient;
```
bojeil-google, line 40: I think you don't need to do this. This reads from a real location and generates tokens.
It is better to define a dummy class which meets the AuthClient interface.

See how I did it in Python: https://github.com/googleapis/google-auth-library-python/blob/63ac08afe44c3da51a6fb3bfc2fcdb6588e040ef/tests/test_downscoped.py#L82 

This will also simplify the implementation below. You wouldn't need to stub its underlying implementation if you do so.

test/test.downscopedclient.ts
```diff
@@ -0,0 +1,406 @@
...
+        },
+      ],
+    },
+  };
+  const stsSuccessfulResponse: StsSuccessfulResponse = {
+    access_token: 'DOWNSCOPED_CLIENT_ACCESS_TOKEN_0',
+    issued_token_type: 'urn:ietf:params:oauth:token-type:access_token',
+    token_type: 'Bearer',
+    expires_in: ONE_HOUR_IN_SECS,
+    scope: 'scope1 scope2',
```
bojeil-google, line 70: Let's remove the scope field. This is not [documented](https://cloud.google.com/iam/docs/downscoping-short-lived-credentials#create-credential) in the response.

test/test.downscopedclient.ts
```diff
@@ -0,0 +1,406 @@
...
+        accessBoundary: {
+          accessBoundaryRules: [],
+        },
+      };
+      assert.throws(() => {
+        return new DownscopedClient(client, cabWithEmptyAccessBoundaryRules);
+      }, expectedError);
+    });
+
+    it('should throw on exceed number of access boundary rules', () => {
```
bojeil-google, line 103: should throw when number of access boundary rules is exceeded

test/test.downscopedclient.ts
```diff
@@ -0,0 +1,406 @@
...
+      const expectedError = new Error(
+        'Access boundary rule exceeds limit, max 10 allowed.'
+      );
+      const cabWithExceedingAccessBoundaryRules = {
+        accessBoundary: {
+          accessBoundaryRules: [] as any,
```
bojeil-google, line 109: You don't need the `as any`

test/test.downscopedclient.ts
```diff
@@ -0,0 +1,406 @@
...
+        },
+      };
+      assert.throws(() => {
+        return new DownscopedClient(client, cabWithEmptyAccessBoundaryRules);
+      }, expectedError);
+    });
+
+    it('should throw on exceed number of access boundary rules', () => {
+      const expectedError = new Error(
+        'Access boundary rule exceeds limit, max 10 allowed.'
```
bojeil-google, line 105: Can you define the limit `10` as a global const above?

test/test.downscopedclient.ts
```diff
@@ -0,0 +1,406 @@
...
+        refreshOptions.eagerRefreshThresholdMillis
+      );
+    });
+  });
+
+  describe('getAccessToken()', () => {
+    let sandbox: sinon.SinonSandbox;
+    before(() => {
+      const expectedAccessTokenResponse = {
+        token: 'subject_token',
```
bojeil-google, line 276: This is the final response. It is not the subject token. It is the access token. The subject token is exchanged for an access token.

test/test.downscopedclient.ts
```diff
@@ -0,0 +1,406 @@
...
+          statusCode: 200,
+          response: stsSuccessfulResponse,
+          request: {
+            grant_type: 'urn:ietf:params:oauth:grant-type:token-exchange',
+            requested_token_type:
+              'urn:ietf:params:oauth:token-type:access_token',
+            subject_token: 'subject_token',
+            subject_token_type: 'urn:ietf:params:oauth:token-type:access_token',
+            options:
+              testClientAccessBoundary &&
```
bojeil-google, line 321: You don't need this. This is always set above. It is a const. You can just directly stringify.

test/test.downscopedclient.ts
```diff
@@ -0,0 +1,406 @@
...
+      const refreshedTokenResponse = await downscopedClient.getAccessToken();
+
+      assert.deepStrictEqual(
+        refreshedTokenResponse.token,
+        stsSuccessfulResponse.access_token
+      );
+      scope.done();
+    });
+
+    it('should return new DownscopedClient access token when there is no cached downscoped access token', async () => {
```
bojeil-google, line 347: nit: This can be shortened
should return new access token when no cached token is available

test/test.downscopedclient.ts
```diff
@@ -0,0 +1,406 @@
...
+          statusCode: 200,
+          response: stsSuccessfulResponse,
+          request: {
+            grant_type: 'urn:ietf:params:oauth:grant-type:token-exchange',
+            requested_token_type:
+              'urn:ietf:params:oauth:token-type:access_token',
+            subject_token: 'subject_token',
+            subject_token_type: 'urn:ietf:params:oauth:token-type:access_token',
+            options:
+              testClientAccessBoundary &&
```
bojeil-google, line 359: You don't need this. This is always set above. It is a const. You can just directly stringify.

src/auth/downscopedclient.ts
```diff
@@ -0,0 +1,254 @@
...
+    super.setCredentials(credentials);
+    this.cachedDownscopedAccessToken = credentials;
+  }
+
+  async getAccessToken(): Promise<GetAccessTokenResponse> {
```
bojeil-google, line 145: I think there is a problem in just returning `GetAccessTokenResponse` since it doesn't expose the expiration time.
Can we return an extended interface of `GetAccessTokenResponse` with an `expirationTime` field? We can use epoch time in seconds. This is needed so the token broker can expose this to the token consumer. Otherwise the token consumer can't accurately get this.

test/test.downscopedclient.ts
```diff
@@ -0,0 +1,406 @@
...
+      sandbox
+        .stub(client, 'getAccessToken')
+        .resolves(expectedAccessTokenResponse);
+    });
+
+    after(() => {
+      sandbox.restore();
+    });
+
+    it('should return current unexpired cached DownscopedClient access token', async () => {
```
bojeil-google, line 288: Can you also add the following tests:
- Test when STS rejects with an error.
- Test when the source AuthClient rejects when a token is requested.

src/auth/downscopedclient.ts
```diff
@@ -0,0 +1,254 @@
...
+   * used.
+   */
+  private cachedDownscopedAccessToken: CredentialsWithResponse | null;
+  private readonly stsCredential: sts.StsCredentials;
+  public readonly authClient: AuthClient;
+  public readonly credentialAccessBoundary: CredentialAccessBoundary;
+  public readonly eagerRefreshThresholdMillis: number;
+  public readonly forceRefreshOnFailure: boolean;
+
+  constructor(
```
bojeil-google, line 90: I think we need to allow the quota_project_id to be provided as optional in the header. You can add this later though.

src/auth/downscopedclient.ts
xil222: that part is in another PR I sent yesterday:
https://github.com/googleapis/google-auth-library-nodejs/pull/1216

test/test.downscopedclient.ts
xil222: This is the subject token which is the access token from source credential.
And I am using it to exchange for downscoped access token

xil222: I'll change the variable name to avoid confusion.

test/test.downscopedclient.ts
xil222: I have to keep this, [] by default is type never,
Otherwise I have to export CAB interface and import it in the test.
I don't think it is good to export more unnecessary interfaces.

src/auth/downscopedclient.ts
```diff
@@ -0,0 +1,254 @@
...
+ * Defines an upper bound of permissions available for a GCP credential.
+ */
+interface CredentialAccessBoundary {
+  accessBoundary: {
+    accessBoundaryRules: AccessBoundaryRule[];
+  };
+}
+
+/** Defines an upper bound of permissions on a particular resource. */
+interface AccessBoundaryRule {
```
bojeil-google, line 61: I forgot to mention. These CAB rule interfaces have to also be exported as they are externally visible. Of course we should do this when we are ready for launch.

test/test.downscopedclient.ts
xil222: Yea, I understand and have thought about this. 
In this case, language might differ, this approach could save a 3 line stub for calling getAccessToken().
However, AuthClient has 3 abstract APIs, implementing the other two abstracts APIs getRequestHeader and request which won't get called in tests is less efficient compared to calling stubs.

src/auth/downscopedclient.ts
xil222: I'll add this part as well as test for headers during full implementation.

PR #1204, opened 2021-07-02
feat: add useJWTAccessAlways and defaultServicePath variable

sofisl:
Piecemeal version of #1196

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -152,7 +152,10 @@ export class GoogleAuth {
     this.clientOptions = opts.clientOptions;
   }
 
-  setDefaultJWTValues(client: JWT) {
+  // GAPIC client libraries should always use self-signed JWTs. The following
+  // variables are set on the JWT client in order to indicate the type of library,
+  // and sign the JWT with the correct audience and scopes (if not supplied).
+  setGapicJWTValues(client: JWT) {
     client.defaultServicePath = this.defaultServicePath;
     client.useJWTAccessAlways = this.useJWTAccessAlways;
     client.defaultScopes = this.defaultScopes;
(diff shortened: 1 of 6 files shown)


src/auth/googleauth.ts
```diff
@@ -583,6 +591,8 @@ export class GoogleAuth {
     options = options || {};
     const client = new JWT(options);
     client.fromAPIKey(apiKey);
+    client.defaultServicePath = this.defaultServicePath;
```
bcoe, line 594: I think this is a special edge case where we don't need to set `useJWTAccessAlways`, where we're providing an explicit API key to use (which I'm betting is why we don't set defaultServicePath either).

bcoe (changes requested):
There are two places I notice where we set defaultScopes but don't set ` client.useJWTAccessAlways = this.useJWTAccessAlways;`:

https://github.com/googleapis/google-auth-library-nodejs/blob/57e4dad9e9897c6ae09b05d2f1224e79cb4f6c01/src/auth/googleauth.ts#L270

and 

https://github.com/googleapis/google-auth-library-nodejs/blob/57e4dad9e9897c6ae09b05d2f1224e79cb4f6c01/src/auth/googleauth.ts#L286

I was also concerned by:

https://github.com/googleapis/google-auth-library-nodejs/pull/1204/files#diff-535a540500eec341838bd88c84b1795313a8ff9b556bad87ef89fb2fca061630R573

Where I notice you set `useJWTAccessAlways`, but we miss populating `defaultScopes`. Should we be setting `defaultScopes` here, or I wonder did we already set it in a related code path so we don't need to? we should probably double check.

To help protect future generations from having hiccups like these, why don't we add a private helper:

`setAdditionalJWTConfig` or something along those lines, which always sets all three variables?

bcoe (approved):

PR #1207, opened 2021-07-07
feat(impersonated): add impersonated credentials auth

### LINKED ISSUE #535: Implement ImpersonatedCredentials

bcoe:
Adds ImpersonatedCredentials. This credential type basically takes one source credential and exchanges it for another, different service account.

Fixes #535
Refs #779

---

@salrashid123 I've gone through the original #779 and made some edits to make the work more idiomatic for Node.js.

I'm leaving some review comments on this PR that would get it ready to land.


src/auth/googleauth.ts
```diff
@@ -44,7 +45,11 @@ import {AuthClient} from './authclient';
  * Defines all types of explicit clients that are determined via ADC JSON
  * config file.
  */
-export type JSONClient = JWT | UserRefreshClient | BaseExternalAccountClient;
+export type JSONClient =
+  | JWT
+  | UserRefreshClient
+  | BaseExternalAccountClient
+  | Impersonated;
```
bcoe, line 52: We will want to make sure that this doesn't cause compilation errors for existing libraries like `googleapis` and `nodejs-secret-manager`.

Depending, we may need to treat this as a major change.

src/auth/impersonated.ts
```diff
@@ -0,0 +1,194 @@
...
+   * @param refreshToken Unused parameter
+   */
+  protected async refreshToken(
+    refreshToken?: string | null
+  ): Promise<GetTokenResponse> {
+    const iat = Date.now();
+    if (this.credentials.expiry_date) {
+      if (this.credentials.expiry_date <= iat) {
+        try {
+          await this.sourceClient.getAccessToken();
```
bcoe, line 101: I switched from `.then`, `.catch` to `async`/`await` which is more idiomatic with how we tend to write JavaScript.

src/auth/impersonated.ts
```diff
@@ -0,0 +1,194 @@
...
+          this.credentials.access_token = tokenResponse.accessToken;
+          this.credentials.expiry_date =
+            Date.parse(tokenResponse.expireTime) / 1000;
+          return {
+            tokens: this.credentials,
+            res,
+          };
+        } catch (error) {
+          if (error.status === 403) {
+            if (error.message === 'The caller does not have permission') {
```
bcoe, line 124: Checking for specific error messages feels fragile, what if we check for just the 403 case, and provide an appropriate error message?

src/auth/impersonated.ts
```diff
@@ -0,0 +1,194 @@
...
+  protected requestAsync<T>(
+    opts: GaxiosOptions,
+    retry = false
+  ): GaxiosPromise<T> {
+    try {
+      return super.requestAsync<T>(opts, retry);
+    } catch (err) {
+      let helpfulMessage = null;
+      if (err.status === 403) {
+        helpfulMessage =
```
bcoe, line 155: Likewise, perhaps we can get away with checking for just the `403` and `404` case?

test/test.impersonated.ts
```diff
@@ -0,0 +1,72 @@
...
+    .post('/oauth2/v4/token')
+    .reply(200, body);
+}
+
+describe('impersonated', () => {
+  afterEach(() => {
+    nock.cleanAll();
+  });
+  it('should request impersonated credentials on first request');
+  it.only('should refresh if access token has expired', async () => {
```
bcoe, line 40: There's a small enough amount of setup behavior, that I think it's better to configure each test case in its specific assertion.

test/test.impersonated.ts
```diff
@@ -0,0 +1,72 @@
...
+    const impersonated = new Impersonated({
+      sourceClient: jwt,
+      targetPrincipal: 'target@project.iam.gserviceaccount.com',
+      lifetime: 30,
+      delegates: [],
+      targetScopes: ['https://www.googleapis.com/auth/cloud-platform'],
+    });
+    impersonated.credentials.access_token = 'initial-access-token';
+    impersonated.credentials.expiry_date = Date.now() - 10000;
+    await impersonated.request({url});
```
bcoe, line 64: This does not yet work because we need to mock the:

`iamcredentials.googleapis.com:443/v1/projects/-/serviceAccounts/target@project.iam.gserviceaccount.com:generateAccessToken`

test/test.impersonated.ts
```diff
@@ -0,0 +1,72 @@
...
+    assert.strictEqual(impersonated.credentials.access_token, 'abc123');
+    scopes.forEach(s => s.done());
+  });
+  it(
+    'should throw appropriate exception when 403 occurs refreshing impersonated credentials'
+  );
+  it('should throw appropriate exception when 403 occurs during request');
```
bcoe, line 71: I've added some additional recommendations for test cases, one additional one I can think of is the 404 case.

bcoe (commented):
@salrashid123 sorry for screwing up your branch, I left some refactoring and recommendations in this PR, to act as a baseline.

One additional thought, we should probably add a blurb in the README about how to use the impersonated client?

salrashid123:
thanks,  i retested the code from this PR and it worked fine with the snippet below

1,2 wokred fine but ...how do you supply the impersonated client to a GCP library like `@google-cloud/logging`?

i do know in the original PR i showed 3 working below but ...i must've just had that as a place holder or somthing of a wishlist

is it possible to supply  `Impersonated` to any nodejs library?

i looked at but those credentials and what not are something else like a keyfile
https://googleapis.dev/nodejs/logging/latest/global.html#ClientConfig
https://googleapis.dev/nodejs/google-auth-library/latest/interfaces/GoogleAuthOptions.html


```javascript
const {Logging} = require('@google-cloud/logging');
const { GoogleAuth, JWT, Impersonated  } = require('google-auth-library');
const { Storage } = require('@google-cloud/storage');

var svcAccountFile = '/home/srashid/gcp_misc/certs/mineral-minutia-820-e9a7c8665867.json';
var project_id = 'fabled-ray-104117';
const keys = require(svcAccountFile);

// create a sourceCredential
let saclient = new JWT(
  keys.client_email,
  null,
  keys.private_key,
  ['https://www.googleapis.com/auth/cloud-platform', 'https://www.googleapis.com/auth/iam'],
);

// Use that to impersonate the targetPrincipal
  sourceClient: saclient,
  targetPrincipal: "impersonated-account@fabled-ray-104117.iam.gserviceaccount.com",
  lifetime: 30,
  delegates: [],
  targetScopes: ["https://www.googleapis.com/auth/cloud-platform"]
});


// At this point you can use the client as any other:

// 1. Acquire Headers
const authHeaders = await targetClient.getRequestHeaders();
console.log(authHeaders);

// 2. Use client in authorized session 
targetClient.getAccessToken().then(res => {
  let url = 'https://www.googleapis.com/storage/v1/b?project=' + project_id
  targetClient.requestAsync({ url }).then(resp => {
    console.log(resp.data.items[0]);
  }).catch(function (error) {
    console.error('Unable to list buckets: ' + error);
  });
});

// 3. Use client with GCP Service 

const logging = new Logging({projectId: project_id});
const log = logging.log('mylog');
const text = 'Hello, world!';
const metadata = {
  resource: {type: 'global'},
};
const entry = log.entry(metadata, text);
await log.write(entry);
console.log(`Logged: ${text}`);

```

salrashid123:
thanks.  verified the last push works.   here's a suggestion for the README.md 


https://gist.github.com/salrashid123/6e3c2eaa0575d6c18632931cbf8cb496

src/auth/impersonated.ts
```diff
@@ -0,0 +1,163 @@
...
+   *  final access_token. If set, the sequence of identities must have
+   * "Service Account Token Creator" capability granted to the preceding
+   * identity. For example, if set to [serviceAccountB, serviceAccountC],
+   * the sourceCredential must have the Token Creator role on serviceAccountB.
+   * serviceAccountB must have the Token Creator on serviceAccountC.
+   * Finally, C must have Token Creator on target_principal.
+   * If left unset, sourceCredential must have that role on targetPrincipal.
+   * @param targetScopes scopes to request during the authorization grant.
+   * @param lifetime number of seconds the delegated credential should be
+   * valid for (up to 3600).
```
silvolu, line 73: Lifetime is actually up to 43,200 if the service account is added to an organization policy that includes the constraints/iam.allowServiceAccountCredentialLifetimeExtension list constraint (
https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials#sa-credentials-oauth)

src/auth/impersonated.ts
```diff
@@ -0,0 +1,163 @@
...
+    super(options);
+    this.credentials = {
+      expiry_date: 1,
+      refresh_token: 'impersonated-placeholder',
+    };
+    this.sourceClient = options.sourceClient || new OAuth2Client();
+    this.targetPrincipal = options.targetPrincipal || '';
+    this.delegates = options.delegates || [];
+    this.targetScopes = options.targetScopes || [];
+    this.lifetime = options.lifetime || 3600;
```
silvolu, line 86: 3600 it's good as default value

test/test.impersonated.ts
```diff
@@ -0,0 +1,220 @@
...
+import {CredentialRequest} from '../src/auth/credentials';
+
+const PEM_PATH = './test/fixtures/private.pem';
+
+nock.disableNetConnect();
+
+const url = 'http://example.com';
+
+function createGTokenMock(body: CredentialRequest) {
+  return nock('https://www.googleapis.com')
```
silvolu, line 30: Nit: we're trying to centralize the different token exchange endpoints to 'https://oauth2.googleapis.com/token', even though this is a mock, it might be good for new code to just show that one.

silvolu (approved):
One comment on lifetime to address and one nit comment in the tests, overall looks good!

test/test.impersonated.ts
bcoe: We would need to change the underlying `gtoken` library to stop using the `/v4` endpoint.

In practice, I think we'll often be using self-signed JWTs for many of the cases where we would have used gtoken in the past. Regardless, I've created a tracking ticket based on this comment:

https://github.com/googleapis/node-gtoken/issues/388

bcoe:
CC: @bojeil-google

.readme-partials.yaml
```diff
@@ -606,3 +607,73 @@ body: |-
...
+      sourceClient: client,
+      targetPrincipal: "impersonated-account@projectID.iam.gserviceaccount.com",
+      lifetime: 30,
+      delegates: [],
+      targetScopes: ["https://www.googleapis.com/auth/cloud-platform"]
+    });
+
+    // Get impersonated credentials:
+    const authHeaders = await targetClient.getRequestHeaders();
+    // Do something with `authHeaders["Authorization"]`.
```
JustinBeckwith, line 645: Is there a reason we'd use an array syntax to access the value vs just saying `.Authorization`?

JustinBeckwith: Also the double quotes pokes me in the eyes, reminding me we don't test this code really

.readme-partials.yaml
```diff
@@ -606,3 +607,73 @@ body: |-
...
+
+    // Use impersonated credentials:
+    try {
+      const url = 'https://www.googleapis.com/storage/v1/b?project=anotherProjectID'
+      const resp = targetClient.request({ url });
+      for (var k in resp.data.items) {
```
JustinBeckwith, line 651: - Use `const` instead of `var`
- I don't think `for in` will actually work here, will it?  I thought that enumerated the properties of the object?

.readme-partials.yaml
```diff
@@ -606,3 +607,73 @@ body: |-
...
+        console.log(`${resp.data.items[k]['name']}`);
```
JustinBeckwith, line 652: No need to wrap this in a `${}`, you can directly log the statement

.readme-partials.yaml
```diff
@@ -606,3 +607,73 @@ body: |-
...
+      }
+    } catch (err) {
+      console.error('Unable to list buckets: ' + error);
```
JustinBeckwith, line 655: We generally avoid try/catches like this, where we just console.error - we let it throw like you typically would do in production

src/auth/impersonated.ts
```diff
@@ -0,0 +1,165 @@
...
+  RequestMetadataResponse,
+} from './oauth2client';
+
+export interface ImpersonatedOptions extends RefreshOptions {
+  sourceClient?: OAuth2Client;
+  targetPrincipal?: string;
+  targetScopes?: string[];
+  delegates?: string[];
+  lifetime?: number | 3600;
+  endpoint?: string | 'https://iamcredentials.googleapis.com';
```
JustinBeckwith, line 32: That is a weird type, since `string` includes the static value.

src/auth/impersonated.ts
```diff
@@ -0,0 +1,165 @@
...
+import {
+  GetTokenResponse,
+  Headers,
+  OAuth2Client,
+  RefreshOptions,
+  RequestMetadataResponse,
+} from './oauth2client';
+
+export interface ImpersonatedOptions extends RefreshOptions {
+  sourceClient?: OAuth2Client;
```
JustinBeckwith, line 27: For each of these properties, can we get a jsdoc annotation that describes what they are, and how they work?

src/auth/impersonated.ts
```diff
@@ -0,0 +1,165 @@
...
+   * Refreshes the access token.
+   * @param refreshToken Unused parameter
+   */
+  protected async refreshToken(
+    refreshToken?: string | null
+  ): Promise<GetTokenResponse> {
+    if (this.isTokenExpiring()) {
+      try {
+        await this.sourceClient.getAccessToken();
+        const name = 'projects/-/serviceAccounts/' + this.targetPrincipal;
```
JustinBeckwith, line 102: Please use string interpolation

src/auth/impersonated.ts
```diff
@@ -0,0 +1,165 @@
...
+        const u = `${this.endpoint}/v1/${name}:generateAccessToken`;
```
JustinBeckwith, line 103: Lets use more descriptive variable names so we know what these mean

src/auth/impersonated.ts
```diff
@@ -0,0 +1,165 @@
...
+          delegates: this.delegates,
+          scope: this.targetScopes,
+          lifetime: this.lifetime + 's',
+        };
+        const res = await this.sourceClient.request({
+          url: u,
+          data: body,
+          method: 'POST',
+        });
+        const tokenResponse = res.data as TokenResponse;
```
JustinBeckwith, line 114: Instead of doing the cast here, you can include it as a generic argument to `.request<TokenResponse>` above.

src/auth/impersonated.ts
```diff
@@ -0,0 +1,165 @@
...
+          Date.parse(tokenResponse.expireTime) / 1000;
+        return {
+          tokens: this.credentials,
+          res,
+        };
+      } catch (error) {
+        const status = error?.response?.data?.error?.status;
+        const message = error?.response?.data?.error?.message;
+        if (status && message) {
+          throw new Error(`${status}: unable to impersonate: ${message}`);
```
JustinBeckwith, line 126: Both of these will swallow the original exception, and hide the call stack.  Instead of throwing a new error, I would suggest crafting this string, and then prepending it to the message of the existing error:

```js
error.message = `${status}: unable to impersonate: ${message} \n\n ${err.message}`
throw error;
```

Here and in the else statement.

src/auth/impersonated.ts
```diff
@@ -0,0 +1,165 @@
...
+        } else {
+          throw new Error(`unable to impersonate: ${error}`);
+        }
+      }
+    }
+    return {tokens: this.credentials, res: null};
+  }
+
+  protected requestAsync<T>(
```
JustinBeckwith, line 135: Do we need to override this method if we just call the base class implementation?

src/auth/impersonated.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /**
+   * Get a non-expired access token, after refreshing if necessary.
+   *
+   * @param url The URI being authorized.
+   * @returns An object that includes the authorization header.
+   */
+  async getRequestHeaders(): Promise<Headers> {
+    const res = await this.getRequestMetadataAsync();
```
JustinBeckwith, line 149: Kinda the same question here as above.  Do we need this to override a base class method?

test/test.impersonated.ts
```diff
@@ -0,0 +1,220 @@
...
+      sourceClient: jwt,
+      targetPrincipal: 'target@project.iam.gserviceaccount.com',
+      lifetime: 30,
+      delegates: [],
+      targetScopes: ['https://www.googleapis.com/auth/cloud-platform'],
+    });
+    impersonated.credentials.access_token = 'initial-access-token';
+    impersonated.credentials.expiry_date = Date.now() - 10000;
+    let err: null | Error = null;
+    try {
```
JustinBeckwith, line 126: A more consistent pattern here is to use `assert.rejects`, and to drop the try/catch/finally.

JustinBeckwith (changes requested):

.readme-partials.yaml
JustinBeckwith: - Use a better variable name than `k`

src/auth/googleauth.ts
bcoe: I linked against this, and it seemed fine.

src/auth/impersonated.ts
bcoe: I was wondering the same when I looked at this, let me test.

.readme-partials.yaml
```diff
@@ -606,3 +607,69 @@ body: |-
...
+
+    // Get impersonated credentials:
+    const authHeaders = await targetClient.getRequestHeaders();
+    // Do something with `authHeaders.Authorization`.
+
+    // Use impersonated credentials:
+    const url = 'https://www.googleapis.com/storage/v1/b?project=anotherProjectID'
+    const resp = targetClient.request({ url });
+    for (const bucket of resp.data.items) {
+      console.log(resp.data.items.bucket.name);
```
JustinBeckwith, line 651: Should this be `bucket.name`?

JustinBeckwith (approved):

src/auth/impersonated.ts
```diff
@@ -0,0 +1,144 @@
...
+   * the sourceCredential must have the Token Creator role on serviceAccountB.
+   * serviceAccountB must have the Token Creator on serviceAccountC.
+   * Finally, C must have Token Creator on target_principal.
+   * If left unset, sourceCredential must have that role on targetPrincipal.
+   * @param targetScopes scopes to request during the authorization grant.
+   * @param lifetime number of seconds the delegated credential should be
+   * valid for up to 3600 seconds by default, or 43,200 seconds by
+   * extending the token's lifetime, see:
+   * https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials#sa-credentials-oauth
+   * @param endpoint api endpoint override.
```
bojeil-google, line 87: I think this should be the `ImpersonatedOptions` param

src/auth/impersonated.ts
```diff
@@ -0,0 +1,144 @@
...
+        lifetime: this.lifetime + 's',
+      };
+      const res = await this.sourceClient.request<TokenResponse>({
+        url: u,
+        data: body,
+        method: 'POST',
+      });
+      const tokenResponse = res.data;
+      this.credentials.access_token = tokenResponse.accessToken;
+      this.credentials.expiry_date =
```
bojeil-google, line 126: I think we use milliseconds for `expiry_date` in the rest of the repo.

src/index.ts
```diff
@@ -27,6 +27,7 @@ export {IAMAuth, RequestMetadata} from './auth/iam';
 export {IdTokenClient, IdTokenProvider} from './auth/idtokenclient';
 export {Claims, JWTAccess} from './auth/jwtaccess';
 export {JWT, JWTOptions} from './auth/jwtclient';
+export {Impersonated, ImpersonatedOptions} from './auth/impersonated';
```
bojeil-google, line 30: You may want to add a test for `Impersonated`:
https://github.com/googleapis/google-auth-library-nodejs/blob/3e0c4c92014f2b7b4543f28ad352f0d135efd08f/test/test.index.ts#L31

test/test.impersonated.ts
```diff
@@ -0,0 +1,305 @@
...
+interface ImpersonatedCredentialRequest {
+  delegates: string[];
+  scope: string[];
+  lifetime: string;
+}
+
+describe('impersonated', () => {
+  afterEach(() => {
+    nock.cleanAll();
+  });
```
bojeil-google, line 44: Suggest adding a new line after this and in between tests.

test/test.impersonated.ts
```diff
@@ -0,0 +1,305 @@
...
+            assert.deepStrictEqual(body.delegates, []);
+            assert.deepStrictEqual(body.scope, [
+              'https://www.googleapis.com/auth/cloud-platform',
+            ]);
+            return true;
+          }
+        )
+        .reply(200, {
+          accessToken: 'qwerty345',
+          expiryDate: tomorrow.toISOString(),
```
bojeil-google, line 67: I think this should be `expireTime`?

test/test.impersonated.ts
```diff
@@ -0,0 +1,305 @@
...
```
bojeil-google, line 111: I think this should be `expireTime`?

test/test.impersonated.ts
```diff
@@ -0,0 +1,305 @@
...
+      nock('https://iamcredentials.googleapis.com')
+        .post(
+          '/v1/projects/-/serviceAccounts/target@project.iam.gserviceaccount.com:generateAccessToken',
+          () => {
+            return true;
+          }
+        )
+        .reply(200, {
+          accessToken: 'qwerty345',
+          expiryDate: tomorrow.toISOString(),
```
bojeil-google, line 154: I think this should be `expireTime`?

test/test.impersonated.ts
```diff
@@ -0,0 +1,305 @@
...
+      nock('https://iamcredentials.googleapis.com')
+        .post(
+          '/v1/projects/-/serviceAccounts/target@project.iam.gserviceaccount.com:generateAccessToken',
+          () => {
+            return true;
+          }
+        )
+        .reply(200, {
+          accessToken: 'qwerty456',
+          expiryDate: tomorrow.toISOString(),
```
bojeil-google, line 165: I think this should be `expireTime`?

test/test.impersonated.ts
```diff
@@ -0,0 +1,305 @@
...
+            assert.deepStrictEqual(body.delegates, []);
+            assert.deepStrictEqual(body.scope, [
+              'https://www.googleapis.com/auth/cloud-platform',
+            ]);
+            return true;
+          }
+        )
+        .reply(200, {
+          accessToken: 'qwerty345',
+          expiryDate: tomorrow.toISOString(),
```
bojeil-google, line 282: I think this should be `expireTime`?

test/test.impersonated.ts
```diff
@@ -0,0 +1,305 @@
...
+    );
+    const impersonated = new Impersonated({
+      sourceClient: jwt,
+      targetPrincipal: 'target@project.iam.gserviceaccount.com',
+      lifetime: 30,
+      delegates: [],
+      targetScopes: ['https://www.googleapis.com/auth/cloud-platform'],
+    });
+    await impersonated.request({url});
+    assert.strictEqual(impersonated.credentials.access_token, 'qwerty345');
```
bojeil-google, line 85: Worth checking `impersonated.credentials.expiry_date` for the expected expiration to make sure it is set correctly (here and elsewhere).

test/test.impersonated.ts
```diff
@@ -0,0 +1,305 @@
...
+      lifetime: 30,
+      delegates: [],
+      targetScopes: ['https://www.googleapis.com/auth/cloud-platform'],
+    });
+    impersonated.credentials.access_token = 'initial-access-token';
+    impersonated.credentials.expiry_date = Date.now() - 10000;
+    await assert.rejects(impersonated.request({url}), /NOT_FOUND/);
+    scopes.forEach(s => s.done());
+  });
+  it('handles errors without context', async () => {
```
bojeil-google, line 229: Should you also add one more test when the underlying `sourceClient` fails to generate a token?

.readme-partials.yaml
```diff
@@ -606,3 +607,69 @@ body: |-
...
 
   A complete example can be found in [`samples/verifyIdToken-iap.js`](https://github.com/googleapis/google-auth-library-nodejs/blob/master/samples/verifyIdToken-iap.js).
+
+  ## Impersonated Credentials Client
+
+  Google Cloud Impersonated credentials used for [Creating short-lived service account credentials](https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials).
+
+  Provides authentication for applications where local credentials impersonates a remote service account using [IAM Credentials API](https://cloud.google.com/iam/docs/reference/credentials/rest).
+
+  An Impersonated Credentials Client is instantiated with a `sourceClient`. This
```
bojeil-google, line 617: Worth documenting that the `sourceClient` should have one of the scopes:
- https://www.googleapis.com/auth/iam
- https://www.googleapis.com/auth/cloud-platform

bojeil-google (changes requested):
Thanks for putting this together! Change looks good. My comments are mostly minor issues.

.readme-partials.yaml
```diff
@@ -606,3 +607,69 @@ body: |-
...
+    const url = 'https://www.googleapis.com/storage/v1/b?project=anotherProjectID'
+    const resp = targetClient.request({ url });
+    for (const bucket of resp.data.items) {
+      console.log(bucket.name);
+    }
+
+    // Use impersonated credentials with google-cloud client library
+    // Note: this works only with certain cloud client libraries utilizing gRPC 
+    //    e.g., SecretManager, KMS, AIPlatform
+    // will not currently work with libraries using REST, e.g., Storage, Compute
```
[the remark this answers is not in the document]
salrashid123: @bojeil-google ref  https://github.com/googleapis/google-auth-library-nodejs/issues/1210

bojeil-google:
Hey @bcoe thanks for addressing the comments. I noticed some comments may have been accidentally missed. Perhaps you forgot to commit those changes?
This includes: adding space between tests, 2021 copyright, using `AuthClient` instead of `OAuth2Client` for source `AuthClient`, updating `ImpersonatedOptions` param description and adding a test for `Impersonated` in `test.index.ts`.

This may be relevant but when I try to "show changes since my last review", I get an error message:

We went looking everywhere, but couldn’t find those commits.
Sometimes commits can disappear after a force-push. Head back to the latest changes here.

bcoe:
@bojeil-google the push had just failed.

bojeil-google (approved):

bcoe:
@bojeil-google thank you for the thorough review, this is much better than it would have been if landed earlier (also it identified a bug with refreshing token logic).

PR #1212, opened 2021-07-17
chore(serverless): add missing var to snippet block

ace-n:
Fixes b/185015090


bcoe:
@ace-n what's the behavior of fetching an ID Token if no target audience is provided?

JustinBeckwith (commented):
Looks like we need an `npm run fix` here?

bcoe (approved):

PR #1213, opened 2021-07-20
feat: Implement support for exposing refreshHandler callback to handle token refresh.

xil222:
Design Doc: [link](https://docs.google.com/document/d/1MJb9SH7cgGQYTUb6cShtMoOuaU3b05XtU4BVijGVok0/edit?resourcekey=0-K6R_URgisKyatWB6cXPK5g#heading=h.rjoi9m8cpb7k)
For this PR, developer has to provide a callback function returns access_token and expire_time both mandatory.
@bojeil-google @bcoe @lsirac Your feedback is highly appreciated!

There're 8 test cases:
getRequestHeader():
1. expired token, calling refreshHandler
2. token not expired, return authorized header with current token
3. no credentials in client, calling refreshHandler

request():
1. one with calling refreshHandler callback case
2. no credentials in client, calling refreshHandler

getAccessToken():
1. call refreshHandler when token expires and no refresh token
2. no credentials in client, calling refreshHandler

RefreshHandler returns no access token:
1. define refresh handler callback with empty access token, rejected when trying to refresh in getAccessToken()

Modify other error message and description in tests due to addition of refreshHandler logic.
Modify other test cases in request() to add another mock for 2nd successful request.

Note:
Since the logic of throwing error is in [refreshTokenNoCache](https://github.com/googleapis/google-auth-library-nodejs/blob/506bf9921d9a6101cd5977b241274b46604692d8/src/auth/oauth2client.ts#L682) I didn’t add "no refreshHandler callback related information" to the error message in the test [case](https://github.com/googleapis/google-auth-library-nodejs/blob/506bf9921d9a6101cd5977b241274b46604692d8/test/test.oauth2.ts#L1429)

### DIFF
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -298,6 +298,11 @@ export interface GenerateAuthUrlOpts {
   code_challenge?: string;
 }
 
+export interface DownscopedAccessTokenResponse {
+  access_token: string | null;
+  expiry_date: number | null;
+}
+
 export interface GetTokenCallback {
   (
     err: GaxiosError | null,
@@ -749,7 +754,13 @@ export class OAuth2Client extends AuthClient {
       !this.credentials.access_token || this.isTokenExpiring();
     if (shouldRefresh) {
       if (!this.credentials.refresh_token) {
-        throw new Error('No refresh token is set.');
+        const downscopedCreds = await this.refreshHandler();
+        if (downscopedCreds) {
+          this.setCredentials(downscopedCreds);
+          return {token: this.credentials.access_token};
+        } else {
+          throw new Error('No refresh token is set.');
+        }
       }
 
       const r = await this.refreshAccessTokenAsync();
@@ -789,6 +800,15 @@ export class OAuth2Client extends AuthClient {
       thisCreds.token_type = thisCreds.token_type || 'Bearer';
       const headers = {
         Authorization: thisCreds.token_type + ' ' + thisCreds.access_token,
+      };
+      return {headers: this.addSharedMetadataHeaders(headers)};
+    }
+
+    const downscopedCreds = await this.refreshHandler();
+    if (downscopedCreds) {
+      this.setCredentials(downscopedCreds);
+      const headers = {
+        Authorization: 'Bearer ' + this.credentials.access_token,
       };
       return {headers: this.addSharedMetadataHeaders(headers)};
     }
@@ -934,7 +954,7 @@ export class OAuth2Client extends AuthClient {
       const res = (e as GaxiosError).response;
       if (res) {
         const statusCode = res.status;
-        // Retry the request for metadata if the following criteria are true:
+        // Retry the request for metadata if the following criterias are true:
         // - We haven't already retried.  It only makes sense to retry once.
         // - The response was a 401 or a 403
         // - The request didn't send a readableStream
@@ -945,15 +965,40 @@ export class OAuth2Client extends AuthClient {
         //   fails on the first try because it's expired. Some developers may
         //   choose to enable forceRefreshOnFailure to mitigate time-related
         //   errors.
+        // Or the following criterias are true:
+        // - We haven't already retried.  It only makes sense to retry once.
+        // - The response was a 401 or a 403
+        // - The request didn't send a readableStream
+        // - No refresh_token was available
+        // - An access_token and a refreshHandler callback were available, but
+        //   either no expiry_date was available or the forceRefreshOnFailure
+        //   flag is set. The access_token fails on the first try because it's
+        //   expired. Some developers may choose to enable forceRefreshOnFailure
+        //   to mitigate time-related errors.
         const mayRequireRefresh =
           this.credentials &&
           this.credentials.access_token &&
           this.credentials.refresh_token &&
           (!this.credentials.expiry_date || this.forceRefreshOnFailure);
+        const mayRequireRefreshWithNoRefreshToken =
+          this.credentials &&
+          this.credentials.access_token &&
+          !this.credentials.refresh_token &&
+          (!this.credentials.expiry_date || this.forceRefreshOnFailure);
+        const downscopedCreds = await this.refreshHandler();
         const isReadableStream = res.config.data instanceof stream.Readable;
         const isAuthErr = statusCode === 401 || statusCode === 403;
         if (!retry && isAuthErr && !isReadableStream && mayRequireRefresh) {
           await this.refreshAccessTokenAsync();
+          return this.requestAsync<T>(opts, true);
+        } else if (
+          !retry &&
+          isAuthErr &&
+          !isReadableStream &&
+          mayRequireRefreshWithNoRefreshToken &&
+          downscopedCreds
+        ) {
+          this.setCredentials(downscopedCreds);
           return this.requestAsync<T>(opts, true);
         }
       }
@@ -1315,6 +1360,13 @@ export class OAuth2Client extends AuthClient {
     }
     return new LoginTicket(envelope, payload);
   }
+
+  /**
+   * Returns a Promise that resolves with DownscopedAccessTokenResponse
+   * type if refreshHandler() is defined.
+   * If there is no refresh handler callback set up, Promise resolves void.
+   */
+  async refreshHandler(): Promise<DownscopedAccessTokenResponse | void> {}
 
   /**
    * Returns true if a token is expired or will expire within
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -810,7 +810,7 @@ export class OAuth2Client extends AuthClient {
       !this.refreshHandler
     ) {
       throw new Error(
-        'No access, refresh token or API key or refresh handler callback is set.'
+        'No access, refresh token, API key or refresh handler callback is set.'
       );
     }
 
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -28,7 +28,6 @@ import {BodyResponseCallback} from '../transporters';
 import {AuthClient} from './authclient';
 import {CredentialRequest, Credentials} from './credentials';
 import {LoginTicket, TokenPayload} from './loginticket';
-import {getErrorFromOAuthErrorResponse} from './oauth2common';
 /**
  * The results from the `generateCodeVerifierAsync` method.  To learn more,
  * See the sample:
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -298,9 +298,13 @@ export interface GenerateAuthUrlOpts {
   code_challenge?: string;
 }
 
-export interface DownscopedAccessTokenResponse {
+export interface AccessTokenResponse {
   access_token: string | null;
   expiry_date: number | null;
+}
+
+export interface GetRefreshHandlerCallack {
+  (): Promise<AccessTokenResponse>;
 }
 
 export interface GetTokenCallback {
@@ -431,6 +435,8 @@ export class OAuth2Client extends AuthClient {
   eagerRefreshThresholdMillis: number;
 
   forceRefreshOnFailure: boolean;
+
+  refreshHandlerCallback?: GetRefreshHandlerCallack;
 
   /**
    * Handles OAuth2 flow for Google APIs.
@@ -754,10 +760,12 @@ export class OAuth2Client extends AuthClient {
       !this.credentials.access_token || this.isTokenExpiring();
     if (shouldRefresh) {
       if (!this.credentials.refresh_token) {
-        const downscopedCreds = await this.refreshHandler();
-        if (downscopedCreds) {
-          this.setCredentials(downscopedCreds);
-          return {token: this.credentials.access_token};
+        if (this.refreshHandlerCallback && this.credentials.access_token) {
+          const refreshedAccessToken = await this.refreshHandler();
+          if (refreshedAccessToken && refreshedAccessToken.access_token) {
+            this.setCredentials(refreshedAccessToken);
+            return {token: this.credentials.access_token};
+          }
         } else {
           throw new Error('No refresh token is set.');
         }
@@ -792,7 +800,12 @@ export class OAuth2Client extends AuthClient {
     url?: string | null
   ): Promise<RequestMetadataResponse> {
     const thisCreds = this.credentials;
-    if (!thisCreds.access_token && !thisCreds.refresh_token && !this.apiKey) {
+    if (
+      !thisCreds.access_token &&
+      !thisCreds.refresh_token &&
+      !this.apiKey &&
+      !this.refreshHandlerCallback
+    ) {
       throw new Error('No access, refresh token or API key is set.');
     }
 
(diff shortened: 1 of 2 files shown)


samples/package.json
```diff
@@ -14,7 +14,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@googleapis/iam": "^0.2.0",
-    "google-auth-library": "^7.3.0",
+    "google-auth-library": "^7.1.1",
```
bcoe, line 17: this should remain at `7.3.0`, if you need to test samples you can run `npm link ../` within the samples directory to link the current version of the library.

bcoe (changes requested):
One nit about the edit in `samples/package.json`, otherwise this looks reasonable to me.

src/auth/oauth2client.ts
```diff
@@ -749,7 +754,13 @@ export class OAuth2Client extends AuthClient {
       !this.credentials.access_token || this.isTokenExpiring();
     if (shouldRefresh) {
       if (!this.credentials.refresh_token) {
-        throw new Error('No refresh token is set.');
+        const downscopedCreds = await this.refreshHandler();
```
bojeil-google, line 757: Let's not use `downscopedCreds` as a name as this can be used in non-downscoping scenarios.

bojeil-google: Also since this is an optional property `refreshHandler`. You need to check it exists before running it.

src/auth/oauth2client.ts
```diff
@@ -749,7 +754,13 @@ export class OAuth2Client extends AuthClient {
...
+        if (downscopedCreds) {
```
bojeil-google, line 758: Based on the above, this is not sufficient as `downscopedCreds.access_token` may be null.
I suggest requiring these fields.

src/auth/oauth2client.ts
```diff
@@ -793,6 +804,15 @@ export class OAuth2Client extends AuthClient {
       return {headers: this.addSharedMetadataHeaders(headers)};
     }
 
+    const downscopedCreds = await this.refreshHandler();
```
bojeil-google, line 807: Let's not use `downscopedCreds` as a name as this can be used in non-downscoping scenarios.

bojeil-google: Also since this is an optional property `refreshHandler`. You need to check it exists before running it.

src/auth/oauth2client.ts
```diff
@@ -1316,6 +1361,13 @@ export class OAuth2Client extends AuthClient {
     return new LoginTicket(envelope, payload);
   }
 
+  /**
+   * Returns a Promise that resolves with DownscopedAccessTokenResponse
+   * type if refreshHandler() is defined.
+   * If there is no refresh handler callback set up, Promise resolves void.
+   */
+  async refreshHandler(): Promise<DownscopedAccessTokenResponse | void> {}
```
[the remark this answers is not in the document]
bojeil-google: Also this should reject when no tokens are returned instead of returning null. I think rejecting is better as the `refreshHandler` can provide a reason for no tokens being returned.

src/auth/oauth2client.ts
```diff
@@ -793,6 +804,15 @@ export class OAuth2Client extends AuthClient {
       return {headers: this.addSharedMetadataHeaders(headers)};
     }
```
bojeil-google, line 806: The check above:
`if (!thisCreds.access_token && !thisCreds.refresh_token && !this.apiKey) {`

It should be updated to included the `refreshHandler`:
`if (!thisCreds.access_token && !thisCreds.refresh_token && !this.apiKey && !this.refreshHandler) {`

src/auth/oauth2client.ts
```diff
@@ -945,16 +965,41 @@ export class OAuth2Client extends AuthClient {
...
           this.credentials &&
           this.credentials.access_token &&
           this.credentials.refresh_token &&
           (!this.credentials.expiry_date || this.forceRefreshOnFailure);
+        const mayRequireRefreshWithNoRefreshToken =
+          this.credentials &&
+          this.credentials.access_token &&
+          !this.credentials.refresh_token &&
+          (!this.credentials.expiry_date || this.forceRefreshOnFailure);
+        const downscopedCreds = await this.refreshHandler();
```
bojeil-google, line 988: Same issues as above. Also you are calling this too early and calling it even for the refresh token flow.

bojeil-google (changes requested):
Hey @xil222 I took a first look. I will continue next week. Thanks.

src/auth/oauth2client.ts
```diff
@@ -298,6 +298,11 @@ export interface GenerateAuthUrlOpts {
   code_challenge?: string;
 }
 
+export interface DownscopedAccessTokenResponse {
+  access_token: string | null;
+  expiry_date: number | null;
```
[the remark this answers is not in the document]
xil222: yes, we agreed on the former for now.

src/auth/oauth2client.ts
```diff
@@ -945,16 +981,44 @@ export class OAuth2Client extends AuthClient {
...
           this.credentials &&
           this.credentials.access_token &&
           this.credentials.refresh_token &&
           (!this.credentials.expiry_date || this.forceRefreshOnFailure);
+        const mayRequireRefreshWithNoRefreshToken =
+          this.credentials &&
+          this.credentials.access_token &&
+          !this.credentials.refresh_token &&
+          (!this.credentials.expiry_date || this.forceRefreshOnFailure);
+        // const downscopedCreds = await this.refreshHandler();
```
bcoe, line 1004: Seems odd that this is commented out.

src/auth/oauth2client.ts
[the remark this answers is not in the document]
xil222: ok, I add a check  for all 3 cases after calling refreshHandler and but calling setCredentials

src/auth/oauth2client.ts
xil222: ok, I add a check for all 3 cases after calling refreshHandler and but calling setCredentials.
But would this be redundant with the above comment that throw error if refreshHandler returns no token 
being returned.

src/auth/oauth2client.ts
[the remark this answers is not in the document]
xil222: I added a check for token if not rejected, but would it be redundant with the check for downscopedCreds.access_token not being null?

test/test.oauth2.ts
```diff
@@ -1348,5 +1400,58 @@ describe('oauth2', () => {
...
+
+    it('should refresh credential when refreshHandler is available in getAccessToken()', async () => {
+      const expectedRefreshedAccessToken = {
+        access_token: 'access_token',
+        expiry_date: 123456789,
+      };
+      client.refreshHandlerCallback = async () => {
+        return expectedRefreshedAccessToken;
+      };
+      client.setCredentials({
```
bojeil-google, line 1446: Are you testing expiration time detection? If so, please update the test description:
`it('should call refreshHandler on expiration and no refresh token available'`

test/test.oauth2.ts
```diff
@@ -1348,5 +1400,58 @@ describe('oauth2', () => {
...
+      assert.strictEqual(
+        refreshedAccessToken.expiry_date,
+        expectedRefreshedAccessToken.expiry_date
+      );
+    });
+
+    it('should refresh credential when refreshHandler is available in getAccessToken()', async () => {
+      const expectedRefreshedAccessToken = {
+        access_token: 'access_token',
+        expiry_date: 123456789,
```
bojeil-google, line 1441: Can you use a more realistic time (non-expired). e.g. `new Date().getTime() + 3600 * 1000`

test/test.oauth2.ts
```diff
@@ -1149,6 +1150,36 @@ describe('oauth2', () => {
...
+
+      it('should call refreshHandler and set credential if token expires in request()', async () => {
+        const scope = nock('http://example.com')
+          .get('/access')
+          .reply(code, {
+            error: {code, message: 'Invalid Credentials'},
+          });
+        const expectedRefreshedAccessToken = {
+          access_token: 'access_token',
+          expiry_date: 123456789,
```
bojeil-google, line 1162: Please use more realistic expiry dates in all these tests. This is expired.

test/test.oauth2.ts
```diff
@@ -1348,5 +1400,58 @@ describe('oauth2', () => {
         /No refresh token is set./
       );
     });
+
+    it('should throw error is refreshHandler does not return token', async () => {
```
bojeil-google, line 1404: should throw if the refreshHandler doesn't return a token

test/test.oauth2.ts
```diff
@@ -1348,5 +1400,58 @@ describe('oauth2', () => {
...
+      const expectedRefreshedAccessToken = {
+        access_token: '',
+        expiry_date: 123456789,
+      };
+      client.refreshHandlerCallback = async () => {
+        return expectedRefreshedAccessToken;
+      };
+      await assert.rejects(
+        client.refreshHandler(),
```
bojeil-google, line 1413: This should not be exposed. This should be the `getAccessToken()` call.

test/test.oauth2.ts
```diff
@@ -1348,5 +1400,58 @@ describe('oauth2', () => {
...
+        'There is no access token being returned.'
+      );
+    });
+
+    it('should call refreshHandler and set credential if providing refresh handler callback', async () => {
```
bojeil-google, line 1418: should call refreshHandler when no refresh token is available

test/test.oauth2.ts
```diff
@@ -1149,6 +1150,36 @@ describe('oauth2', () => {
           done();
         });
       });
+
```
bojeil-google, line 1153: Can you try to structure your tests as follows (whenever you can):
```
// Setup test block. You define the variables and stubs here.
// Add empty line.
// Run test block. You call the API you want to test.
// Add empty line.
// Assert test behavior block.
```

Sometimes you can group the last 2 blocks.

This will make it easy to know what you are actually testing.

test/test.oauth2.ts
```diff
@@ -1149,6 +1150,36 @@ describe('oauth2', () => {
...
+      it('should call refreshHandler and set credential if token expires in request()', async () => {
+        const scope = nock('http://example.com')
+          .get('/access')
```
bojeil-google, line 1156: How is the test passing below?
You don't call http://example.com/access

test/test.oauth2.ts
```diff
@@ -1149,6 +1150,36 @@ describe('oauth2', () => {
...
+          expiry_date: 123456789,
+        };
+        client.refreshHandlerCallback = async () => {
+          return expectedRefreshedAccessToken;
+        };
+        client.setCredentials({
+          access_token: 'initial-access-token',
+          expiry_date: new Date().getTime() - 1000,
+        });
+        client.request({url: 'http://example.com'}, () => {
```
bojeil-google, line 1171: This also returns a Promise. I suggest just testing the promise response.
You also need to assert that the authorization header is added.

test/test.oauth2.ts
```diff
@@ -1338,6 +1369,27 @@ describe('oauth2', () => {
       assert.deepStrictEqual(info.scopes, tokenInfo.scope.split(' '));
     });
 
+    it('should refresh request header when refreshHandler is available', async () => {
```
bojeil-google, line 1372: This test tests refresh handler flow when the existing token is expired.
Can you add a test when it is not expired? Also update the test description to clarify when a test is testing expiration flow or non-expiration flow.

bojeil-google (changes requested):

src/auth/oauth2client.ts
```diff
@@ -427,6 +437,8 @@ export class OAuth2Client extends AuthClient {
 
   forceRefreshOnFailure: boolean;
 
+  refreshHandlerCallback?: GetRefreshHandlerCallack;
```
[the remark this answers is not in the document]
xil222: @bcoe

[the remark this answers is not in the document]
bcoe: `refreshHandler` would be better I think, because callback tends to mean a function, whereas we accept a promise.

test/test.oauth2.ts
xil222: Honestly, I think the idea here is only checking for accessTokens being refreshed and there're enough test cases for requestHeaders already, I added an assert checking authorization header is being added.

In this file, there were other 19 client.request() being used without testing promise. I am not sure it is consistent with current code-style to declare a new responseType(for test purpose) and only testing two of client.request(). If @bcoe is okay with this, I'll test on the Promise.

xil222:
There're 5 test cases:
getRequestHeader():
1. expired token, calling refreshHandler
2. token not expired, return authorized header with current token

request():
1. one with calling refreshHandler callback case

getAccessToken():
1. call refreshHandler when token expires and no refresh token

RefreshHandler returns no access token:
1. define refresh handler callback with empty access token, rejected when trying to refresh in getAccessToken()

Modify other error message and description in tests due to addition of refreshHandler logic.

Note:
Since the logic of throwing error is in [refreshTokenNoCache](https://github.com/googleapis/google-auth-library-nodejs/blob/506bf9921d9a6101cd5977b241274b46604692d8/src/auth/oauth2client.ts#L682) I didn’t add "no refreshHandler callback related information" to the error message in the test [case](https://github.com/googleapis/google-auth-library-nodejs/blob/506bf9921d9a6101cd5977b241274b46604692d8/test/test.oauth2.ts#L1429)

test/test.oauth2.ts
```diff
@@ -1384,73 +1393,79 @@ describe('oauth2', () => {
...
+
       const requestMetaData = await client.getRequestHeaders(
         'http://example.com'
       );
+
       assert.deepStrictEqual(requestMetaData, expectedMetadata);
     });
 
-    it('should throw if tries to refresh but no refresh token is available', async () => {
+    it('should return authorization header on current access token if it has not expired', async () => {
```
bojeil-google, line 1404: nit: `should return cached authorization header on getRequestHeaders() if not expired`

test/test.oauth2.ts
```diff
@@ -1338,15 +1378,95 @@ describe('oauth2', () => {
...
+      };
+
+      const requestMetaData = await client.getRequestHeaders(
+        'http://example.com'
+      );
+
+      assert.deepStrictEqual(requestMetaData, expectedMetadata);
+    });
+
+    it('should throw if tries to refresh but no refreshHandler callback or refresh token is available', async () => {
```
bojeil-google, line 1420: nit: `should throw on getRequestHeaders() when no refreshHandler or refresh token is available`

test/test.oauth2.ts
```diff
@@ -1338,15 +1378,95 @@ describe('oauth2', () => {
...
+
+    it('should call refreshHandler in getAccessToken() on expiration and no refresh token available', async () => {
+      const expectedRefreshedAccessToken = {
+        access_token: 'access_token',
+        expiry_date: new Date().getTime() + 3600 * 1000,
+      };
+      client.refreshHandler = async () => {
+        return expectedRefreshedAccessToken;
+      };
+      client.setCredentials({
```
bojeil-google, line 1440: All your tests for `refreshHandler` rely on cached credentials being available. Please add tests when no credentials are available (don't use `setCredentials`)

Add tests for:
- `getAccessToken()`
- `request()`
- `getRequestHeaders()`

test/test.oauth2.ts
```diff
@@ -1338,15 +1378,95 @@ describe('oauth2', () => {
...
+        access_token: '',
+        expiry_date: new Date().getTime() + 3600 * 1000,
+      };
+      client.refreshHandler = async () => {
+        return expectedRefreshedAccessToken;
+      };
+      client.setCredentials({
+        access_token: 'initial-access-token',
+        expiry_date: new Date().getTime() - 1000,
+      });
```
bojeil-google, line 1464: Please minimize using `setCredentials()` to inject expired credentials unless you are specifically testing expiration flows. Please see my reasoning in the previous PRs I reviewed. The oauth2client should be able to be initialized with access tokens and solely with a `refreshHandler`.

```
const client = new OAuth2Client();
client.refreshHandler = async () => {...};
```

bojeil-google (changes requested):

bojeil-google (approved):
The tests look a lot better now. Glad we uncovered this. Thanks for fixing them.

bcoe (approved):

PR #1216, opened 2021-07-21
feat: Adds support for STS response not returning expires_in field.

xil222:
stsResponse doesnt return expire_in if subject_token isnt service account access token. 
Change the expire_in field to optional and add changes to downscopedclient and baseexternalclient.
bug: b/193924974
Your feedback is greatly appreciated! @bojeil-google @bcoe


src/auth/downscopedclient.ts
```diff
@@ -214,10 +214,15 @@ export class DownscopedClient extends AuthClient {
       this.credentialAccessBoundary
     );
 
+    const sourceCredExpireDate =
```
lsirac, line 217: Let's add a comment here: 

The STS endpoint will only return the expiration time for the downscoped access token if the
original access token represents a service account. The downscoped token's expiration time will always match the source credential expiration. When no expires_in is returned, we can copy the source credential's expiration time.

test/test.baseexternalclient.ts
```diff
@@ -1053,6 +1053,74 @@ describe('BaseExternalAccountClient', () => {
         scopes.forEach(scope => scope.done());
       });
 
+      it('should return credential with no expiry date if STS response does not return one', async () => {
```
lsirac, line 1056: I think expires_in is always returned for external account credentials, so this test might not be needed/confusing.

test/test.downscopedclient.ts
```diff
@@ -374,6 +374,126 @@ describe('DownscopedClient', () => {
...
+        stsSuccessfulResponseWithoutExpireInField.access_token
+      );
+      assert.strictEqual(
+        downscopedClient.credentials.expiry_date,
+        sourceCredentials.expiry_date
+      );
+      scope.done();
+    });
+
+    it('should has no expiry date if source cred has no expiry time and STS response does not return one', async () => {
```
lsirac, line 435: nit: have

test/test.downscopedclient.ts
```diff
@@ -374,6 +374,126 @@ describe('DownscopedClient', () => {
       );
       scope.done();
     });
+
+    it('should copy source cred expiry time if STS response does not return expiry time', async () => {
```
lsirac, line 378: This test confuses me. Based on the description, all we should be doing is (which you are doing):
1. Initialize a source credential with a certain expiration time
2. Mock STS to not return expires_in
3. Initialize DownscopedClient with the source credential
4. Call getAccessToken, and assert that the expiration matches the source credential

I don't think you need mock the clock so that the credentials you set expire. If you want to test that path, that can be a separate test?

test/test.baseexternalclient.ts
xil222: Yes, I think this usually not needed. However, since I changed logic by setting expire_in to optional,
I want to cover the line expireDate = null in refreshAccessTokenAsync() else case in baseexternalclient.ts.

src/auth/baseexternalclient.ts
```diff
@@ -378,10 +378,11 @@ export abstract class BaseExternalAccountClient extends AuthClient {
         stsResponse.access_token
       );
     } else {
```
bcoe, line 380: If you made this `else if (stsResponse.expires_in) {`

Then there's no need to cast to a `number`.

bcoe (changes requested):

test/test.downscopedclient.ts
```diff
@@ -411,9 +411,11 @@ describe('DownscopedClient', () => {
 
       clock.tick(1);
       const refreshedTokenResponse = await downscopedClient.getAccessToken();
+
+      const responseExpireIn = stsSuccessfulResponse.expires_in as number;
```
lsirac, line 415: nit: responseExpiresIn or expiresIn

lsirac (approved):

test/test.downscopedclient.ts
```diff
@@ -534,12 +536,100 @@ describe('DownscopedClient', () => {
...
       );
       await assert.rejects(downscopedClient.getAccessToken(), expectedError);
     });
+
+    it('should copy source cred expiry time if STS response does not return expiry time', async () => {
+      const now = new Date().getTime();
+      const credentials = {
+        access_token: 'ACCESS_TOKEN',
+        expiry_date: now + ONE_HOUR_IN_SECS * 1000,
+      };
```
bojeil-google, line 552: Why do you need to set these credentials and to call `client.setCredentials(credentials);`?
I suggest removing them for this test to just focus on the expiration time population logic.

Instead you can expose a public property on `TestAuthClient` to populate the expiration time on `client.credentials.expiry_date` when `getAccessToken()` is called.

test/test.downscopedclient.ts
```diff
@@ -534,12 +536,100 @@ describe('DownscopedClient', () => {
...
+          statusCode: 200,
+          response: stsSuccessfulResponseWithoutExpireInField,
+          request: {
+            grant_type: 'urn:ietf:params:oauth:grant-type:token-exchange',
+            requested_token_type:
+              'urn:ietf:params:oauth:token-type:access_token',
+            subject_token: 'subject_token_0',
+            subject_token_type: 'urn:ietf:params:oauth:token-type:access_token',
+            options:
+              testClientAccessBoundary &&
```
bojeil-google, line 613: You don't need this check.

test/test.downscopedclient.ts
```diff
@@ -534,12 +536,100 @@ describe('DownscopedClient', () => {
...
```
bojeil-google, line 569: You don't need this check.

test/test.downscopedclient.ts
```diff
@@ -534,12 +536,100 @@ describe('DownscopedClient', () => {
...
+      scope.done();
+    });
+
+    it('should have no expiry date if source cred has no expiry time and STS response does not return one', async () => {
+      const credentials = {
+        access_token: 'ACCESS_TOKEN',
+      };
+      const stsSuccessfulResponseWithoutExpireInField = Object.assign(
+        stsSuccessfulResponse,
+        {}
```
bojeil-google, line 599: Same issue here. The arguments need to be swapped.

test/test.downscopedclient.ts
```diff
@@ -534,12 +536,100 @@ describe('DownscopedClient', () => {
...
+        downscopedClient.credentials.expiry_date,
+        credentials.expiry_date
+      );
+      scope.done();
+    });
+
+    it('should have no expiry date if source cred has no expiry time and STS response does not return one', async () => {
+      const credentials = {
+        access_token: 'ACCESS_TOKEN',
+      };
```
bojeil-google, line 596: I think `TestAuthClient ` already doesn't populate the expiration, so this may not be needed.

bojeil-google (changes requested):
Thanks for putting this together. I mostly have suggestions and propose tweaks to the unit tests.

test/test.downscopedclient.ts
xil222: I think there're some confusions here, getAccessToken() is contained both in downscopedClient and authClient however functions differently, in testAuthClient, it is only supposed to return the access token / subject token, the latter one is returning an object including(token, expirationTime, res). I am mocking testAuthClient class extends on authClient class but not downscopedClient.

In implementation, the actual way of getting sourceCredExpireDate() is this.authClient.credentials?.expiry_date but not calling getAccessToken(), so calling client.setCredential(credentials) is the only and built-in way of setting expiration time.

test/test.downscopedclient.ts
```diff
@@ -49,6 +49,10 @@ class TestAuthClient extends AuthClient {
     throw new Error('Cannot get subject token.');
   }
 
+  set expirationTime(expirationTime: number) {
```
bojeil-google, line 52: I think this could also be cleared or set to null?
`set expirationTime(expirationTime: number|undefined|null) {`

test/test.baseexternalclient.ts
```diff
@@ -356,14 +357,34 @@ describe('BaseExternalAccountClient', () => {
         const client = new TestExternalAccountClient(
           externalAccountOptionsWithCreds
         );
+        // Listen to tokens events. On every event, push to list of
+        // emittedEvents.
+        client.on('tokens', tokens => {
```
bojeil-google, line 362: BTW, should we add a similar test for downscoped client to assert emitted event?

bojeil-google (changes requested):

bojeil-google (approved):

bcoe (approved):

PR #1218, opened 2021-07-23
samples: update TODO section

averikitsch:

### DIFF
--- a/samples/idtokens-serverless.js
+++ b/samples/idtokens-serverless.js
@@ -29,6 +29,7 @@ function main(
    * TODO(developer): Uncomment these variables before running the sample.
    */
   // const url = 'https://TARGET_URL';
+  // const targetAudience = null;
   const {GoogleAuth} = require('google-auth-library');
   const auth = new GoogleAuth();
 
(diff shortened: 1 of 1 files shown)


averikitsch:
`.origin` will strip the GCF url of the actual function route. I have tested with Cloud Run and it works either way.

averikitsch:
@googleapis/yoshi-nodejs PTAL

samples/idtokens-serverless.js
```diff
@@ -29,14 +29,15 @@ function main(
    * TODO(developer): Uncomment these variables before running the sample.
    */
   // const url = 'https://TARGET_URL';
+  // const targetAudience = null;
```
JustinBeckwith, line 32: As written, this code would fail because we're using a `const` instead of `let`, and trying to assign down on line 40.  Would another option here be moving `const {URL} = require('url');` with the other requires, and always doing the assignment of `const targetAudience = new URL(url)` on line 32?

JustinBeckwith (changes requested):
One issue with the snippet we should discuss - also, please use `docs:` or `samples:` - when you use `fix:`, it signals to the release tooling that this should cause a release.  This PR doesn't have any material impact on the library code, so we wouldn't want to do that.

JustinBeckwith (approved):

PR #1219, opened 2021-07-26
fix: Fixes issues in downscoped client implementation.

xil222:
Fixes issues like comments, STS endpoint, getAccessToken() return type, and add a couple of more tests in downscopedClient implementation.
@bojeil-google 

Hi Ben, this might need you sign [CLA](https://opensource.corp.google.com/github/prinfo?q=http://github.com/googleapis/google-auth-library-nodejs/pull/1219)

### DIFF
--- a/src/auth/downscopedclient.ts
+++ b/src/auth/downscopedclient.ts
@@ -37,9 +37,10 @@ const STS_SUBJECT_TOKEN_TYPE = 'urn:ietf:params:oauth:token-type:access_token';
 const STS_ACCESS_TOKEN_URL = 'https://sts.googleapis.com/v1/token';
 
 /**
- * Credential Access Boundary Rules upper bound number.
+ * The maximum number of access boundary rules a Credential Access Boundary
+ * can contain.
  */
-export const MAX_BOUNDARY_LIMIT = 10;
+export const MAX_ACCESS_BOUNDARY_RULES_COUNT = 10;
 
 /**
  * Offset to take into account network delays and server clock skews.
@@ -54,8 +55,8 @@ interface CredentialsWithResponse extends Credentials {
 }
 
 /**
- * Internal interface for tracking and returning Downscoped access token
- * expiration time.
+ * Internal interface for tracking and returning the Downscoped access token
+ * expiration time in epoch time (seconds).
  */
 interface DownscopedAccessTokenResponse extends GetAccessTokenResponse {
   expirationTime?: number | null;
@@ -88,12 +89,14 @@ interface AvailabilityCondition {
 }
 
 /**
- * Downscoped with Credential Access Boundary Client.
- * This is used to instantiate AuthClients for exchanging downscoped access
- * tokens using GCP access token fetched from source credentials such as service
- * account, user account or external account.
- * The process of exchanging GCP access token for downscoped token is based
- * on STS token endpoint.
+ * Defines a set of Google credentials that are downscoped from an existing set
+ * of Google OAuth2 credentials. This is useful to restrict the Identity and
+ * Access Management (IAM) permissions that a short-lived credential can use.
+ * The common pattern of usage is to have a token broker with elevated access
+ * generate these downscoped credentials from higher access source credentials
+ * and pass the downscoped short-lived access tokens to a token consumer via
+ * some secure authenticated channel for limited access to Google Cloud Storage
+ * resources.
  */
 export class DownscopedClient extends AuthClient {
   private cachedDownscopedAccessToken: CredentialsWithResponse | null;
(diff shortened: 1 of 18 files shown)


src/auth/downscopedclient.ts
```diff
@@ -74,35 +87,44 @@ interface AvailabilityCondition {
...
-   * used.
-   */
   private cachedDownscopedAccessToken: CredentialsWithResponse | null;
   private readonly stsCredential: sts.StsCredentials;
-  public readonly authClient: AuthClient;
-  public readonly credentialAccessBoundary: CredentialAccessBoundary;
   public readonly eagerRefreshThresholdMillis: number;
   public readonly forceRefreshOnFailure: boolean;
 
   constructor(
```
bojeil-google, line 104:
```
  /**
   * Instantiates a downscoped client object using the provided source
   * AuthClient and credential access boundary rules.
   * To downscope permissions of a source AuthClient, a Credential Access Boundary
   * that specifies which resources the new credential can access, as well as an
   * upper bound on the permissions that are available on each resource, has to be
   * defined. A downscoped client can then be instantiated using the source
   * AuthClient and the Credential Access Boundary.
   * @param authClient The source AuthClient to be downscoped based on the
   *    provided Credential Access Boundary rules.
   * @param credentialAccessBoundary The Credential Access Boundary which
   *    contains a list of access boundary rules. Each rule contains information on
   *    the resource that the rule applies to, the upper bound of the permissions
   *    that are available on that resource and an optional condition to further
   *    restrict permissions.
   * @param additionalOptions Optional additional behavior customization
   *   options. These currently customize expiration threshold time and
   *   whether to retry on 401/403 API request errors.
   */
```

src/auth/downscopedclient.ts
```diff
@@ -129,39 +148,41 @@ export class DownscopedClient extends AuthClient {
...
   /**
    * Provides a mechanism to inject Downscoped access tokens directly.
-   * When the provided credential expires, a new credential, using the
-   * external account options are retrieved.
-   * Notice DownscopedClient is the broker class mainly used for generate
-   * downscoped access tokens, it is unlikely we call this function in real
-   * use case.
-   * We implement to make this a helper function for testing all cases in getAccessToken().
+   * The expiry_date field is required for other clients dealing with
+   * token expiration checking and refreshing.
```
bojeil-google, line 152: The expiry_date field is required to facilitate determination of the token expiration which would make it easier for the token consumer to handle.

src/auth/downscopedclient.ts
```diff
@@ -129,39 +148,41 @@ export class DownscopedClient extends AuthClient {
...
    * @param credentials The Credentials object to set on the current client.
    */
   setCredentials(credentials: Credentials) {
+    if (!credentials.expiry_date) {
+      throw new Error(
+        'Credentials expiry date field has to be set up in downscopedClient.'
```
bojeil-google, line 158: The access token expiry_date field is missing in the provided credentials.

src/auth/downscopedclient.ts
```diff
@@ -129,39 +148,41 @@ export class DownscopedClient extends AuthClient {
...
     super.setCredentials(credentials);
     this.cachedDownscopedAccessToken = credentials;
   }
 
-  async getAccessToken(): Promise<GetAccessTokenResponse> {
+  async getAccessToken(): Promise<DownscopedAccessTokenResponse> {
     // If the cached access token is unavailable or expired, force refresh.
-    // The Downscoped access token will be returned in GetAccessTokenResponse format.
-    // If cached access token is unavailable or expired, force refresh.
+    // The Downscoped access token will be returned in DownscopedAccessTokenRespons
```
bojeil-google, line 167: typo

src/auth/downscopedclient.ts
```diff
@@ -207,7 +229,7 @@ export class DownscopedClient extends AuthClient {
       subjectTokenType: STS_SUBJECT_TOKEN_TYPE,
     };
 
-    // Exchange the source access token for a Downscoped access token.
+    // Exchange the GCP access token for a Downscoped access token.
```
bojeil-google, line 232: The returned downscoped access token is also a GCP access token. Maybe we can keep this "source AuthClient access token"

test/test.downscopedclient.ts
```diff
@@ -269,22 +296,40 @@ describe('DownscopedClient', () => {
     });
   });
 
-  describe('getAccessToken()', () => {
-    let sandbox: sinon.SinonSandbox;
-    before(() => {
-      const expectedAccessTokenResponse = {
-        token: 'subject_token',
+  describe('setCredential()', () => {
+    it('should throw error if no expire time is set in credential.', async () => {
```
bojeil-google, line 300: nit: no need for dot at the end of test description

test/test.downscopedclient.ts
```diff
@@ -269,22 +296,40 @@ describe('DownscopedClient', () => {
...
+        testClientAccessBoundary
+      );
+      assert.throws(() => {
+        downscopedClient.setCredentials(credentials);
+      }, expectedError);
     });
 
-    after(() => {
-      sandbox.restore();
+    it('should not throw error if expire time is set in credential.', async () => {
```
bojeil-google, line 316: nit: no need for dot at the end of test description

test/test.downscopedclient.ts
```diff
@@ -269,22 +296,40 @@ describe('DownscopedClient', () => {
...
+        access_token: 'DOWNSCOPED_CLIENT_ACCESS_TOKEN',
+        expiry_date: now + ONE_HOUR_IN_SECS * 1000,
+      };
+      const downscopedClient = new DownscopedClient(
+        client,
+        testClientAccessBoundary
+      );
+      assert.doesNotThrow(() => {
+        downscopedClient.setCredentials(credentials);
+      });
```
bojeil-google, line 328: Assert the credentials are set correctly.
```
const tokenResponse = await downscopedClient.getAccessToken();
assert.deepStrictEqual(tokenResponse.token, credentials.access_token);
assert.deepStrictEqual(tokenResponse.expirationTime, credentials.expiry_date);
```

test/test.downscopedclient.ts
```diff
@@ -298,6 +343,18 @@ describe('DownscopedClient', () => {
       downscopedClient.setCredentials(credentials);
```
bojeil-google, line 343: You can test the expiration is calculated correctly (with the offset) by using a mock clock and run tick with expirationDelta - 1 and confirm the cached token is returned.

test/test.downscopedclient.ts
```diff
@@ -336,44 +391,135 @@ describe('DownscopedClient', () => {
...
+        downscopedClient.credentials.access_token
+      );
+      assert.deepStrictEqual(
+        actualResponse.expirationTime,
+        downscopedClient.credentials.expiry_date
+      );
+      scope.done();
+    });
+
+    it('should throw error when getting subject token', async () => {
```
bojeil-google, line 514: should throw when the source AuthClient rejects on token request

bojeil-google (changes requested):
Thanks for making the requested changes! Most of the requested changes are minor and should be easy to resolve.

src/auth/downscopedclient.ts
```diff
@@ -48,10 +53,18 @@ interface CredentialsWithResponse extends Credentials {
   res?: GaxiosResponse | null;
 }
 
+/**
+ * Internal interface for tracking and returning Downscoped access token
```
lsirac, line 57: "the downscoped access token's expiration time"

src/auth/downscopedclient.ts
```diff
@@ -74,35 +87,44 @@ interface AvailabilityCondition {
...
+    ) {
       throw new Error('At least one access boundary rule needs to be defined.');
-    } else if (cab.accessBoundary.accessBoundaryRules.length > 10) {
-      throw new Error('Access boundary rule exceeds limit, max 10 allowed.');
+    } else if (
+      credentialAccessBoundary.accessBoundary.accessBoundaryRules.length >
+      MAX_BOUNDARY_LIMIT
+    ) {
+      throw new Error(
+        `Access boundary rule exceeds limit, max ${MAX_BOUNDARY_LIMIT} allowed.`
```
lsirac, line 121: nit:  "The provided access boundary has more than ${MAX_BOUNDARY_LIMIT} access boundary rules."

src/auth/downscopedclient.ts
lsirac, line 167: nit: DownscopedAccessTokenResponse

src/auth/downscopedclient.ts
xil222: this repo not always using 's, and used capped for Downscoped, I might have to follow the styles.

bcoe:
@xil222 please don't hesitate to reach out when @bojeil-google's comments are addressed.

src/auth/downscopedclient.ts
```diff
@@ -129,39 +172,42 @@ export class DownscopedClient extends AuthClient {
...
-   * Notice DownscopedClient is the broker class mainly used for generate
-   * downscoped access tokens, it is unlikely we call this function in real
-   * use case.
-   * We implement to make this a helper function for testing all cases in getAccessToken().
+   * The expiry_date field is required to facilitate determination of the token
+   * expiration which would make it easier for the token consumer to handle.
    * @param credentials The Credentials object to set on the current client.
    */
   setCredentials(credentials: Credentials) {
+    if (!credentials.expiry_date) {
```
lsirac, line 180: I don't know about this. It is technically possible to get a downscoped access token without an expiry (source creds no expiry, STS does not return one). Would be weird to return a downscoped access token without expiry, but then they can't inject it here. Though, I'm not sure why a developer would even use this method.

@bojeil-google thoughts?

bojeil-google: `setCredentials` is defined in the abstract super class `AuthClient`, so it will always be there.
I doubt it will be used. Most customers will just provide a source `AuthClient`. In that case, we should have the ability to get the underlying expiration time from the source `AuthClient`. We are trying to make sure if the customer provides it to us (though unlikely), they also provide the expiration time to make things easier and more determinable. I guess we can always relax it in the future if a customer has a good reason for it, though I highly doubt this will be used.

bojeil-google (approved):

lsirac (approved):

bcoe (approved):

PR #1227, opened 2021-08-05
feat: add GoogleAuth.sign() support to external account client

### LINKED ISSUE #1215: GoogleAuth.sign does not work with ExternalCredentials
#### Environment details

  - OS: Mac OS 11.4
  - Node.js version: 14.16.1
  - npm version: 6.14.12
  - `google-auth-library` version: 7.3.0

#### Steps to reproduce
```ts
const { GoogleAuth } = require('google-auth-library');

const auth = new GoogleAuth({ keyFilename: './clientLibraryConfig.json' });
auth.sign('abc').then(console.log); // fails
```
output:
```
(node:76183) UnhandledPromiseRejectionWarning: Error: Cannot sign data without `client_email`.
    at GoogleAuth.sign (/Users/calvinhuang/scale/gcptest/node_modules/google-auth-library/build/src/auth/googleauth.js:638:19)
```
Attempting to use workload identity federation with cloud storage signing does not work because of this, where clientLibraryConfig.json is the config retrieved from the workload identity pools page.

it is possible to work around with a bit of a hack:

```ts
const { GoogleAuth } = require('google-auth-library');

const auth = new GoogleAuth({ keyFilename: './clientLibraryConfig.json' });
(async () => {
  await auth.getClient();
  auth.jsonContent = { client_email: 'service-account-email' };
  await auth.sign('abc').then(console.log);
})()
```

bojeil-google:
External account credentials previously did not support signing blobs.
The implementation previously depended on service account keys or
the service account email in order to call IAMCredentials signBlob.

When service account impersonation is used with external account
credentials, we can get the impersonated service account email and
call the signBlob API with the generated access token, provided the
token has the `iam.serviceAccounts.signBlob` permission. This is
included in the "Service Account Token Creator" role.

Fixes https://github.com/googleapis/google-auth-library-nodejs/issues/1215

### DIFF
--- a/src/auth/baseexternalclient.ts
+++ b/src/auth/baseexternalclient.ts
@@ -177,6 +177,19 @@ export abstract class BaseExternalAccountClient extends AuthClient {
     this.projectNumber = this.getProjectNumber(this.audience);
   }
 
+  /** The service account email to be impersonated, if available. */
+  getServiceAccountEmail(): string | null {
+    if (this.serviceAccountImpersonationUrl) {
+      // Parse email from URL. The formal looks as follows:
+      // https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/name@project-id.iam.gserviceaccount.com:generateAccessToken
+      const matches = this.serviceAccountImpersonationUrl.match(
+        /serviceAccounts\/([^:]+):generateAccessToken$/
+      );
+      return (matches && matches[1]) || null;
+    }
+    return null;
+  }
+
   /**
    * Provides a mechanism to inject GCP access tokens directly.
    * When the provided credential expires, a new credential, using the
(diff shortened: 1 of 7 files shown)


silvolu (approved):

src/auth/baseexternalclient.ts
```diff
@@ -177,6 +177,19 @@ export abstract class BaseExternalAccountClient extends AuthClient {
     this.projectNumber = this.getProjectNumber(this.audience);
   }
 
+  /** The service account email to be impersonated, if available. */
+  getServiceAccountEmail(): string | null {
+    if (this.serviceAccountImpersonationUrl) {
+      // Parse email from URL. The formal looks as follows:
+      // https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/name@project-id.iam.gserviceaccount.com:generateAccessToken
+      const matches = this.serviceAccountImpersonationUrl.match(
+        /serviceAccounts\/([^:]+):generateAccessToken$/
```
bcoe, line 186: nit: I would be tempted to use a named capture group here, makes for slightly easier to follow code than `matches[1]`.

bcoe (approved):

src/auth/baseexternalclient.ts
bojeil-google: Done

PR #1232, opened 2021-08-12
feat: Make full implementation on DownscopedClient.

xil222:
This full implementation for downscoped client is just to satisfy requirement of the abstract class AuthClient, this is neither urgent and nor a blocker in terms of functionality since they're not expected to be used by token brokers.

Three components are added:
1. Add quota_project_id as an optional parameter in constructor.
2. Implement and test getRequestHeaders() API.
3. Implement and test request() API.
These implementations are mostly copied from BaseExternalClient.

### DIFF
--- a/src/auth/downscopedclient.ts
+++ b/src/auth/downscopedclient.ts
@@ -12,7 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {GaxiosOptions, GaxiosPromise, GaxiosResponse} from 'gaxios';
+import {
+  GaxiosError,
+  GaxiosOptions,
+  GaxiosPromise,
+  GaxiosResponse,
+} from 'gaxios';
+import * as stream from 'stream';
 
 import {BodyResponseCallback} from '../transporters';
 import {Credentials} from './credentials';
@@ -122,11 +128,14 @@ export class DownscopedClient extends AuthClient {
    * @param additionalOptions Optional additional behavior customization
    *   options. These currently customize expiration threshold time and
    *   whether to retry on 401/403 API request errors.
+   * @param quotaProjectId Optional quota project id for setting up in the
+   *   x-goog-user-project header.
    */
   constructor(
     private readonly authClient: AuthClient,
     private readonly credentialAccessBoundary: CredentialAccessBoundary,
-    additionalOptions?: RefreshOptions
+    additionalOptions?: RefreshOptions,
+    quotaProjectId?: string
   ) {
     super();
     // Check 1-10 Access Boundary Rules are defined within Credential Access
@@ -168,6 +177,7 @@ export class DownscopedClient extends AuthClient {
         .eagerRefreshThresholdMillis as number;
     }
     this.forceRefreshOnFailure = !!additionalOptions?.forceRefreshOnFailure;
+    this.quotaProjectId = quotaProjectId;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -57,6 +57,10 @@ export {
   BaseExternalAccountClient,
   BaseExternalAccountClientOptions,
 } from './auth/baseexternalclient';
+export {
+  CredentialAccessBoundary,
+  DownscopedClient,
+} from './auth/downscopedclient';
```
bojeil-google, line 63: Add test to make sure this is exported in `test.index.ts`

bojeil-google (changes requested):
Looks good. Just one test missing.

bojeil-google (approved):

test/test.downscopedclient.ts
```diff
@@ -683,30 +708,562 @@ describe('DownscopedClient', () => {
...
+              grant_type: 'urn:ietf:params:oauth:grant-type:token-exchange',
+              requested_token_type:
+                'urn:ietf:params:oauth:token-type:access_token',
+              subject_token: 'subject_token_1',
+              subject_token_type:
+                'urn:ietf:params:oauth:token-type:access_token',
+              options: JSON.stringify(testClientAccessBoundary),
+            },
+          },
+        ]),
```
bcoe, line 1239: I appreciate the thoroughness of this test, this would be a good pattern for us to use elsewhere.

src/auth/downscopedclient.ts
```diff
@@ -232,7 +246,65 @@ export class DownscopedClient extends AuthClient {
...
+   * Authenticates the provided HTTP request, processes it and resolves with the
+   * returned response.
+   * @param opts The HTTP request options.
+   * @param retry Whether the current attempt is a retry after a failed attempt.
+   * @return A promise that resolves with the successful response.
+   */
+  protected async requestAsync<T>(
+    opts: GaxiosOptions,
+    retry = false
+  ): Promise<GaxiosResponse<T>> {
```
bcoe, line 271: Is this logic identical to `BaseExternalClient`? If so I'm wondering if we could figure out a way to extract this into a mixin:

https://www.typescriptlang.org/docs/handbook/mixins.html

One option could be even doing this:

```js
protected async requestAsync() {
  BaseExternalClient.prototype.requestAsync.apply(opts, retry);
}
```

I haven't played with this approach in TypeScript only JavaScript, but seems like it should work.

bcoe (changes requested):
Can we figure out a way to make base external client and downscoped client share logic?

src/auth/downscopedclient.ts
xil222: I totally agree with the idea of reducing common code & duplication. 
The reason I'm copying the implementation is that @bojeil-google and I discussed this way is easier solution.
I just researched on both approaches, I think Mixin approach works for Javascript(I used that before), but for Typescript it looks different that could not be easily replicated in this scenario. The second approach doesn't work because some APIs like requestAsync are not public.

From my perspective, there is a good way to resolve the code duplication by creating a new class inheriting from AuthClient serving as parent class(I mentioned in design [doc](https://docs.google.com/document/d/1MJb9SH7cgGQYTUb6cShtMoOuaU3b05XtU4BVijGVok0/edit?resourcekey=0-K6R_URgisKyatWB6cXPK5g)), for both BaseExternalAccountOptions and DownscopedClient but that requires much effort which might affect timeline for the launch.

Is it possible for us to proceed with this short term solution and optimize on the replication in the future after the launch? (Since these copied APIs are there but not real use cases for developers)

bcoe: Yes, I'm open to this in the short term, but could you please open a tracking issue to loop back and fix 👍

xil222: http://b/198674048

bcoe (approved):

PR #1251, opened 2021-09-13
feat: add workforce config support.

xil222:
See go/workforce-pools-client-support.

Add support of work_force_user_config field for workforce pool and related logic in ExternalClient(calling constructor for IdentityPoolClient) BaseExternalClient(the parent class of IdentityPoolClient). 
The logic change is only related to refreshAccessToken() flow, since this API is non-public, we use getAccessToken() to test the flow instead. 

Add 16 tests:
ExternalClient:
1. fromJson() for IdentityPoolClient, throw an error is work_force_user_project provided by audience is not workforce audience.
2. fromJson() For IdentityPoolClient, return expected response is work_force_user_project provided and audience is workforce audience.

BaseExternalClient:
1. getAccessToken() Should apply basic auth on workforce configs with client auth provided(no impersonation).
2. getAccessToken() Should apply work_force_user_project on workforce configs without client auth(no impersonation).
3. getAccessToken() Should not throw if workforce audience and client auth but work_force_user_project not provided(no impersonation).
4. getAccessToken() Should not throw if workforce audience and no client auth but work_force_user_project provided( impersonation).
5. Constructor(), throw an error is work_force_user_project provided by audience is not workforce audience.
6. Constructor(), return expected response is work_force_user_project provided and audience is workforce audience.
7.  getProjectId(), should resolve with workforce projectID if no client auth not and workforce user project are defined.
8. getProjectId(), should not pass workforce user project if client auth is defined.

IdentityPoolClient:
1. getAccessToken() Should apply basic auth on workforce configs with client auth provided(no impersonation).
2. getAccessToken() Should apply work_force_user_project on workforce configs without client auth(no impersonation).
3. getAccessToken() Should not throw if workforce audience and client auth but work_force_user_project not provided(no impersonation).
4. getAccessToken() Should not throw if workforce audience and no client auth but work_force_user_project provided( impersonation).
5. Constructor(), throw an error is work_force_user_project provided by audience is not workforce audience.
6. Constructor(), return expected response is work_force_user_project provided and audience is workforce audience.

### DIFF
--- a/src/auth/baseexternalclient.ts
+++ b/src/auth/baseexternalclient.ts
@@ -59,7 +59,7 @@ export const CLOUD_RESOURCE_MANAGER =
   'https://cloudresourcemanager.googleapis.com/v1/projects/';
 /** The workforce audience pattern. */
 const WORKFORCE_AUDIENCE_PATTERN =
-  '//iam.googleapis.com/locations/[^/]+/workforcePools/.+/providers/.+';
+  '//iam.googleapis.com/locations/[^/]+/workforcePools/[^/]+/providers/.+';
 
 /**
  * Base external account credentials json interface.
@@ -310,10 +310,9 @@ export abstract class BaseExternalAccountClient extends AuthClient {
 
   /**
    * @return A promise that resolves with the project ID corresponding to the
-   *   current workload identity pool or current workforce pool. For workforce
-   *   pool credential, it returns the project ID corresponding to the
-   *   workforcePoolUserProject if client auth not determined. When not
-   *   determinable, this resolves with null.
+   *   current workload identity pool or current workforce pool if
+   *   determinable. For workforce pool credential, it returns the project ID
+   *   corresponding to the workforcePoolUserProject.
    *   This is introduced to match the current pattern of using the Auth
    *   library:
    *   const projectId = await auth.getProjectId();
@@ -325,22 +324,20 @@ export abstract class BaseExternalAccountClient extends AuthClient {
    *   https://cloud.google.com/resource-manager/reference/rest/v1/projects/get#authorization-scopes
    */
   async getProjectId(): Promise<string | null> {
+    const projectNumber = this.projectNumber || this.workforcePoolUserProject;
     if (this.projectId) {
       // Return previously determined project ID.
       return this.projectId;
-    } else if (this.projectNumber) {
+    } else if (projectNumber) {
       // Preferable not to use request() to avoid retrial policies.
       const headers = await this.getRequestHeaders();
       const response = await this.transporter.request<ProjectInfo>({
         headers,
-        url: `${CLOUD_RESOURCE_MANAGER}${this.projectNumber}`,
+        url: `${CLOUD_RESOURCE_MANAGER}${projectNumber}`,
         responseType: 'json',
       });
       this.projectId = response.data.projectId;
       return this.projectId;
-    } else if (!this.clientAuth && this.workforcePoolUserProject) {
-      // Return workforcePoolUserProject if client auth is not defined.
-      return this.workforcePoolUserProject;
     }
     return null;
   }
--- a/src/auth/baseexternalclient.ts
+++ b/src/auth/baseexternalclient.ts
@@ -57,6 +57,9 @@ export const EXTERNAL_ACCOUNT_TYPE = 'external_account';
 /** Cloud resource manager URL used to retrieve project information. */
 export const CLOUD_RESOURCE_MANAGER =
   'https://cloudresourcemanager.googleapis.com/v1/projects/';
+/** The workforce audience pattern. */
+const WORKFORCE_AUDIENCE_PATTERN =
+  '//iam.googleapis.com/locations/[^/]+/workforcePools/.+/providers/.+';
 
 /**
  * Base external account credentials json interface.
@@ -128,8 +131,8 @@ export abstract class BaseExternalAccountClient extends AuthClient {
   private readonly subjectTokenType: string;
   private readonly serviceAccountImpersonationUrl?: string;
   private readonly stsCredential: sts.StsCredentials;
-  private readonly clientAuth: ClientAuthentication | undefined;
-  private readonly workforcePoolUserProject: string | undefined;
+  private readonly clientAuth?: ClientAuthentication;
+  private readonly workforcePoolUserProject?: string;
   public projectId: string | null;
   public projectNumber: string | null;
   public readonly eagerRefreshThresholdMillis: number;
@@ -176,6 +179,16 @@ export abstract class BaseExternalAccountClient extends AuthClient {
     this.subjectTokenType = options.subject_token_type;
     this.quotaProjectId = options.quota_project_id;
     this.workforcePoolUserProject = options.workforce_pool_user_project;
+    const workforceAudiencePattern = new RegExp(WORKFORCE_AUDIENCE_PATTERN);
+    if (
+      this.workforcePoolUserProject &&
+      !this.audience.match(workforceAudiencePattern)
+    ) {
+      throw new Error(
+        'workforcePoolUserProject should not be set for non-workforce pool ' +
+          'credentials.'
+      );
+    }
     if (
       typeof options.service_account_impersonation_url !== 'undefined' &&
       !this.validateGoogleAPIsUrl(
@@ -297,8 +310,10 @@ export abstract class BaseExternalAccountClient extends AuthClient {
 
   /**
    * @return A promise that resolves with the project ID corresponding to the
-   *   current workload identity pool. When not determinable, this resolves with
-   *   null.
+   *   current workload identity pool or current workforce pool. For workforce
+   *   pool credential, it returns the project ID corresponding to the
+   *   workforcePoolUserProject if client auth not determined. When not
+   *   determinable, this resolves with null.
    *   This is introduced to match the current pattern of using the Auth
    *   library:
    *   const projectId = await auth.getProjectId();
@@ -323,6 +338,9 @@ export abstract class BaseExternalAccountClient extends AuthClient {
       });
       this.projectId = response.data.projectId;
       return this.projectId;
+    } else if (!this.clientAuth && this.workforcePoolUserProject) {
+      // Return workforcePoolUserProject if client auth is not defined.
+      return this.workforcePoolUserProject;
     }
     return null;
   }
@@ -408,20 +426,17 @@ export abstract class BaseExternalAccountClient extends AuthClient {
     };
 
     // Exchange the external credentials for a GCP access token.
-    let stsResponse;
     // Client auth is prioritized over passing the workforcePoolUserProject
     // parameter for STS token exchange.
-    if (!this.clientAuth && this.workforcePoolUserProject) {
-      stsResponse = await this.stsCredential.exchangeToken(
-        stsCredentialsOptions,
-        undefined,
-        {userProject: this.workforcePoolUserProject}
-      );
-    } else {
-      stsResponse = await this.stsCredential.exchangeToken(
-        stsCredentialsOptions
-      );
-    }
+    const additionalOptions =
+      !this.clientAuth && this.workforcePoolUserProject
+        ? {userProject: this.workforcePoolUserProject}
+        : undefined;
+    const stsResponse = await this.stsCredential.exchangeToken(
+      stsCredentialsOptions,
+      undefined,
+      additionalOptions
+    );
 
     if (this.serviceAccountImpersonationUrl) {
       this.cachedAccessToken = await this.getImpersonatedAccessToken(
--- a/src/auth/baseexternalclient.ts
+++ b/src/auth/baseexternalclient.ts
@@ -409,9 +409,8 @@ export abstract class BaseExternalAccountClient extends AuthClient {
 
     // Exchange the external credentials for a GCP access token.
     let stsResponse;
-    // If there is workforce pool user project but no client auth defined,
-    // pass user project as one additional parameter for STS token exchange.
-    // Otherwise, only pass stsCredentialsOptions for STS token exchange.
+    // Client auth is prioritized over passing the workforcePoolUserProject
+    // parameter for STS token exchange.
     if (!this.clientAuth && this.workforcePoolUserProject) {
       stsResponse = await this.stsCredential.exchangeToken(
         stsCredentialsOptions,
--- a/src/auth/baseexternalclient.ts
+++ b/src/auth/baseexternalclient.ts
@@ -409,14 +409,14 @@ export abstract class BaseExternalAccountClient extends AuthClient {
 
     // Exchange the external credentials for a GCP access token.
     let stsResponse;
-    // If there is work force pool user project but no client defined,
-    // pass user project config for STS token exchange.
-    // Otherwise, only pass stsCredentialsOptions for token exchange.
+    // If there is workforce pool user project but no client auth defined,
+    // pass user project as one additional parameter for STS token exchange.
+    // Otherwise, only pass stsCredentialsOptions for STS token exchange.
     if (!this.clientAuth && this.workforcePoolUserProject) {
       stsResponse = await this.stsCredential.exchangeToken(
         stsCredentialsOptions,
         undefined,
-        {user_project: this.workforcePoolUserProject}
+        {userProject: this.workforcePoolUserProject}
       );
     } else {
       stsResponse = await this.stsCredential.exchangeToken(
--- a/src/auth/baseexternalclient.ts
+++ b/src/auth/baseexternalclient.ts
@@ -71,6 +71,7 @@ export interface BaseExternalAccountClientOptions {
   client_id?: string;
   client_secret?: string;
   quota_project_id?: string;
+  workforce_pool_user_project?: string;
 }
 
 /**
@@ -127,6 +128,8 @@ export abstract class BaseExternalAccountClient extends AuthClient {
   private readonly subjectTokenType: string;
   private readonly serviceAccountImpersonationUrl?: string;
   private readonly stsCredential: sts.StsCredentials;
+  private readonly clientAuth: ClientAuthentication | undefined;
+  private readonly workforcePoolUserProject: string | undefined;
   public projectId: string | null;
   public projectNumber: string | null;
   public readonly eagerRefreshThresholdMillis: number;
@@ -152,7 +155,7 @@ export abstract class BaseExternalAccountClient extends AuthClient {
           `received "${options.type}"`
       );
     }
-    const clientAuth = options.client_id
+    this.clientAuth = options.client_id
       ? ({
           confidentialClientType: 'basic',
           clientId: options.client_id,
@@ -162,13 +165,17 @@ export abstract class BaseExternalAccountClient extends AuthClient {
     if (!this.validateGoogleAPIsUrl('sts', options.token_url)) {
       throw new Error(`"${options.token_url}" is not a valid token url.`);
     }
-    this.stsCredential = new sts.StsCredentials(options.token_url, clientAuth);
+    this.stsCredential = new sts.StsCredentials(
+      options.token_url,
+      this.clientAuth
+    );
     // Default OAuth scope. This could be overridden via public property.
     this.scopes = [DEFAULT_OAUTH_SCOPE];
     this.cachedAccessToken = null;
     this.audience = options.audience;
     this.subjectTokenType = options.subject_token_type;
     this.quotaProjectId = options.quota_project_id;
+    this.workforcePoolUserProject = options.workforce_pool_user_project;
     if (
       typeof options.service_account_impersonation_url !== 'undefined' &&
       !this.validateGoogleAPIsUrl(
@@ -401,9 +408,21 @@ export abstract class BaseExternalAccountClient extends AuthClient {
     };
 
     // Exchange the external credentials for a GCP access token.
-    const stsResponse = await this.stsCredential.exchangeToken(
-      stsCredentialsOptions
-    );
+    let stsResponse;
+    // If there is work force pool user project but no client defined,
+    // pass user project config for STS token exchange.
+    // Otherwise, only pass stsCredentialsOptions for token exchange.
+    if (!this.clientAuth && this.workforcePoolUserProject) {
+      stsResponse = await this.stsCredential.exchangeToken(
+        stsCredentialsOptions,
+        undefined,
+        {user_project: this.workforcePoolUserProject}
+      );
+    } else {
+      stsResponse = await this.stsCredential.exchangeToken(
+        stsCredentialsOptions
+      );
+    }
 
     if (this.serviceAccountImpersonationUrl) {
       this.cachedAccessToken = await this.getImpersonatedAccessToken(
--- a/src/auth/externalclient.ts
+++ b/src/auth/externalclient.ts
@@ -30,6 +30,10 @@ import {
 } from './identitypoolclient';
 import {AwsClient, AwsClientOptions} from './awsclient';
 
+/** The workforce audience pattern. */
+const WORKFORCE_AUDIENCE_PATTERN =
+  '//iam.googleapis.com/projects/.+/locations/.+/workforcePools/.+/providers/.+';
+
 export type ExternalAccountClientOptions =
   | IdentityPoolClientOptions
   | AwsClientOptions;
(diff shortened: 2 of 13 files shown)


src/auth/externalclient.ts
```diff
@@ -64,6 +70,18 @@ export class ExternalAccountClient {
...
+      options &&
+      options.workforce_pool_user_project &&
+      options.audience &&
+      !options.audience.match(workforceAudiencePattern)
+    ) {
+      throw new Error(
+        'The workforce_pool_user_project parameter should only be provided ' +
+          'for a Workforce Pool configuration.'
+      );
+    }
```
lsirac, line 84: This check should be done in the IdentityPoolClient constructor (otherwise they can initialize it themselves and we wouldn't catch this error).

src/auth/baseexternalclient.ts
```diff
@@ -401,9 +408,21 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+      stsResponse = await this.stsCredential.exchangeToken(
+        stsCredentialsOptions,
+        undefined,
+        {user_project: this.workforcePoolUserProject}
+      );
+    } else {
+      stsResponse = await this.stsCredential.exchangeToken(
+        stsCredentialsOptions
+      );
+    }
```
lsirac, line 425: Note that this change is specific to IdentityPoolClient.

test/test.awsclient.ts
```diff
@@ -67,6 +69,24 @@ describe('AwsClient', () => {
     },
     awsOptions
   );
+  const awsOptionsWithWorkforceUserProject = Object.assign(
```
lsirac, line 72: Revert this file. Changes are specific to IdentityPoolClient.

test/test.externalclient.ts
```diff
@@ -77,6 +77,32 @@ describe('ExternalAccountClient', () => {
...
+      '//iam.googleapis.com/projects/123/locations/global/workforcePools/providers/oidc',
+      '//iam.googleapis.com/projects/locations/global/workforcePools/providers/azure',
+      '//iam.googleapis.com/projects//locations/global/workforcePools/providers/azure',
+      '//iam.googleapis.com/projects/123/locations/global/workforcePools/pool/providers',
+      '//iam.googleapis.com/projects/123/locations/global/workforcePools/pool/providers/',
+      '//iam.googleapis.com/projects/123/locations//workforcepools/pool/providers/azure',
+      '//iam.googleapis.com/projects/123/locations/workforcepools/pool/providers/azure',
+    ];
+
+    const invalidWorkforceAwsClientAudiences = [
```
lsirac, line 93: remove

test/test.externalclient.ts
```diff
@@ -77,6 +77,32 @@ describe('ExternalAccountClient', () => {
...
+    const invalidWorkforceIdentityPoolClientAudiences = [
+      '//iam.googleapis.com/projects/123/locations/global/workloadIdentityPools/pool/providers/oidc',
+      '//iam.googleapis.com/projects/123/locations/global/workforcepools//providers/oidc',
+      '//iam.googleapis.com/projects/123/locations/global/workforcepools/pool/providers/oidc',
+      '//iam.googleapis.com/projects/123/locations/global/workforcePools/providers/oidc',
+      '//iam.googleapis.com/projects/locations/global/workforcePools/providers/azure',
+      '//iam.googleapis.com/projects//locations/global/workforcePools/providers/azure',
+      '//iam.googleapis.com/projects/123/locations/global/workforcePools/pool/providers',
+      '//iam.googleapis.com/projects/123/locations/global/workforcePools/pool/providers/',
+      '//iam.googleapis.com/projects/123/locations//workforcepools/pool/providers/azure',
```
lsirac, line 89: Same here: 

//iam.googleapis.com/projects/123/locations//workforce**P**ools/pool/providers/azure

test/test.externalclient.ts
```diff
@@ -77,6 +77,32 @@ describe('ExternalAccountClient', () => {
       forceRefreshOnFailure: true,
     };
 
+    const invalidWorkforceIdentityPoolClientAudiences = [
+      '//iam.googleapis.com/projects/123/locations/global/workloadIdentityPools/pool/providers/oidc',
+      '//iam.googleapis.com/projects/123/locations/global/workforcepools//providers/oidc',
+      '//iam.googleapis.com/projects/123/locations/global/workforcepools/pool/providers/oidc',
+      '//iam.googleapis.com/projects/123/locations/global/workforcePools/providers/oidc',
+      '//iam.googleapis.com/projects/locations/global/workforcePools/providers/azure',
```
lsirac, line 85: Same here:

//iam.googleapis.com/projects/locations/global/workforcePools/**pool**/providers/azure

test/test.externalclient.ts
```diff
@@ -77,6 +77,32 @@ describe('ExternalAccountClient', () => {
       forceRefreshOnFailure: true,
     };
 
+    const invalidWorkforceIdentityPoolClientAudiences = [
+      '//iam.googleapis.com/projects/123/locations/global/workloadIdentityPools/pool/providers/oidc',
+      '//iam.googleapis.com/projects/123/locations/global/workforcepools//providers/oidc',
```
lsirac, line 82: We want to change one thing, otherwise we don't know if it's failing for the same reason as a previous audience (no pool name provided) or not. 

//iam.googleapis.com/projects/123/locations/global/workforce**P**ools//providers/oidc

test/test.externalclient.ts
```diff
@@ -116,6 +142,103 @@ describe('ExternalAccountClient', () => {
...
+        );
+
+        assert.deepStrictEqual(
+          ExternalAccountClient.fromJSON(workforceFileSourcedOptions),
+          expectedClient
+        );
+      }
+    });
+
+    it('should return AwsClient with expected workforce configs', () => {
```
lsirac, line 174: remove this test

src/auth/baseexternalclient.ts
```diff
@@ -401,9 +408,21 @@ export abstract class BaseExternalAccountClient extends AuthClient {
     };
 
     // Exchange the external credentials for a GCP access token.
-    const stsResponse = await this.stsCredential.exchangeToken(
-      stsCredentialsOptions
-    );
+    let stsResponse;
+    // If there is work force pool user project but no client defined,
```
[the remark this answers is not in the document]
xil222: 'if possible?' That doesn't look formal to me.

src/auth/baseexternalclient.ts
xil222: I mentioned in the description. I cant change it here, since baseExternalClient is the parent class of IdentityPoolClient, the STS exchange this.stsCredential.exchangeToken in refreshAccessTokenAsync() is non-public and only implemented in BaseExternalClient.

src/auth/identitypoolclient.ts
```diff
@@ -100,6 +105,18 @@ export class IdentityPoolClient extends BaseExternalAccountClient {
         'Missing subject_token_field_name for JSON credential_source format'
       );
     }
+    const workforceAudiencePattern = new RegExp(WORKFORCE_AUDIENCE_PATTERN);
+    if (
+      options &&
```
lsirac, line 110: nit: don't need to check if options or options.audience. both required?

src/auth/baseexternalclient.ts
```diff
@@ -71,6 +71,7 @@ export interface BaseExternalAccountClientOptions {
   client_id?: string;
   client_secret?: string;
   quota_project_id?: string;
+  workforce_pool_user_project?: string;
```
lsirac, line 74: Since this is specific to IdentityPoolClient, I think this should be in IdentityPoolClientOptions

test/test.baseexternalclient.ts
```diff
@@ -82,6 +82,25 @@ describe('BaseExternalAccountClient', () => {
...
+    'urn:ietf:params:oauth:token-type:id_token';
+
+  const externalAccountOptionsWithClientAuthAndWorkforceUserProject =
+    Object.assign(
+      {
+        client_id: 'CLIENT_ID',
+        client_secret: 'SECRET',
+      },
+      externalAccountOptionsWorkforceUserProject
+    );
```
lsirac, line 103: nit: these are used once, why not inline?

test/test.baseexternalclient.ts
```diff
@@ -484,6 +503,80 @@ describe('BaseExternalAccountClient', () => {
         scope.done();
       });
 
+      it('should resolve with the expected response on workforce configs with client auth', async () => {
```
lsirac, line 506: This isn't very descriptive.

something like:

'should use client auth over passing the workforce user project when both are provided'

test/test.baseexternalclient.ts
```diff
@@ -484,6 +503,80 @@ describe('BaseExternalAccountClient', () => {
...
+        // Confirm raw GaxiosResponse appended to response.
+        assertGaxiosResponsePresent(actualResponse);
+        delete actualResponse.res;
+        assert.deepStrictEqual(actualResponse, {
+          token: stsSuccessfulResponse.access_token,
+        });
+        scope.done();
+      });
+
+      it('should resolve with the expected response on workforce configs without client auth', async () => {
```
lsirac, line 543: Same here. 

maybe 'should pass the workforce user project on workforce configs when client auth is not provided '

test/test.externalclient.ts
```diff
@@ -116,6 +129,57 @@ describe('ExternalAccountClient', () => {
...
+        '//iam.googleapis.com/projects/123/locations/global/workforcePools/workforcePools/providers/provider',
+        '//iam.googleapis.com/projects/workforcePool/locations/global/workforcePools/pool/providers/provider',
+        '//iam.googleapis.com/projects/projectId/locations/global/workforcePools/workloadPools/providers/oidc',
+      ];
+      const workforceFileSourcedOptions = Object.assign(
+        {},
+        fileSourcedOptions as any
+      );
+      workforceFileSourcedOptions.workforce_pool_user_project =
+        'work_force_pool_user_project';
```
lsirac, line 143: workforce_pool_user_project

test/test.externalclient.ts
```diff
@@ -116,6 +129,57 @@ describe('ExternalAccountClient', () => {
       );
     });
 
+    it('should return IdentityPoolClient with expected workforce configs', () => {
+      const validWorkforceIdentityPoolClientAudiences = [
+        '//iam.googleapis.com/projects/123/locations/global/workforcePools/workforcePools/providers/provider',
+        '//iam.googleapis.com/projects/workforcePool/locations/global/workforcePools/pool/providers/provider',
+        '//iam.googleapis.com/projects/projectId/locations/global/workforcePools/workloadPools/providers/oidc',
+      ];
+      const workforceFileSourcedOptions = Object.assign(
```
lsirac, line 138: Does this work?

const workforceFileSourcedOptions = Object.assign(
        {},
        fileSourcedOptions as any
        {workforce_pool_user_project: workforce_pool_user_project,

test/test.externalclient.ts
```diff
@@ -116,6 +129,57 @@ describe('ExternalAccountClient', () => {
...
+            invalidWorkforceIdentityPoolClientAudience;
+
+          assert.throws(() => {
+            return ExternalAccountClient.fromJSON(
+              workforceIdentityPoolClientInvalidOptions
+            );
+          });
+        });
+      }
+    );
```
lsirac, line 181: Let's move the loop inside the test, it'll be much cleaner:

```
it('should throw an error when an invalid workforce audience is provided with a workforce user project', () => {
  for (...)
    const workforceIdentityPoolClientInvalidOptions = Object.assign(
      {},
      fileSourcedOptions as any, 
      {audience: invalidWorkforceIdentityPoolClientAudience, workforce_pool_user_project: workforce_pool_user_project}
    );

    assert.throws(() => {
      return ExternalAccountClient.fromJSON(
        workforceIdentityPoolClientInvalidOptions
      );
    });
});
```

test/test.identitypoolclient.ts
```diff
@@ -63,6 +64,26 @@ describe('IdentityPoolClient', () => {
     },
     fileSourcedOptions
   );
+  const fileSourcedOptionsWithWorkforceUserProject = Object.assign(
```
lsirac, line 67: Same comments apply in this file

src/auth/baseexternalclient.ts
lsirac: This can definitely be changed, but I'll leave it up to @bojeil-google.

src/auth/baseexternalclient.ts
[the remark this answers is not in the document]
lsirac: This is just a code comment. I think keeping it short and simple is the way to go.

"Client auth is prioritized over passing the workforcePoolUserProject parameter for STS token exchange".

test/test.externalclient.ts
xil222: https://github.com/googleapis/google-auth-library-nodejs/pull/1229#discussion_r688270490

src/auth/baseexternalclient.ts
xil222: this one also @bojeil-google

test/test.baseexternalclient.ts
xil222: https://github.com/googleapis/google-auth-library-nodejs/blob/3ab1fc46610a090b8a207790b3a1d9f116b3c8a1/test/test.baseexternalclient.ts#L114,
just trying to follow code style, this is once used once also

test/test.externalclient.ts
```diff
@@ -116,6 +129,57 @@ describe('ExternalAccountClient', () => {
       );
     });
 
+    it('should return IdentityPoolClient with expected workforce configs', () => {
```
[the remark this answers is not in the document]
xil222: I guess its 'should return an IdentityPoolClient with a workforce config'

test/test.identitypoolclient.ts
```diff
@@ -232,6 +289,28 @@ describe('IdentityPoolClient', () => {
         return new IdentityPoolClient(urlSourcedOptionsNoHeaders);
       });
     });
+
+    it('should not throw on expected workforce configs', () => {
```
lsirac, line 293: 'should not throw on valid workforce configs'

src/auth/identitypoolclient.ts
```diff
@@ -100,6 +105,15 @@ export class IdentityPoolClient extends BaseExternalAccountClient {
...
     }
+    if (options && options.workforce_pool_user_project && options.audience) {
+      const workforceAudiencePattern = new RegExp(WORKFORCE_AUDIENCE_PATTERN);
+      if (!options.audience.match(workforceAudiencePattern)) {
+        throw new Error(
+          'The workforce_pool_user_project parameter should only be provided ' +
+            'for a Workforce Pool configuration.'
+        );
+      }
+    }
```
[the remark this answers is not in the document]
xil222: I checked above.

src/auth/identitypoolclient.ts
```diff
@@ -36,6 +36,10 @@ interface SubjectTokenJsonResponse {
   [key: string]: string;
 }
 
+/** The workforce audience pattern. */
+const WORKFORCE_AUDIENCE_PATTERN =
+  '//iam.googleapis.com/projects/.+/locations/.+/workforcePools/.+/providers/.+';
```
lsirac, line 41: //iam.googleapis.com/locations/.+/workforcePools/.+/providers/.+

xil222: I only saw this format in the old design [doc](https://docs.google.com/document/d/1PFz90avGF0w6Vb-8macSESmVj_du2MUgvW3YM9PDxtk/edit), not sure that was outdated, was that the reference?

src/auth/baseexternalclient.ts
```diff
@@ -401,9 +408,20 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+      stsResponse = await this.stsCredential.exchangeToken(
+        stsCredentialsOptions,
+        undefined,
+        {userProject: this.workforcePoolUserProject}
+      );
+    } else {
+      stsResponse = await this.stsCredential.exchangeToken(
+        stsCredentialsOptions
+      );
+    }
```
bojeil-google, line 424: You can probably do:
```
const additionalOptions = !this.clientAuth && this.workforcePoolUserProject ? {userProject: this.workforcePoolUserProject} : undefined;
const stsResponse = await this.stsCredential.exchangeToken(
  stsCredentialsOptions, undefined, additionalOptions
);
```
```

File: src/auth/identitypoolclient.ts [hunk 24]
```diff
@@ -100,6 +105,16 @@ export class IdentityPoolClient extends BaseExternalAccountClient {
         'Missing subject_token_field_name for JSON credential_source format'
       );
     }
+    const workforceAudiencePattern = new RegExp(WORKFORCE_AUDIENCE_PATTERN);
```
bojeil-google, line 108: Check above. This check should be in the base class.

test/test.baseexternalclient.ts
```diff
@@ -82,6 +82,24 @@ describe('BaseExternalAccountClient', () => {
     client_id: 'CLIENT_ID',
     client_secret: 'SECRET',
   };
+  const externalAccountOptionsWorkforceUserProject = Object.assign(
```
bojeil-google, line 85: I think we should test the following scenarios:
- Workforce audience + workforce_pool_user_project + no impersonation
- Workforce audience + workforce_pool_user_project + client auth + no impersonation => user project not passed
- Workforce audience + client auth + no impersonation => confirm no error thrown
- Workforce audience + workforce_pool_user_project + impersonation => confirm before impersonation userProject passed

It is important we test with no impersonation. Workforce pools are most commonly used without impersonation.

src/auth/identitypoolclient.ts
```diff
@@ -36,6 +36,10 @@ interface SubjectTokenJsonResponse {
   [key: string]: string;
 }
 
+/** The workforce audience pattern. */
+const WORKFORCE_AUDIENCE_PATTERN =
+  '//iam.googleapis.com/locations/.+/workforcePools/.+/providers/.+';
```
bojeil-google, line 41: I think `.+` in between locations and workforcePools is insufficient.
You can have something like:
`//iam.googleapis.com/locations/region/workloadIdentityPools/workforcePools/...`
I believe `region/workloadIdentityPools` would match `.+`

Granted workload identity pools always have projects so the above example is not realistic.

But I think you can make sure to exclude `/` from match.

test/test.identitypoolclient.ts
```diff
@@ -63,6 +64,26 @@ describe('IdentityPoolClient', () => {
     },
     fileSourcedOptions
   );
+  const fileSourcedOptionsWithWorkforceUserProject = Object.assign(
+    {},
```
bojeil-google, line 68: See test cases to cover above.

src/auth/baseexternalclient.ts
```diff
@@ -127,6 +128,8 @@ export abstract class BaseExternalAccountClient extends AuthClient {
   private readonly subjectTokenType: string;
   private readonly serviceAccountImpersonationUrl?: string;
   private readonly stsCredential: sts.StsCredentials;
+  private readonly clientAuth: ClientAuthentication | undefined;
+  private readonly workforcePoolUserProject: string | undefined;
```
bojeil-google, line 132: Note that `getProjectId()` should use this to determine project ID:
`async getProjectId(): Promise<string | null> {`

`GoogleAuth` relies on this.

See python implementation: https://github.com/googleapis/google-auth-library-python/pull/868/files

bojeil-google (changes requested):

src/auth/baseexternalclient.ts
bojeil-google: I think we should support this in base external account class (as @xil222 is already doing). See other comment for reasoning.

lsirac (approved):

bcoe:
@xil222 will take a look as soon as @bojeil-google has given 👍

xil222:
> @xil222 will take a look as soon as @bojeil-google has given 👍

ok, thanks

src/auth/baseexternalclient.ts
```diff
@@ -316,6 +338,9 @@ export abstract class BaseExternalAccountClient extends AuthClient {
       });
       this.projectId = response.data.projectId;
       return this.projectId;
+    } else if (!this.clientAuth && this.workforcePoolUserProject) {
+      // Return workforcePoolUserProject if client auth is not defined.
+      return this.workforcePoolUserProject;
```
bojeil-google, line 343: This won't work. The `workforce_pool_user_project` value can be a project ID or number. In fact, right now, it is only a project number. We have to call cloud resource manager. Please see the Python PR.

You can fix by doing something like:

```
    const projectNumber = this.projectNumber || this.workforcePoolUserProject;
    if (this.projectId) {
      // Return previously determined project ID.
      return this.projectId;
    } else if (projectNumber) {
      // Preferable not to use request() to avoid retrial policies.
      const headers = await this.getRequestHeaders();
      const response = await this.transporter.request<ProjectInfo>({
        headers,
        url: `${CLOUD_RESOURCE_MANAGER}${projectNumber}`,
        responseType: 'json',
      });
      this.projectId = response.data.projectId;
      return this.projectId;
    } ...
```

src/auth/baseexternalclient.ts
```diff
@@ -57,6 +57,9 @@ export const EXTERNAL_ACCOUNT_TYPE = 'external_account';
 /** Cloud resource manager URL used to retrieve project information. */
 export const CLOUD_RESOURCE_MANAGER =
   'https://cloudresourcemanager.googleapis.com/v1/projects/';
+/** The workforce audience pattern. */
+const WORKFORCE_AUDIENCE_PATTERN =
+  '//iam.googleapis.com/locations/[^/]+/workforcePools/.+/providers/.+';
```
bojeil-google, line 62: `'//iam.googleapis.com/locations/[^/]+/workforcePools/[^/]+/providers/.+'`

You can refine it further as the pool ID cannot contain slashes

test/test.baseexternalclient.ts
```diff
@@ -1307,6 +1529,55 @@ describe('BaseExternalAccountClient', () => {
         });
         scopes.forEach(scope => scope.done());
       });
+
+      it('should not throw if workforce user project is provided but client auth is not', async () => {
```
bojeil-google, line 1533: Suggest renaming: should still pass workforce user project when no client auth is used

test/test.identitypoolclient.ts
```diff
@@ -337,6 +423,176 @@ describe('IdentityPoolClient', () => {
...
+        // Confirm raw GaxiosResponse appended to response.
+        assertGaxiosResponsePresent(actualResponse);
+        delete actualResponse.res;
+        assert.deepStrictEqual(actualResponse, {
+          token: stsSuccessfulResponse.access_token,
+        });
+        scope.done();
+      });
+
+      it('should not throw if workforce user project and service account impersonation provided but client auth is not', async () => {
```
bojeil-google, line 542: Suggest rename test: should still pass workforce user project when impersonation and no client auth are used

bojeil-google (changes requested):

test/test.baseexternalclient.ts
```diff
@@ -423,29 +424,119 @@ describe('BaseExternalAccountClient', () => {
   });
 
   describe('getProjectId()', () => {
-    it('should return workforcePoolUserProject if client auth is not defined', async () => {
+    it('should resolve with workforce projectID if no client auth not and workforce user project are defined', async () => {
```
bojeil-google, line 427: I think this test is sufficient. You don't need the next one. You already cover the different scenarios above.
Rename to: should resolve for workforce pools when workforce_pool_user_project is provided

bojeil-google: Can you also add a test in `describe('projectNumber', () => {` to confirm projectNumber is null for workforce pools with workforce_pool_user_project?
```
    it('should return null for workforce pools with workforce_pool_user_project', () => {
      const options = Object.assign(
        {},
        externalAccountOptionsWorkforceUserProject
      );
      const client = new TestExternalAccountClient(options);

      assert(client.projectNumber === null);
    });
```

bojeil-google (changes requested):

bojeil-google (approved):
Changes look good from my end. Thanks a lot for putting this together!

bcoe (approved):

PR #1324, opened 2021-11-26
fix: await event emitting

### LINKED ISSUE #1323: await for emit('tokens')
**Is your feature request related to a problem? Please describe.**

when implementing a token cache, using the `on('tokens')` event is great. unfortunately, this does not work for async caches, because the `emit()` call is not awaited.

 **Describe the solution you'd like**

ideally, the code would `await` the call:

```
await this.emit('tokens', tokens);
```

here:

- https://github.com/googleapis/google-auth-library-nodejs/blob/5ed910513451c82e2551777a3e2212964799ef8e/src/auth/oauth2client.ts#L642
- https://github.com/googleapis/google-auth-library-nodejs/blob/5ed910513451c82e2551777a3e2212964799ef8e/src/auth/oauth2client.ts#L705

 **Describe alternatives you've considered**

we can hack around by overriding the internal methods that alter the refreshtoken, or proxy all methods and see if the `tokens` event was emitted. but this is a bit hack.

tripodsan:
Fixes #1323

### DIFF
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -639,7 +639,7 @@ export class OAuth2Client extends AuthClient {
       tokens.expiry_date = new Date().getTime() + res.data.expires_in * 1000;
       delete (tokens as CredentialRequest).expires_in;
     }
-    this.emit('tokens', tokens);
+    await this.emit('tokens', tokens);
     return {tokens, res};
   }
 
@@ -702,7 +702,7 @@ export class OAuth2Client extends AuthClient {
       tokens.expiry_date = new Date().getTime() + res.data.expires_in * 1000;
       delete (tokens as CredentialRequest).expires_in;
     }
-    this.emit('tokens', tokens);
+    await this.emit('tokens', tokens);
     return {tokens, res};
   }
 


bcoe:
@tripodsan I'm having a little bit of trouble understanding why you'd neet the `await` on `emit`, I believe `emit` is a synchronous action which will return either `true` or `false, see:

https://nodejs.org/api/events.html#emitteremiteventname-args

bcoe:
@tripodsan perhaps share some pseudo code that shows what you'd like to be able to do?

tripodsan:
> @tripodsan I'm having a little bit of trouble understanding why you'd neet the `await` on `emit`, I believe `emit` is a synchronous action which will return either `true` or `false, see:
> 
> https://nodejs.org/api/events.html#emitteremiteventname-args

yes, but `await` _works_ for sync and async functions. eg:

```
function helloSync() {
  console.log('hello');
}

async function helloASync() {
  await new Promise((resolve) => setTimeout(resolve, 100));
  console.log('hello');
}

async function run() {
  await helloSync();
  await helloASync();
}

run().catch((e) => console.error(e));
```

```
$ node index.js
hello
hello
```

tripodsan:
> @tripodsan perhaps share some pseudo code that shows what you'd like to be able to do?

I want to implement a token cache, that uses an async storage backend. something like:

```
const client = new google.auth.OAuth2(
  clientId,
  clientSecret,
  redirectUri,
);

client.on('tokens', async (tokens) => {
  await fs.writeFile('.auth.json', JSON.stringify(tokens), 'utf-8');
})
```

in my case, the backend is a DB that only supports async  I/O.

bcoe:
@tripodsan I believe this should work:

```js
const client = new google.auth.OAuth2(
  clientId,
  clientSecret,
  redirectUri,
);

client.on('tokens', async (tokens) => {
  await fs.writeFile('.auth.json', JSON.stringify(tokens), 'utf-8');
})
```

Without any changes to code, are you getting an exception?

tripodsan:
@bcoe yes, it works. but not correctly. since the event is processed async, it happens after the `await client.getAccessToken()`. which is a problem, if the auth happens multiple times.

also, writing unit tests is impossible, as the async code might happen after the test terminated.

and in AWS lambda, pending promises are not executed when the lambda function terminates.

tripodsan:
@bcoe btw: I think adobe updated the CCLA to include me. can you re-run the CLA check?

bcoe (approved):

bcoe (changes requested):
Can you provide a test that demonstrates the issue you're having?

I'm looking at the typing for event emitters in Node.js:

http://definitelytyped.org/docs/from--from/interfaces/nodejs.eventemitter.html

And the type returned is `boolean`, so if `this.emit` has the effect you're describing I think it's an unexpected side-effect.

A failing test would help.

tripodsan:
> And the type returned is `boolean`, so if `this.emit` has the effect you're describing I think it's an unexpected side-effect.

ah, your are right. it doesn't work with the await. I think I need to to a workaround, like this:

```ts
    it.only('should be possible to record tokens asynchronously', async () => {
      const scopes = mockExample();
      const fakeRefreshToken = 'refresh-token-placeholder';
      client.credentials = {refresh_token: fakeRefreshToken};

      let cachedRefreshToken;

      async function storeToken(tokens: any) {
        // assume the refresh token is stored in some async backend
        await new Promise(resolve => setTimeout(resolve, 5));
        cachedRefreshToken = tokens.refresh_token;
      }

      const pendingWrites: Promise<void>[] = [];
      client.on('tokens', tokens => {
        pendingWrites.push(storeToken(tokens));
      });

      await client.request({url: 'http://example.com'});

      await Promise.all(pendingWrites);

      assert.strictEqual(cachedRefreshToken, fakeRefreshToken);
    });
``

bcoe:
@tripodsan sorry I missed this response 👍 looks like you found an appraoch that works for you, awesome.

PR #1363, opened 2022-02-12
feat: Add AWS Session Token to Metadata Requests

sai-sunder-s:

Fixes #194738780 🦕

### DIFF
--- a/linkinator.config.json
+++ b/linkinator.config.json
@@ -4,7 +4,7 @@
     "https://codecov.io/gh/googleapis/",
     "www.googleapis.com",
     "img.shields.io",
-    "http://169.254.169.254/latest/api/token\""
+    "http://169.254.169.254/latest/api/token%22"
   ],
   "silent": true,
   "concurrency": 10
--- a/linkinator.config.json
+++ b/linkinator.config.json
@@ -4,7 +4,7 @@
     "https://codecov.io/gh/googleapis/",
     "www.googleapis.com",
     "img.shields.io",
-    "http://169.254.169.254/latest/api/token"
+    "http://169.254.169.254/latest/api/token\""
   ],
   "silent": true,
   "concurrency": 10
--- a/linkinator.config.json
+++ b/linkinator.config.json
@@ -3,7 +3,8 @@
   "skip": [
     "https://codecov.io/gh/googleapis/",
     "www.googleapis.com",
-    "img.shields.io"
+    "img.shields.io",
+    "http://169.254.169.254/latest/api/token"
   ],
   "silent": true,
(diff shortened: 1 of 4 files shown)


src/auth/awsclient.ts
```diff
@@ -197,10 +218,21 @@ export class AwsClient extends BaseExternalAccountClient {
     );
   }
 
+  private async getAwsSessionToken(): Promise<string> {
+    const opts: GaxiosOptions = {
+      url: this.awsSessionTokenUrl,
+      method: 'PUT',
+      responseType: 'text',
+      headers: {'x-aws-ec2-metadata-token-ttl-seconds': '21600'},
```
bojeil-google, line 226: I think `21600` seconds is a long time. Since we don't cache these responses and regenerate signed requests on demand, is there a need to make it this long?

src/auth/awsclient.ts
```diff
@@ -197,10 +218,21 @@ export class AwsClient extends BaseExternalAccountClient {
...
+    };
+    const response = await this.transporter.request<string>(opts);
+    return response.data;
+  }
+
   /**
    * @return A promise that resolves with the current AWS region.
    */
-  private async getAwsRegion(): Promise<string> {
+  private async getAwsRegion(headers: Headers): Promise<string> {
```
bojeil-google, line 235: Suggest adding the new headers field in the javadoc comments here and elsewhere for consistency.

bojeil-google (commented):
Thanks for putting this together! I just have a couple of suggestions.

src/auth/awsclient.ts
sai-sunder-s: Discussed offline and decided to go with 300 seconds (5 mins)

bojeil-google (approved):
No issues from my end. Thanks for putting this together!

lsirac (approved):

linkinator.config.json
```diff
@@ -3,7 +3,8 @@
   "skip": [
     "https://codecov.io/gh/googleapis/",
     "www.googleapis.com",
-    "img.shields.io"
+    "img.shields.io",
+    "http://169.254.169.254/latest/api/token%22"
```
bcoe, line 7:
```suggestion
    "img.shields.io"
```

This file is generated, so removing this from ignore -- I don't think we should need it unless we link to this URL somewhere in the ref docs.

bcoe (approved):

bcoe:
@[sai-sunder-s](https://github.com/sai-sunder-s) this looks good to me, appologies for the slow review.

bcoe (approved):

PR #1364, opened 2022-02-15
feat: Support instantiating `GoogleAuth` with an `AuthClient`

danielbankhead:
This will allow downstream clients to support features like Downscoped Tokens without breaking changes

### DIFF
--- a/samples/downscopedclient.js
+++ b/samples/downscopedclient.js
@@ -56,6 +56,7 @@ async function main() {
 
   const oauth2Client = new OAuth2Client();
   const googleAuth = new GoogleAuth({
+    auth: oauth2Client,
     scopes: 'https://www.googleapis.com/auth/cloud-platform',
   });
   const projectId = await googleAuth.getProjectId();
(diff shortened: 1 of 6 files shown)


samples/downscopedclient.js
```diff
@@ -75,31 +77,15 @@ async function main() {
...
-      },
-      authorizeRequest: async opts => {
-        opts = opts || {};
-        const url = opts.url || opts.uri;
-        const headers = await oauth2Client.getRequestHeaders(url);
-        opts.headers = Object.assign(opts.headers || {}, headers);
-        return opts;
-      },
-    },
+    authClient: new GoogleAuth({auth: oauth2Client}),
```
bcoe, line 80: Well, this is much nicer.

samples/test/downscoping-with-cab.test.js
```diff
@@ -30,9 +30,9 @@ const {promisify} = require('util');
 const exec = promisify(cp.exec);
 // Copy values from the output of samples/scripts/downscoping-with-cab-setup.js.
 // GCS bucket name.
-const bucketName = 'cab-int-bucket-z2zsauf4sj';
+const bucketName = 'cab-int-bucket-brd3qlsuok';
 // GCS object name.
-const objectName = 'cab-first-z2zsauf4sj.txt';
+const objectName = 'cab-first-"brd3qlsuok.txt';
```
bcoe, line 35: curious why we changed this object name.

bcoe (approved):
This implementation seems super clean, great work.

I'm always somewhat cautious with updates to the client library, perhaps you could link and run the storage sample/system tests against this branch? at which point let's ship.

samples/test/downscoping-with-cab.test.js
danielbankhead: Good question: The bucket and object previously didn’t exist - and the test has silently failed until this PR as we were asserting a string that existed in the error message (‘first’).

PR #1371, opened 2022-02-17
fix: Rename `auth` to `authClient` & Use Generics for `AuthClient`

danielbankhead:
Reduces type conflicts for downstream consumers.

NOTE: https://github.com/googleapis/nodejs-common/blob/6d1e19edc0765b4d5cc7a37ccc2fcb0e441e0815/src/service.ts#L65-L73 will need to be updated immediately after this change (as `GoogeAuth` != `AuthClient`)

### DIFF
--- a/samples/downscopedclient.js
+++ b/samples/downscopedclient.js
@@ -64,10 +64,10 @@ async function main() {
   const cabClient = new DownscopedClient(client, cab);
 
   // OAuth 2.0 Client
-  const oauth2Client = new OAuth2Client();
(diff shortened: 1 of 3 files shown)


bcoe (approved):

bcoe:
testing against gax, before we merge.

danielbankhead:
Also, https://github.com/googleapis/nodejs-common/pull/732 should be merged before this is merged

bcoe:
Great work making this non-breaking.

PR #1397, opened 2022-04-20
fix: Fixing Implementation of GoogleAuth.sign() for external account credentials

### LINKED ISSUE #1239: Mismatched signature when signing with impersonated credentials
A user of the cloud storage node library has reported an [issue](https://github.com/googleapis/nodejs-storage/issues/1550) when trying to sign URLs using Workload Identity Federation. The sign url feature of the library ultimately uses GoogleAuth.sign(), so in this case I think it should be resolving to the [recently implemented](https://github.com/googleapis/google-auth-library-nodejs/pull/1227/files) case for signing with impersonated credentials. Evidently, it *does* sign, it just produces a different signature than when the credentials and private key are used directly. Possibly some issue where the wrong email ends up getting used?

aeitzman:
Currently, creating signed storage URLs does not work for external account credentials because the storage library expects client_email to be returned from GoogleAuth.getCredentials(). Changing the logic so the same client email that is used to the sign the blob (extracted from the Service Account Impersonation URL) is returned from the getCredentials() call.

Fixes #1239  🦕

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -756,11 +756,11 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
   private async getCredentialsAsync(): Promise<CredentialBody> {
     const client = await this.getClient();
 
-    if (
-      client instanceof BaseExternalAccountClient &&
-      client.getServiceAccountEmail()
-    ) {
-      return {client_email: client.getServiceAccountEmail() as string};
+    if (client instanceof BaseExternalAccountClient) {
+      const serviceAccountEmail = client.getServiceAccountEmail();
+      if (serviceAccountEmail) {
+        return {client_email: serviceAccountEmail};
+      }
     }
 
     if (this.jsonContent) {
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -731,8 +731,8 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
    * The callback function handles a credential object that contains the
    * client_email and private_key (if exists).
    * getCredentials first checks if the client is using an external account and
-   * uses the service account email in place of client_email
-   * if that doesn't exist, it checks for these values from the user JSON.
+   * uses the service account email in place of client_email.
+   * If that doesn't exist, it checks for these values from the user JSON.
    * If the user JSON doesn't exist, and the environment is on GCE, it gets the
    * client_email from the cloud metadata server.
    * @param callback Callback that handles the credential object that contains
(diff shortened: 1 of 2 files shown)


aeitzman:
@lsirac

src/auth/googleauth.ts
```diff
@@ -752,7 +754,14 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
 
   private async getCredentialsAsync(): Promise<CredentialBody> {
-    await this.getClient();
+    const client = await this.getClient();
+
+    if (
+      client instanceof BaseExternalAccountClient &&
+      client.getServiceAccountEmail()
+    ) {
+      return {client_email: client.getServiceAccountEmail() as string};
```
bcoe, line 763: what if you check whether or not this is null in the `if`, then you don't need `as string`.

bcoe (approved):

test/test.googleauth.ts
```diff
@@ -2340,6 +2340,18 @@ describe('googleauth', () => {
         assert.deepStrictEqual(res.data, data);
         scopes.forEach(s => s.done());
       });
+
+      it('getCredentials() should return the service account email for external accounts', async () => {
```
bcoe, line 2344: Nit, You could do:

```
describe('getCredentials()', () => {
  it('should return the service account email for external accounts', async () => 
  {
        // Set up a mock to return path to a valid credentials file.
        const email = saEmail;
        const configWithImpersonation = createExternalAccountJSON();
        configWithImpersonation.service_account_impersonation_url =
          getServiceAccountImpersonationUrl();
        const auth = new GoogleAuth({credentials: configWithImpersonation});
        const body = await auth.getCredentials();
        assert.notStrictEqual(null, body);
        assert.strictEqual(email, body.client_email);
      });
});
```

So that tests read a bit better.

src/auth/googleauth.ts
```diff
@@ -730,8 +730,10 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
   /**
    * The callback function handles a credential object that contains the
    * client_email and private_key (if exists).
-   * getCredentials checks for these values from the user JSON at first.
-   * If it doesn't exist, and the environment is on GCE, it gets the
+   * getCredentials first checks if the client is using an external account and
+   * uses the service account email in place of client_email
```
lsirac, line 734: missing period

src/auth/googleauth.ts
```diff
@@ -730,8 +730,10 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
+   * if that doesn't exist, it checks for these values from the user JSON.
```
lsirac, line 735: If

lsirac (approved):

PR #1406, opened 2022-05-05
feat!: make transporter attribute type more generic

forty:
Hello,

currently transporter attributes are inferred or explicitly typed as `DefaultTransporter`, which is too bad since the `Transporter` type exist I assume to make room for custom implementation (which I use, but currently I have to either extends DefaultTransporter, or force TS to ignore the typing)


danielbankhead (approved):
Thanks! LGTM

danielbankhead:
As a note - I think this is a worth-while change, but would be backwards-incompatible as `Transporter` does not have `static USER_AGENT` and the `configure` method (which are public for consumers). Marking this PR for a future major release.

PR #1407, opened 2022-05-05
chore!: clean up the Transporter interface.

### LINKED ISSUE #1405: gtoken request does not use the transporter
#### Environment details

  - OS: Linux
  - Node.js version: 12
  - npm version: Using PNPM 6
  - `google-auth-library` version: 7.14.1

#### Steps to reproduce

This library uses the [gtoken](https://github.com/googleapis/node-gtoken) for JWT. The issue is that when using this library, the `transporter` is not used to do the gaxios request, and as a result, it doesn't work with custom `transporter` (in our case we use this to customize node http agent).

Given the very small size of gtoken (350 LoC), I think if would make sense to re-implement it here, using the `transporter` system. But maybe it would be better to also have such `transporter` mecanism in gtoken so that we could use it here.

forty:
Hello (again),

I took the opportunity that I was looking into it to clean up a bit the Transporter interface, as it seems to be used with Promises everywhere in the project.

Fixes #1405


bcoe:
@forty thanks for the contribution, we'll get this triaged soon.

danielbankhead (approved):
Thanks for your contribution! This looks great. Since the `Transporter` and `BodyResponseCallback<T>` interfaces are a part of the public API, this is considered breaking and will roll into the next major release.

PR #1419, opened 2022-06-01
feat!: Adding Pluggable Auth Support to ADC

aeitzman:
See go/pluggable-auth-design.
Adding classes required for supporting pluggable auth and some functionality.
Will add the implementation for running the executable and reading from a cached file in a later pull request.

### DIFF
--- a/src/auth/executableresponse.ts
+++ b/src/auth/executableresponse.ts
@@ -1,0 +1,166 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const SAML_SUBJECT_TOKEN_TYPE = 'urn:ietf:params:oauth:token-type:saml2';
+const OIDC_SUBJECT_TOKEN_TYPE1 = 'urn:ietf:params:oauth:token-type:id_token';
+const OIDC_SUBJECT_TOKEN_TYPE2 = 'urn:ietf:params:oauth:token-type:jwt';
+
+/**
+ * Interface defining JSON format of response from 3rd party executable run by
+ * pluggable auth client.
+ */
+export interface ExecutableResponseJson {
+  /**
+   * Version of the response, only version 1 currently supported.
+   * Always required.
+   */
+  version: number;
+  /**
+   * Whether the executable ran successfully. Always required.
+   */
+  success: boolean;
+  /**
+   * Epoch time for expiration of the token in seconds, required for successful
+   * responses.
+   */
+  expiration_time?: number;
+  /**
+   * Type of subject token in the response, currently supported values are:
+   * urn:ietf:params:oauth:token-type:saml2
+   * urn:ietf:params:oauth:token-type:id_token
+   * urn:ietf:params:oauth:token-type:jwt
+   */
+  token_type?: string;
+  /**
+   * Error code from executable, required when unsuccessful.
+   */
+  code?: string;
+  /**
+   * Error message from executable, required when unsuccessful.
+   */
+  message?: string;
+  /**
+   * ID token to be used as a subject token when token_type is id_token or jwt.
+   */
+  id_token?: string;
+  /**
+   * Response to be used as a subject token when token_type is saml2.
+   */
+  saml_response?: string;
+}
+
+/**
+ * Class defining a response from a 3rd party executable run by the pluggable
+ * auth client.
+ */
+export class ExecutableResponse {
+  readonly version: number;
+  readonly success: boolean;
+  readonly expirationTime?: number;
+  readonly tokenType?: string;
+  readonly errorCode?: string;
+  readonly errorMessage?: string;
+  readonly subjectToken?: string;
+
+  /**
+   * Instantiates an ExecutableResponse instance using the provided JSON object
+   * from the output of the executable.
+   * @param responseJson Response from a 3rd party executable, loaded from a
+   * run of the executable or a cached output file.
+   */
+  constructor(responseJson: ExecutableResponseJson) {
+    // Check that always required fields exist in the json response.
+    if (!responseJson.version) {
+      throw Error("Executable response must contain a 'version' field.");
+    }
+    if (responseJson.success === undefined) {
+      throw Error("Executable response must contain a 'success' field.");
+    }
+
+    this.version = responseJson.version;
+    this.success = responseJson.success;
+
+    // Check that fields required when response is successful exist.
+    if (this.success) {
+      if (!responseJson.expiration_time) {
+        throw Error(
+          "Executable response must contain an 'expiration_time' field when successful."
+        );
+      }
+      if (!responseJson.token_type) {
+        throw Error(
+          "Executable response must contain a 'token_type' field when successful."
+        );
+      }
+      this.expirationTime = responseJson.expiration_time;
+      this.tokenType = responseJson.token_type;
+
+      // Check that type specific fields exist based on token_type provided and
+      // set subjectToken.
+      if (this.tokenType === SAML_SUBJECT_TOKEN_TYPE) {
+        if (!responseJson.saml_response) {
+          throw Error(
+            `Executable response must contain a 'saml_response' field when token_type=${SAML_SUBJECT_TOKEN_TYPE}.`
+          );
+        }
+        this.subjectToken = responseJson.saml_response;
+      } else if (
+        this.tokenType === OIDC_SUBJECT_TOKEN_TYPE1 ||
+        this.tokenType === OIDC_SUBJECT_TOKEN_TYPE2
+      ) {
+        if (!responseJson.id_token) {
+          throw Error(
+            "Executable response must contain a 'id_token' field when " +
+              `token_type=${OIDC_SUBJECT_TOKEN_TYPE1} or ${OIDC_SUBJECT_TOKEN_TYPE2}.`
+          );
+        }
+        this.subjectToken = responseJson.id_token;
+      } else {
+        throw Error(
+          "Executable response must contain a 'token_type' field when successful " +
+            `and it must be one of ${OIDC_SUBJECT_TOKEN_TYPE1}, ${OIDC_SUBJECT_TOKEN_TYPE2}, or ${SAML_SUBJECT_TOKEN_TYPE}.`
+        );
+      }
+    } else {
+      // Check that fields required for unsuccessful response exist.
+      if (!responseJson.code || !responseJson.message) {
+        throw Error(
+          "Executable response must contain a 'code' and 'message' field when unsuccessful."
+        );
+      }
+      this.errorCode = responseJson.code;
+      this.errorMessage = responseJson.message;
+    }
+  }
+
+  /**
+   * @return a boolean representing if the response has a valid token. Returns
+   * true when the response was successful and the token is not expired.
+   */
+  isValid(): boolean {
+    return !this.isExpired() && this.success;
+  }
+
+  /**
+   * @return a boolean representing if the response is expired. Returns true if
+   * the expiration_time field was not provided or if the provided time has
+   * passed.
+   */
+  isExpired(): boolean {
+    return (
+      !this.expirationTime ||
+      this.expirationTime < Math.round(Date.now() / 1000)
+    );
+  }
+}
--- a/src/auth/executableresponse.ts
+++ b/src/auth/executableresponse.ts
@@ -105,7 +105,7 @@ export class ExecutableResponse {
       this.expirationTime = responseJson.expiration_time;
       this.tokenType = responseJson.token_type;
 
-      // Validate that token type and subject token value.
+      // Validate token type and subject token value.
       if (this.tokenType === SAML_SUBJECT_TOKEN_TYPE) {
         if (!responseJson.saml_response) {
           throw Error(
--- a/src/auth/executableresponse.ts
+++ b/src/auth/executableresponse.ts
@@ -31,31 +31,31 @@ export interface ExecutableResponseJson {
    */
   success: boolean;
   /**
-   * Epoch time for expiration of the token in seconds, required for successful
-   * responses.
+   * The epoch time for expiration of the token in seconds, required for
+   * successful responses.
    */
   expiration_time?: number;
   /**
-   * Type of subject token in the response, currently supported values are:
+   * The type of subject token in the response, currently supported values are:
    * urn:ietf:params:oauth:token-type:saml2
    * urn:ietf:params:oauth:token-type:id_token
    * urn:ietf:params:oauth:token-type:jwt
    */
   token_type?: string;
   /**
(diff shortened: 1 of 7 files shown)


src/auth/executableresponse.ts
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * Interface defining JSON format of response from 3rd party executable run by
+ * pluggable auth client.
+ */
+export interface ExecutableResponseJson {
+  /**
+   * Version of the response, only version 1 currently supported.
+   * Always required.
+   */
```
lsirac, line 27: // The version of the JSON response. Only version 1 is currently supported.

Same below. I *think* it should be // and not /**

src/auth/executableresponse.ts
```diff
@@ -0,0 +1,166 @@
...
+  id_token?: string;
+  /**
+   * Response to be used as a subject token when token_type is saml2.
+   */
+  saml_response?: string;
+}
+
+/**
+ * Class defining a response from a 3rd party executable run by the pluggable
```
lsirac, line 64: Defines the response of  a 3rd party executable run by the pluggable auth client.

src/auth/executableresponse.ts
```diff
@@ -0,0 +1,166 @@
...
+   * from the output of the executable.
+   * @param responseJson Response from a 3rd party executable, loaded from a
+   * run of the executable or a cached output file.
+   */
+  constructor(responseJson: ExecutableResponseJson) {
+    // Check that always required fields exist in the json response.
+    if (!responseJson.version) {
+      throw Error("Executable response must contain a 'version' field.");
+    }
+    if (responseJson.success === undefined) {
```
lsirac, line 87: if (!responseJson.success)?

src/auth/executableresponse.ts
```diff
@@ -0,0 +1,166 @@
...
+        throw Error(
+          "Executable response must contain a 'token_type' field when successful."
+        );
+      }
+      this.expirationTime = responseJson.expiration_time;
+      this.tokenType = responseJson.token_type;
+
+      // Check that type specific fields exist based on token_type provided and
+      // set subjectToken.
+      if (this.tokenType === SAML_SUBJECT_TOKEN_TYPE) {
```
lsirac, line 111: Maybe we can split this up to reduce nesting and make it easier to follow:

```
e.g.:

if (
  this.tokenType !== SAML_SUBJECT_TOKEN_TYPE1 ||
  this.tokenType !== OIDC_SUBJECT_TOKEN_TYPE1 ||
  this.tokenType !== OIDC_SUBJECT_TOKEN_TYPE2
) {
 throw Error(...);
} 

if (this.tokenType === SAML_SUBJECT_TOKEN_TYPE) {
  this.subjectToken = responseJson.saml_response;
} else {
  this.subjectToken = responseJson.id_token;
}

if (!this.subjectToken) throw...
```

src/auth/executableresponse.ts
```diff
@@ -0,0 +1,166 @@
...
+        }
+        this.subjectToken = responseJson.id_token;
+      } else {
+        throw Error(
+          "Executable response must contain a 'token_type' field when successful " +
+            `and it must be one of ${OIDC_SUBJECT_TOKEN_TYPE1}, ${OIDC_SUBJECT_TOKEN_TYPE2}, or ${SAML_SUBJECT_TOKEN_TYPE}.`
+        );
+      }
+    } else {
+      // Check that fields required for unsuccessful response exist.
```
lsirac, line 136: // Both code and message must be provided for unsuccessful responses.

src/auth/pluggableauthclient.ts
```diff
@@ -0,0 +1,253 @@
...
+  constructor(
+    options: PluggableAuthClientOptions,
+    additionalOptions?: RefreshOptions
+  ) {
+    super(options, additionalOptions);
+    this.command = options.credential_source.command;
+    if (!this.command) {
+      throw new Error('No valid Pluggable Auth "credential_source" provided.');
+    }
+    //Check if the provided timeout exists and if it is valid.
```
lsirac, line 126: // Check

src/auth/pluggableauthclient.ts
```diff
@@ -0,0 +1,253 @@
...
+      executableResponse = await this.retrieveCachedResponse();
+    }
+
+    // If no response from output file, call the executable
+    if (!executableResponse) {
+      // Set up environment map with required values for the executable
+      const envMap = new Map();
+      envMap.set('GOOGLE_EXTERNAL_ACCOUNT_AUDIENCE', this.audience);
+      envMap.set('GOOGLE_EXTERNAL_ACCOUNT_TOKEN_TYPE', this.subjectTokenType);
+      // Set to 0 for workloadIdentityFederation
```
lsirac, line 183: // Interactive mode is not supported.

src/auth/pluggableauthclient.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (process.env[GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES] !== '1') {
+      throw new Error(
+        'Pluggable Auth executables need to be explicitly allowed to run by ' +
+          'setting the GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES environment ' +
+          'Variable to 1.'
+      );
+    }
+
+    let executableResponse: ExecutableResponse | undefined;
+    // Try to get cached executable response from output file
```
lsirac, line 172: Missing period

lsirac (commented):
First quick pass

src/auth/executableresponse.ts
aeitzman: Success is a boolean so if the check is changed to that it will throw an error if success=false, we only want this error to throw if the success field wasn't provided at all.

src/auth/pluggableauthclient.ts
aeitzman: FYI, comment is the same for the java PR, (says set to 0 for workload identity federation but doesn't mention why) I agree that the additional comment is good but should add it to the other repo PR's as well.

src/auth/executableresponse.ts
aeitzman: is the // vs /** reasoning because the interface won't be exported externally or a different reason? I was a bit confused about what comment style to use in which contexts.

lsirac: I am not sure. I would consult the style guide

src/auth/executableresponse.ts
lsirac: You can still do that - just add an error for each. Comment still applies. Also the error doesn't have to be generic either, you can put 'saml_response' or 'id_token' but might as well have specific ones

src/auth/executableresponse.ts
aeitzman: From reading the docs, looks like the JSDoc style comments (/**) should be used for all top level exports from a module, so the interface + class here and the public properties. From that I should actually add comments for the properties in the ExecutableResponse class as well.

src/auth/executableresponse.ts
```diff
@@ -0,0 +1,195 @@
+// Copyright 2022 Google LLC
```
bcoe, line 1: nit: I would call this `executable-reponse.ts`, there's a chance some of our files don't follow this convention, but it tends to be how files are named in JS.

src/auth/executableresponse.ts
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * Instantiates an ExecutableResponse instance using the provided JSON object
+   * from the output of the executable.
+   * @param responseJson Response from a 3rd party executable, loaded from a
+   * run of the executable or a cached output file.
+   */
+  constructor(responseJson: ExecutableResponseJson) {
+    // Check that always required fields exist in the json response.
+    if (!responseJson.version) {
+      throw Error("Executable response must contain a 'version' field.");
```
bcoe, line 108: I think there would be value in creating typed errors, as it makes flow control easier.

```
try {

} catch (err) {
  if (err instanceof MissingVersion) {
     // do something.
  } else {
    throw err.
  }
}
```

We could consider introducing an `errors.ts` file if we find that we're using the same errors across files.

src/auth/externalclient.ts
```diff
@@ -29,10 +29,15 @@ import {
...
+import {
+  PluggableAuthClient,
+  PluggableAuthClientOptions,
+} from './pluggableauthclient';
 
 export type ExternalAccountClientOptions =
   | IdentityPoolClientOptions
-  | AwsClientOptions;
+  | AwsClientOptions
+  | PluggableAuthClientOptions;
```
bcoe, line 40: I believe this type change will mean we should treat the change as breaking, we can link against `googleapis-common` and `google-gax` to confirm.

IMO this isn't a big deal, we just take a major bump for the auth client, which allows us to intentionally upgrade auth in `google-gax`.

test/test.pluggableauthclient.ts
```diff
@@ -0,0 +1,129 @@
...
+  describe('Constructor', () => {
+    it('should throw when credential_source is missing command', () => {
+      const expectedError = new Error(
+        'No valid Pluggable Auth "credential_source" provided.'
+      );
+      const invalidCredentialSource = Object.assign(
+        {},
+        pluggableAuthCredentialSource
+      );
+      delete (invalidCredentialSource as any)['command'];
```
bcoe, line 49: nit: this is reasonable in a test, but you should add an eslint ignore line to quiet the linter.

bcoe (changes requested):
Left a few comments, most notably I think we should potentially consider this change breaking unless `PluggableAuthClientOptions` is a superset of the other client option types.

src/auth/executableresponse.ts
```diff
@@ -0,0 +1,195 @@
...
+  readonly subjectToken?: string;
+
+  /**
+   * Instantiates an ExecutableResponse instance using the provided JSON object
+   * from the output of the executable.
+   * @param responseJson Response from a 3rd party executable, loaded from a
+   * run of the executable or a cached output file.
+   */
+  constructor(responseJson: ExecutableResponseJson) {
+    // Check that always required fields exist in the json response.
```
lsirac, line 106: // Check that the required fields exist in the JSON response.

src/auth/executableresponse.ts
```diff
@@ -0,0 +1,195 @@
...
+
+      // Validate subject token.
+      if (this.tokenType === SAML_SUBJECT_TOKEN_TYPE) {
+        if (!responseJson.saml_response) {
+          throw Error(
+            `Executable response must contain a 'saml_response' field when token_type=${SAML_SUBJECT_TOKEN_TYPE}.`
+          );
+        }
+        this.subjectToken = responseJson.saml_response;
+      } else if (
```
lsirac, line 152: This else if condition is redundant given that you validated the token type above

src/auth/pluggableauthclient.ts
```diff
@@ -0,0 +1,253 @@
...
+
+import {
+  BaseExternalAccountClient,
+  BaseExternalAccountClientOptions,
+} from './baseexternalclient';
+import {RefreshOptions} from './oauth2client';
+import {ExecutableResponse} from './executableresponse';
+
+/**
+ * Interface defining JSON format for pluggable auth credentials.
```
lsirac, line 23: Defines the credential source format for pluggable auth credentials.

src/auth/pluggableauthclient.ts
```diff
@@ -0,0 +1,253 @@
...
+    /**
+     * Optional output file location that will be checked for a cached response
+     * from a previous run of the executable.
+     */
+    output_file?: string;
+  };
+}
+
+/**
+ * Error class for errors thrown from executable run by PluggableAuthClient.
```
lsirac, line 46: the executable

src/auth/pluggableauthclient.ts
```diff
@@ -0,0 +1,253 @@
...
+const GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES =
+  'GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES';
+
+/**
+ * Maximum currently supported executable version.
+ */
+const MAXIMUM_EXECUTABLE_VERSION = 1;
+
+/**
+ * Pluggable auth external account client. This is used to call a user provided
```
lsirac, line 88: Maybe worth adding some more info here (you can see what I wrote in my PR)

src/auth/pluggableauthclient.ts
```diff
@@ -0,0 +1,253 @@
...
+ * executable which returns a subject token to be exchanged for a
+ * Google access token.
+ */
+export class PluggableAuthClient extends BaseExternalAccountClient {
+  /**
+   * Command used to retrieve the third party token.
```
lsirac, line 94: Not sure if this needs to be "The command" e.g. the/an X everywhere else

src/auth/pluggableauthclient.ts
```diff
@@ -0,0 +1,253 @@
...
+   */
+  private readonly timeoutMillis: number;
+  /**
+   * Path to file to check for cached executable response.
+   */
+  private readonly outputFile?: string;
+
+  /**
+   * Instantiates a PluggableAuthClient instance using the provided JSON
+   * Object loaded from an external account credentials file.
```
lsirac, line 109: object

src/auth/pluggableauthclient.ts
```diff
@@ -0,0 +1,253 @@
...
+    options: PluggableAuthClientOptions,
+    additionalOptions?: RefreshOptions
+  ) {
+    super(options, additionalOptions);
+    this.command = options.credential_source.command;
+    if (!this.command) {
+      throw new Error('No valid Pluggable Auth "credential_source" provided.');
+    }
+    // Check if the provided timeout exists and if it is valid.
+    if (options.credential_source.timeout_millis === undefined) {
```
lsirac, line 127: !options.credential_source.timeout_millis?

src/auth/executableresponse.ts
aeitzman: is there something special about this file that means it should have the hyphen or should all the new files have that format?

src/auth/pluggableauthclient.ts
aeitzman: There's an edge case here I found in testing where if a timeout of 0 is provided, then that would return false and the default timeout would be used instead of throwing the invalid timeout error. I don't think its a huge deal either way since a user provided timeout of 0 doesn't make any sense in the first place, but I think having it be  explicit what we are checking here and only using the default if the field isn't provided makes sense.

src/auth/executableresponse.ts
aeitzman: Added typed errors for specific executableresponse validation issues

src/auth/externalclient.ts
aeitzman: That seems fine to me, the only change I'll have to make for that is to add the "next major: breaking change" tag to the final PR into main correct?

src/auth/executableresponse.ts
bcoe: Ideally we'd have a hypten separating words:

`executable-response.ts`, `external-client.ts`, etc., just a stylistic nit we could fix later -- I realize the codebase isn't consistent.

src/auth/externalclient.ts
bcoe: I would merge this PR as:

```
feat!: Adding Plugable Auth Support for ADC
```

And we should make sure to remember this is a breaking change when we land the feature -- before we merge your branch I would also like to test with gax, grpc, google-api-nodejs-client, and a GAPIC library.

bcoe (approved):

lsirac (approved):

Issue #1382, opened 2022-03-13
Support ADC json when impersonating service user


bailer:
**Is your feature request related to a problem? Please describe.**
When generating an ADC json you are able to impersonate a service user which outputs a JSON file that is not currently handled.
e.g. 
```
gcloud auth application-default login --impersonate-service-account=server@example.iam.gserviceaccount.com
```
Which generates the following JSON format: 
```
{
  "delegates": [],
  "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/server@example.iam.gserviceaccount.com:generateAccessToken",
  "source_credentials": {
    "client_id": "1234567890.apps.googleusercontent.com",
    "client_secret": "<secret>",
    "refresh_token": "<refrest_token>",
    "type": "authorized_user"
  },
  "type": "impersonated_service_account"
}
```
Currently the library throws the following error when doing this:
`The incoming JSON object does not contain a client_email field`
 **Describe the solution you'd like**
I'd like the library to detect this format and handle it correctly
 **Describe alternatives you've considered**
I can manually alter the file by putting the contents of `source_credentials` at the top level of the JSON (though unsure if this has any side-effect). I can also generate a key for the service user but since it's not recommended, impersonating the service user and using ADC would be preferred.
 **Additional context**
I'm very new to GCP and authentication in that regard and was trying to set up local development without generating keys but still using a service user. I found that you can impersonate a service user while generating the ADC JSON but it's not working with this library. I found that e.g. `google-auth-library-python` supports this format so I'm thinking the node library also should.

imathews:
Just a friendly +1 on this — I'm running into the same issue with `google-auth-node`, which is preventing us from utilizing impersonated service accounts in our local dev environment.

PR #1425, opened 2022-06-27
feat: handle impersonated ADC

cstanger:
------------

This PR is intent to add the handing of impersonated ADC in the context of the credentials instance creation.

Fixes 
- #[1382](https://github.com/googleapis/google-auth-library-nodejs/issues/1382) 
- #[1382](https://github.com/googleapis/google-api-nodejs-client/issues/2932 ) (googleapis)

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -24,13 +24,13 @@ import {Crypto, createCrypto} from '../crypto/crypto';
 import {DefaultTransporter, Transporter} from '../transporters';
 
 import {Compute, ComputeOptions} from './computeclient';
-import {CredentialBody, JWTInput} from './credentials';
+import {CredentialBody, ImpersonatedJWTInput, JWTInput} from './credentials';
 import {IdTokenClient} from './idtokenclient';
 import {GCPEnv, getEnv} from './envDetect';
 import {JWT, JWTOptions} from './jwtclient';
 import {Headers, OAuth2ClientOptions, RefreshOptions} from './oauth2client';
 import {UserRefreshClient, UserRefreshClientOptions} from './refreshclient';
-import {Impersonated, ImpersonatedOptions} from './impersonated';
+import {Impersonated, ImpersonatedOptions, IMPERSONATED_ACCOUNT_TYPE} from './impersonated';
 import {
   ExternalAccountClient,
   ExternalAccountClientOptions,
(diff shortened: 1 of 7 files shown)


src/auth/googleauth.ts
```diff
@@ -459,13 +463,74 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
+    if (start_index === -1 || end_index === -1 || start_index > end_index) {
+      throw new Error(
+        `Cannot extract target principal from ${impersonation_url}`
+      );
+    }
+
+    const target_principal = impersonation_url.substring(
+      start_index + 1,
+      end_index
+    );
```
danielbankhead, line 513: This works, but how about a regex for this portion? This regex captures all characters after the last `/` (as `generateAccessToken$` matches the end of the output) and labels it 'target' if it's available.

```suggestion
    const targetPrincipal = 
    /(?<target>[^/]+):generateAccessToken$/.exec(json.service_account_impersonation_url)?.groups?.target;
    
    if (!targetPrincipal) {
      throw new RangeError(
        `Cannot extract target principal from ${json.service_account_impersonation_url}`
      );
    }
```

src/auth/googleauth.ts
```diff
@@ -459,13 +463,74 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
+
+    // Create source client for impersonation
+    const source_client = new UserRefreshClient(
+      json.source_credentials.client_id,
+      json.source_credentials.client_secret,
+      json.source_credentials.refresh_token
+    );
+
+    // Extreact service account from service_account_impersonation_url
+    const impersonation_url = json.service_account_impersonation_url;
```
danielbankhead, line 501: The variable names here should be in `camelCase`

src/auth/impersonated.ts
```diff
@@ -30,7 +30,7 @@ export interface ImpersonatedOptions extends RefreshOptions {
   /**
    * Scopes to request during the authorization grant.
    */
-  targetScopes?: string[];
+  targetScopes?: string | string[];
```
danielbankhead, line 33: Any way to implement this without changing this type?

danielbankhead (changes requested):
Looks great! Just a few items before we're ready to merge.

src/auth/impersonated.ts
cstanger: indeed that's possible by using `targetScopes: this.getAnyScopes() as string[],` when instantiating the `Impersonated` class (in src/auth/googleauth.ts line 519)

The `BaseExternalAccountClient` as well as the  `JWT` class implement the scopes attribute with `string | string[]; `, thats why I thought to make it consistent. Also in oder to use the `getAnyScopes()` helper directly  

Any concerns adopting the ImpersonatedOptions type ?

src/auth/googleauth.ts
cstanger: 👍 will adopt that

src/auth/googleauth.ts
cstanger: Will adopt to regex and explicitly test for correct targetPrincipal handling

src/auth/impersonated.ts
danielbankhead: Since this is a `public` property that's not a union type, this would technically be breaking for consumers that expect this to be an `Array`

danielbankhead (approved):

PR #1431, opened 2022-07-11
feat: adding executable and file handling for Pluggable Auth Client

aeitzman:
Added PluggableAuthHandler to run user provided executable and read from cached file output + associated tests.
Also added integration test and readme documentation changes.


bcoe:
@sofisl @alexander-fenster, let's prioritize this review in our backlog grooming tomorrow.

src/auth/pluggable-auth-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+        reject(new Error('Executable failed due to timeout.'));
+      }, this.timeoutMillis);
+
+      child.on('close', (code: number) => {
+        // Cancel timeout if executable closes before timeout is reached.
+        clearTimeout(timeout);
+        if (code === 0) {
+          // If executable completed successfully, try to return parsed response
+          const responseJson = JSON.parse(output) as ExecutableResponseJson;
+          resolve(new ExecutableResponse(responseJson));
```
lsirac, line 107: We are going to make expiration_time optional, can you do that in this PR?

src/auth/pluggable-auth-client.ts
```diff
@@ -246,14 +262,19 @@ export class PluggableAuthClient extends BaseExternalAccountClient {
...
           serviceAccountEmail
         );
       }
-      executableResponse = await this.retrieveResponseFromExecutable(envMap);
+      executableResponse = await this.handler.retrieveResponseFromExecutable(
+        envMap
+      );
     }
 
     if (executableResponse) {
```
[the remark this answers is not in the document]
aeitzman: yes, since executableResponse is set to be type ExecutableResponse | undefined (which is needed to have the behavior where retrieveCachedResponse can return undefined if no response is found). If this if check doesn't surround the rest of the logic it will throw ts errors when trying to access the executableResponse properties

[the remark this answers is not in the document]
lsirac: seems like retrieveResponseFromExecutable either returns it or throws?

src/auth/executable-response.ts
```diff
@@ -183,15 +178,15 @@ export class ExecutableResponse {
...
+   * provided timeout has passed.
    */
   isExpired(): boolean {
-    return (
-      !this.expirationTime ||
-      this.expirationTime < Math.round(Date.now() / 1000)
-    );
+    if (!this.expirationTime) {
+      return false;
+    } else {
```
lsirac, line 187: nit: remove the else

src/auth/pluggable-auth-handler.ts
```diff
@@ -0,0 +1,208 @@
...
+export class PluggableAuthHandler {
+  private readonly commandComponents: Array<string>;
+  private readonly timeoutMillis: number;
+  private readonly outputFile?: string;
+
+  /**
+   * Instantiates a PluggableAuthHandler instance using the provided
+   * PluggableAuthHandlerOptions object.
+   */
+  constructor(options: PluggableAuthHandlerOptions) {
```
lsirac, line 57: Maybe options should be passed to retrieveResponseFromExecutable instead?

src/auth/pluggable-auth-client.ts
```diff
@@ -69,7 +72,9 @@ export class ExecutableError extends Error {
   readonly code: string;
 
   constructor(message: string, code: string) {
-    super(message);
+    super(
+      `The executable failed with exit code: ${code} and error message: ${message}.`
+    );
     this.name = 'ExecutableError';
     this.code = code;
     Object.setPrototypeOf(this, ExecutableError.prototype);
```
lsirac, line 80: We should mention that expiration_time is optional near the response samples on line 140ish

src/auth/pluggable-auth-handler.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  /**
+   * Calls user provided executable to get a 3rd party subject token and
+   * returns the response.
+   * @param envMap a Map of additional Environment Variables required for
+   *   the executable.
+   * @return A promise that resolves with the executable response.
+   */
+  retrieveResponseFromExecutable(
```
lsirac, line 78: Let's do the GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES check here and not in the pluggable-auth-client since this is where executable handling happens

src/auth/pluggable-auth-handler.ts
```diff
@@ -0,0 +1,208 @@
...
+              new Error(
+                `The output file contained an invalid response: ${responseString}`
+              )
+            );
+          }
+        });
+    });
+  }
+
+  private static parseCommand(command: string): Array<string> {
```
lsirac, line 191: Add a comment here to explain

src/auth/pluggable-auth-client.ts
```diff
@@ -246,72 +264,38 @@ export class PluggableAuthClient extends BaseExternalAccountClient {
...
-      }
-      // Check that response is not expired.
-      if (executableResponse.isExpired()) {
-        throw new Error('Executable response is expired.');
-      }
-      // Return subject token from response.
-      return executableResponse.subjectToken as string;
-    } else {
-      throw new Error('No valid response returned from executable.');
+    if (executableResponse.version > MAXIMUM_EXECUTABLE_VERSION) {
```
lsirac, line 282: why not validate executable response in the handler?

test/test.pluggableauthhandler.ts
```diff
@@ -0,0 +1,455 @@
...
+    expectedEnvMap.set('GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE', '0');
+    const expectedOpts = {
+      env: {
+        ...process.env,
+        ...Object.fromEntries(expectedEnvMap),
+      },
+    };
+
+    beforeEach(() => {
+      // Stub environment variables and set Allow Executables environment variable to 1
```
lsirac, line 122: // Stub environment variables and set GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES variable to 1.

test/test.pluggableauthhandler.ts
```diff
@@ -0,0 +1,455 @@
...
+        GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES: '1',
+      });
+      sandbox.stub(process, 'env').value(envVars);
+      clock = sandbox.useFakeTimers({now: referenceTime});
+
+      defaultResponseJson = {
+        success: true,
+        version: 1,
+        token_type: SAML_SUBJECT_TOKEN_TYPE,
+        saml_response: 'response',
```
lsirac, line 133: Looks like there are no OIDC tests?

test/test.pluggableauthhandler.ts
```diff
@@ -0,0 +1,455 @@
...
+        expiration_time: referenceTime / 1000 + 10,
```
lsirac, line 134: Should also test successful responses without expiration_time

test/test.pluggableauthhandler.ts
```diff
@@ -0,0 +1,455 @@
...
+        new Map<string, string>()
+      );
+      clock.tick(10001);
+
+      await assert.rejects(response, expectedError);
+    });
+
+    it('should throw error when non-json text is returned', async () => {
+      const expectedError = new Error(
+        'The executable returned an invalid response: Invalid response'
```
lsirac, line 274: Maybe "INVALID EXECUTABLE RESPONSE" + a comment that the executable response is appended? Took me a sec to realize you were testing that

test/test.pluggableauthhandler.ts
aeitzman: I didn't test different options for token type in the handler tests since the handler itself doesn't really do anything differently based off response type, it just creates a new instance of ExecutableResponse. The executable response tests test all the different options to make sure that the JSON is parsed correctly and subject token is returned.

lsirac: I understand that POV but we should still have tests that cover OIDC. This way we know that the right values are being passed when creating an ExecutableResponse.

src/auth/pluggable-auth-handler.ts
aeitzman: The options object also contains the output file which is needed by the retrieveCachedResponse function. Could split the options out and just pass them all directly to the functions but having the constructor seemed a bit cleaner to me.

src/auth/pluggable-auth-handler.ts
aeitzman: That would change the behavior. If that check is moved, the output file will be checked and returned even if the env variable is not set. We don't want to return anything if the check fails right?

src/auth/pluggable-auth-client.ts
aeitzman: Having the final validation here makes it so its only done once - the handler implementation here doesn't have a single method that calls both the executable and the file cache, the client calls both individually and validated after.

lsirac: I think it's better not to mix the responsibilities of the client and the handler. There's nothing stopping you from adding a method in the handler that does both (checks cache, calls executable).

src/auth/pluggable-auth-handler.ts
lsirac: I think I covered that in a different comment (having a method that does both and doing it there) but we discussed offline.

src/auth/pluggable-auth-handler.ts
aeitzman: Discussed offline

src/auth/pluggable-auth-client.ts
aeitzman: Discussed offline

test/test.pluggableauthhandler.ts
```diff
@@ -159,7 +157,7 @@ describe('PluggableAuthHandler', () => {
       );
     });
 
-    it('should return executable response when successful', async () => {
+    it('should return SAML executable response when successful', async () => {
```
lsirac, line 160: Should have both an OIDC and SAML test without output_file specified where expiration time is not present in the executable response.

README.md
```diff
@@ -522,6 +522,136 @@ Where the following variables need to be substituted:
...
+{
+  "version": 1,
+  "success": false,
+  "code": "401",
+  "message": "Caller not authorized."
+}
+```
+These are all required fields for an error response. The code and message
+fields will be used by the library as part of the thrown exception.
+
```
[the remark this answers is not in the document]
aeitzman: I put that part up above - made more sense to me to have that immediately follow the successful response examples rather than the error example - line 601

README.md
```diff
@@ -522,6 +522,136 @@ Where the following variables need to be substituted:
...
+##### Security considerations
+The following security practices are highly recommended:
+* Access to the script should be restricted as it will be displaying credentials to stdout. This ensures that rogue processes do not gain  access to the script.
+* The configuration file should not be modifiable. Write access should be restricted to avoid processes modifying the executable command portion.
+
+Given the complexity of using executable-sourced credentials, it is recommended to use
+the existing supported mechanisms (file-sourced/URL-sourced) for providing 3rd party
+credentials unless they do not meet your specific requirements.
+
+You can now [use the Auth library](#using-external-identities) to call Google Cloud
```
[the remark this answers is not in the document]
aeitzman: sounds good - its also this way in the current java readme so you should make that change in your PR that's updating the docs there as well.

lsirac (approved):

Issue #1434, opened 2022-07-29
Refactoring: Quit mixing Promises and async/await


shogo-nakano-desu:
### Problem
- There is a comment by @bcoe that says `TODO`. This comment has been here for two years. 
  - https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/googleauth.ts#L206-L232

### Suggestion
- It would be better to use `async/await` only and reduce the length of code. It will make this repo easier to see through.
- I am going to create PR when this change is accepted but beforehand, I put my plan below. Please use these code for your decision.

#### Before
```ts
// TODO: refactor the below code so that it doesn't mix and match
      // promises and async/await.
      this._getDefaultProjectIdPromise = new Promise(
        // eslint-disable-next-line no-async-promise-executor
        async (resolve, reject) => {
          try {
            const projectId =
              this.getProductionProjectId() ||
              (await this.getFileProjectId()) ||
              (await this.getDefaultServiceProjectId()) ||
              (await this.getGCEProjectId()) ||
              (await this.getExternalAccountClientProjectId());
            this._cachedProjectId = projectId;
            if (!projectId) {
              throw new Error(
                'Unable to detect a Project Id in the current environment. \n' +
                  'To learn more about authentication and Google APIs, visit: \n' +
                  'https://cloud.google.com/docs/authentication/getting-started'
              );
            }
            resolve(projectId);
          } catch (e) {
            reject(e);
          }
        }
      );
    }
```
#### After
```ts
this._getDefaultProjectIdPromise = (async () => {
        const projectId =
          this.getProductionProjectId() ||
          (await this.getFileProjectId()) ||
          (await this.getDefaultServiceProjectId()) ||
          (await this.getGCEProjectId()) ||
          (await this.getExternalAccountClientProjectId());
        this._cachedProjectId = projectId;
        if (!projectId) {
          throw new Error(
            'Unable to detect a Project Id in the current environment. \n' +
              'To learn more about authentication and Google APIs, visit: \n' +
              'https://cloud.google.com/docs/authentication/getting-started'
          );
        }
        return projectId;
      })();
    }
```

sofisl:
This looks good to me, as long as tests pass as well :) Thank you!

shogo-nakano-desu:
Thank you. I am going to create PR 👍

danielbankhead:
Fixed in https://github.com/googleapis/google-auth-library-nodejs/pull/1436

PR #1436, opened 2022-08-03
refactor: remove comment and promises, instead only using async/await

shogo-nakano-desu:
## Description
- There is a comment by @bcoe that says TODO [here](https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/googleauth.ts#L206-L232). So I deleted Promises and changed to use async/await only.

## Check list
  - Here is the issue https://github.com/googleapis/google-auth-library-nodejs/issues/1434
  - To be precise, code coverage "rate" in `googleauth.ts` decreases even though uncovered lines stay the same because lines of code decrease. But I think it is not a problem because actual coverage does not become worse. 
  - Do not need to update in this PR

Refactor #1434 🦕


danielbankhead (approved):
Thanks! LGTM

danielbankhead (approved):
Thanks! LGTM

PR #1437, opened 2022-08-04
feat: adds Pluggable Auth support

### LINKED ISSUE #1433: [tracking issue]: adding executable and file handling for Pluggable Auth Client
Tracking issue for: https://github.com/googleapis/google-auth-library-nodejs/pull/1431

aeitzman:
See go/pluggable-auth-design.

Summary:

Adds a new PluggableAuthClient class that enables the ability for developers to use their own executable to retrieve subject tokens for Workload Identity Federation (instead of a file/URL).
ADC is extended to support this new client.
Adds documentation in Readme
Adds integration test

Fixes: #1433


src/auth/executable-response.ts
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+ * An error thrown by the ExecutableResponse class.
+ */
+export class ExecutableResponseError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'ExecutableResponseError';
+    Object.setPrototypeOf(this, ExecutableResponseError.prototype);
+  }
```
danielbankhead, line 200: I like the use of descriptive error classes, however this constructor, and similar constructors (e.g. `InvalidVersionFieldError`, `InvalidSuccessFieldError`) shouldn't be necessary:
- `name` isn't referenced outside of the class
- `setPrototypeOf` doesn't supply additional functionality
- if the above 2 are removed, the default constructor contains the `super` call

src/auth/pluggable-auth-client.ts
```diff
@@ -0,0 +1,315 @@
...
+   */
+  readonly code: string;
+
+  constructor(message: string, code: string) {
+    super(
+      `The executable failed with exit code: ${code} and error message: ${message}.`
+    );
+    this.name = 'ExecutableError';
+    this.code = code;
+    Object.setPrototypeOf(this, ExecutableError.prototype);
```
danielbankhead, line 83: - `name` isn't referenced outside of the class
- `setPrototypeOf` doesn't supply additional functionality

src/auth/pluggable-auth-client.ts
```diff
@@ -0,0 +1,315 @@
...
+    additionalOptions?: RefreshOptions
+  ) {
+    super(options, additionalOptions);
+    if (!options.credential_source.executable) {
+      throw new Error('No valid Pluggable Auth "credential_source" provided.');
+    }
+    this.command = options.credential_source.executable.command;
+    if (!this.command) {
+      throw new Error('No valid Pluggable Auth "credential_source" provided.');
+    }
```
danielbankhead, line 208: Optional: I think the type-level checking alone should suffice

src/auth/pluggable-auth-client.ts
```diff
@@ -0,0 +1,315 @@
...
+    // Check if the executable is allowed to run.
+    if (process.env[GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES] !== '1') {
+      throw new Error(
+        'Pluggable Auth executables need to be explicitly allowed to run by ' +
+          'setting the GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES environment ' +
+          'Variable to 1.'
+      );
+    }
+
+    let executableResponse: ExecutableResponse | undefined;
```
danielbankhead, line 260: This should default to `= undefined`, otherwise this would produce an error at runtime if `#outputFile`  is falsy

src/auth/pluggable-auth-handler.ts
```diff
@@ -0,0 +1,219 @@
...
+  private readonly outputFile?: string;
+
+  /**
+   * Instantiates a PluggableAuthHandler instance using the provided
+   * PluggableAuthHandlerOptions object.
+   */
+  constructor(options: PluggableAuthHandlerOptions) {
+    if (!options.command) {
+      throw new Error('No command provided.');
+    }
```
danielbankhead, line 59: Optional: I think the type-level checking alone should suffice

src/auth/pluggable-auth-handler.ts
```diff
@@ -0,0 +1,219 @@
...
+    this.commandComponents = PluggableAuthHandler.parseCommand(
+      options.command
+    ) as Array<string>;
+    this.timeoutMillis = options.timeoutMillis;
+    if (!this.timeoutMillis) {
+      throw new Error('No timeoutMillis provided.');
+    }
```
danielbankhead, line 66: Is `timeoutMillis = 0` a valid configuration? If so, we can remove this check

src/auth/pluggable-auth-handler.ts
```diff
@@ -0,0 +1,219 @@
...
+   * @param envMap a Map of additional Environment Variables required for
+   *   the executable.
+   * @return A promise that resolves with the executable response.
+   */
+  retrieveResponseFromExecutable(
+    envMap: Map<string, string>
+  ): Promise<ExecutableResponse> {
+    return new Promise((resolve, reject) => {
+      // Spawn process to run executable using added environment variables.
+      const child = childProcess.spawn(
```
danielbankhead, line 82: What are your thoughts on using `childProcess.exec` instead? It seems this may do what you want and with reduced code here (e.g. `stdout` handling and manually setting up the `setTimeout`)

src/auth/pluggable-auth-handler.ts
```diff
@@ -0,0 +1,219 @@
...
+        // If file path cannot be resolved, return undefined.
+        return resolve(undefined);
+      }
+
+      if (!fs.lstatSync(filePath).isFile()) {
+        // If path does not lead to file, return undefined.
+        return resolve(undefined);
+      }
+
+      const readStream = fs.createReadStream(filePath);
```
danielbankhead, line 161: What are your thoughts on using `childProcess.readFile` instead? It may shrink this logic a bit

src/auth/pluggable-auth-handler.ts
```diff
@@ -0,0 +1,219 @@
...
+
+      let filePath: fs.PathLike;
+      try {
+        filePath = fs.realpathSync(this.outputFile as string);
+      } catch {
+        // If file path cannot be resolved, return undefined.
+        return resolve(undefined);
+      }
+
+      if (!fs.lstatSync(filePath).isFile()) {
```
danielbankhead, line 156: This creates a blocking call; `await fs.promises.lstat` could be a user-friendlier alternative

src/auth/pluggable-auth-handler.ts
```diff
@@ -0,0 +1,219 @@
...
+   */
+  retrieveCachedResponse(): Promise<ExecutableResponse | undefined> {
+    return new Promise((resolve, reject) => {
+      if (!this.outputFile || this.outputFile.length === 0) {
+        return resolve(undefined);
+      }
+
+      let filePath: fs.PathLike;
+      try {
+        filePath = fs.realpathSync(this.outputFile as string);
```
danielbankhead, line 150: This creates a blocking call; `await fs.promises.realpath` could be a user-friendlier alternative

danielbankhead (changes requested):
Looks good so far - a couple of changes requested

danielbankhead:
What makes this PR breaking? It seems it can work without breaking existing users.

src/auth/pluggable-auth-handler.ts
aeitzman: Exec was considered, used spawn instead because exec has a max buffer size - part of the language agnostic design was that we wanted to be able to handle very large output which spawn can deal with but exec can't necessarily. SetTimeout is also available for spawn, I chose to implement a manual timeout though since from what I was reading, there isn't actually a way to determine if the child process was closed due to timeout (it just sends a SIGTERM signal, which could come from an external source as well). With the manual timeout implementation, can be more confident that timeout was the issue when throwing the error.

aeitzman:
> What makes this PR breaking? It seems it can work without breaking existing users.

On a previous PR, @bcoe was saying this should be a breaking change: https://github.com/googleapis/google-auth-library-nodejs/pull/1419#pullrequestreview-1003138398

danielbankhead:
> 



> > What makes this PR breaking? It seems it can work without breaking existing users.
> 
> On a previous PR, @bcoe was saying this should be a breaking change: [#1419 (review)](https://github.com/googleapis/google-auth-library-nodejs/pull/1419#pullrequestreview-1003138398)

That makes sense - it looks like `PluggableAuthClientOptions extends BaseExternalAccountClientOptions` so I think we're good there now.

danielbankhead (approved):
LGTM - one test to address, then this should be gtg

src/auth/pluggable-auth-handler.ts
aeitzman: no, 0 would not be valid

src/auth/pluggable-auth-handler.ts
aeitzman: ack

src/auth/executable-response.ts
```diff
@@ -0,0 +1,278 @@
...
+    if (!responseJson.version) {
+      throw new InvalidVersionFieldError(
+        "Executable response must contain a 'version' field."
+      );
+    }
+    if (responseJson.success === undefined) {
+      throw new InvalidSuccessFieldError(
+        "Executable response must contain a 'success' field."
+      );
+    }
```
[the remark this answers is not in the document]
aeitzman: Ack

src/auth/pluggable-auth-client.ts
aeitzman: ack

lsirac (approved):
Thanks Alex!

danielbankhead (approved):
LGTM

PR #1439, opened 2022-08-17
feat: Add `generateIdToken` support for `Impersonated` Clients

### LINKED ISSUE #1318: Impersonated credentials should implement `IdTokenProvider` interface
Currently there is no easy way to acquire an `id_token` for a service account that was impersonated.

For example, if you run an application as SA1 but you would like to get an `id_token` for SA2, you would have to first use `Impersonated` module to get a raw `access_token` for SA2 and then use that in [iamcredentials.generateIdToken()](https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateIdToken) api call manually.

in another example, if you run running [workload identity federation](https://cloud.google.com/iam/docs/workload-identity-federation) (WIF) in AWS and need an `id_token` to access Cloud Run, you would need to acquire the `access_token` for the service account for WIF on aws and then use that same `generateIDToken()` api call.

This FR is to allow the `Impersonated` module to acquire its own `id_token`

---

the workaround i tried was to edit the following

* node_modules/google-auth-library/build/src/auth/impersonated.d.ts
```javascript
export declare class Impersonated extends OAuth2Client implements IdTokenProvider {}

    /**
     * Fetches an ID token.
     * @param targetAudience the audience for the fetched ID token.
     */
    fetchIdToken(targetAudience: string): Promise<string>;    
```

* node_modules/google-auth-library/build/src/auth/impersonated.js
```javascript
     async fetchIdToken(targetAudience) {
        try {
            await this.sourceClient.getAccessToken();
            const name = 'projects/-/serviceAccounts/' + this.targetPrincipal;
            const u = `${this.endpoint}/v1/${name}:generateIdToken`;
            const body = {
                delegates: this.delegates,
                audience: targetAudience,
                includeEmail: true,
            };
            const res = await this.sourceClient.request({
                url: u,
                data: body,
                method: 'POST',
            });
            const tokenResponse = res.data;

            return tokenResponse.token
        }
        catch (error) {
     ...
        }
    } 
```

then the usage would be 

```javascript
    const scopes = 'https://www.googleapis.com/auth/cloud-platform'
    const auth =  new GoogleAuth({
        scopes: scopes
    });

    // First impersonate
    let targetCredentials = 'target-serviceaccount@fabled-ray-104117.iam.gserviceaccount.com'
        sourceClient: client,
        targetPrincipal: targetCredentials,
        lifetime: 30,
        delegates: [],
        targetScopes: [scopes]
    });

    // then get an ID Token
    let idClient = new IdTokenClient({
        targetAudience: 'https://foo.bar',
        idTokenProvider: targetClient
    })

    const res = await idClient.request({
        method: 'GET',
        url: 'https://httpbin.org/get',
      });
    console.log(res.data);

```

---

Finally, please note the following:

ref:
* https://github.com/googleapis/google-auth-library-nodejs/issues/1305
* https://github.com/googleapis/google-auth-library-nodejs/issues/1210
(The last issue describes the limits of using the impersonated module with google cloud client libraries)

danielbankhead:
Fixes #1318 🦕

### DIFF
--- a/src/auth/impersonated.ts
+++ b/src/auth/impersonated.ts
@@ -16,6 +16,7 @@
 
 import {GetTokenResponse, OAuth2Client, RefreshOptions} from './oauth2client';
 import {AuthClient} from './authclient';
+import {IdTokenProvider} from './idtokenclient';
 import {GaxiosError} from 'gaxios';
 
 export interface ImpersonatedOptions extends RefreshOptions {
@@ -52,7 +53,12 @@ export interface TokenResponse {
   expireTime: string;
 }
 
-export class Impersonated extends OAuth2Client {
+export interface FetchIdTokenResponse {
+  /** The OpenId Connect ID token. */
+  token: string;
+}
+
+export class Impersonated extends OAuth2Client implements IdTokenProvider {
   private sourceClient: AuthClient;
   private targetPrincipal: string;
   private targetScopes: string[];
@@ -153,5 +159,29 @@ export class Impersonated extends OAuth2Client {
         throw error;
       }
     }
+  }
+
+  /**
+   * Generates an OpenID Connect ID token for a service account.
+   *
+   * @param targetAudience the audience for the fetched ID token.
+   */
+  async fetchIdToken(targetAudience: string): Promise<string> {
+    await this.sourceClient.getAccessToken();
+
+    const name = `projects/-/serviceAccounts/${this.targetPrincipal}`;
+    const u = `${this.endpoint}/v1/${name}:generateIdToken`;
+    const body = {
+      delegates: this.delegates,
+      audience: targetAudience,
+      includeEmail: true,
+    };
+    const res = await this.sourceClient.request<FetchIdTokenResponse>({
+      url: u,
+      data: body,
+      method: 'POST',
+    });
+
+    return res.data.token;
   }
 }
--- a/src/auth/impersonated.ts
+++ b/src/auth/impersonated.ts
@@ -53,6 +53,14 @@ export interface TokenResponse {
   expireTime: string;
 }
 
+export interface FetchIdTokenOptions {
+  /**
+   * Include the service account email in the token.
+   * If set to `true`, the token will contain `email` and `email_verified` claims.
+   */
+  includeEmail: boolean;
+}
+
 export interface FetchIdTokenResponse {
   /** The OpenId Connect ID token. */
   token: string;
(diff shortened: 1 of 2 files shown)


FrodoTheTrue (approved):

src/auth/impersonated.ts
```diff
@@ -154,4 +160,28 @@ export class Impersonated extends OAuth2Client {
...
+   */
+  async fetchIdToken(targetAudience: string): Promise<string> {
+    await this.sourceClient.getAccessToken();
+
+    const name = `projects/-/serviceAccounts/${this.targetPrincipal}`;
+    const u = `${this.endpoint}/v1/${name}:generateIdToken`;
+    const body = {
+      delegates: this.delegates,
+      audience: targetAudience,
+      includeEmail: true,
```
salrashid123, line 177: just a small comment on this.   The `includeEmail` is an optional parameter for the remote api call thats set specifically set here to true.

Certain services on GCP (and def. other places) look for the `email` field in the OIDC token ([IAP programmaic access as an example](https://cloud.google.com/iap/docs/authentication-howto#obtaining_an_oidc_token_in_all_other_cases)... (trust me, it doens't say the email is required in the docs but IAP def needs it)

---

Coming back to the is PR, i don't know much node so I don't see any feasible programmatic structure here to optionally set if includeEmail is required or not.  For example, in go you can:

https://github.com/googleapis/google-api-go-client/blob/main/impersonate/idtoken.go#L30


 If not structurally feasible in node, IMO, its fine to keep default (similar to what golang does here when accessing compute engine api for its id_token (format=full for gce retruns the email)

https://github.com/googleapis/google-api-go-client/blob/main/idtoken/compute.go#L41

danielbankhead: > Certain services on GCP (and def. other places) look for the email field in the OIDC token ([IAP programmaic access as an example](https://cloud.google.com/iap/docs/authentication-howto#obtaining_an_oidc_token_in_all_other_cases)... (trust me, it doens't say the email is required in the docs but IAP def needs it)

Yep, it's in a blue section here: ["Note: Ensure that the generated ID token has an 'email' claim. This is a requirement for IAP."](https://cloud.google.com/iap/docs/authentication-howto#authenticating_from_a_service_account)

> Coming back to the is PR, i don't know much node so I don't see any feasible programmatic structure here to optionally set if includeEmail is required or not.

It's pretty trivial; I'll refactor and default it to `true`.

ruyadorno (approved):

PR #1447, opened 2022-08-24
fix: remove `projectId` check for `signBlob` calls

### LINKED ISSUE #421: Make the projectId optional for signing with IAM
###### Copied from original issue: https://github.com/stephenplusplus/google-auto-auth/issues/53

<a href="/hiranya911"></a>@hiranya911<br>July 10, 2018 11:27 PM

The signWithApi method currently requires a `projectId` field:

```
_signWithApi (data, callback) {
    if (!this.projectId) {
      callback(new Error('Cannot sign data without a project ID.'));
      return;
    }
    ...
}
```

However, IAM's [signBlob endpoint](https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts/signBlob) can be called without a `projectId` if necessary (using `-` as a wildcard):

```
/projects/-/serviceAccounts/${clientEmail}
```

Can we modify the implementation to default to `-` wildcard when a `projectId` is not given explicitly? This will greatly simplify the usage of the library in some high-level applications, such as [`firebase-admin`](https://github.com/firebase/firebase-admin-node/blob/master/package.json).

danielbankhead:
Removes `projectId` checking for `signBlob` calls.

While we check for it, it is not used for the request (in fact, [it is invalid to use according to the docs](https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signBlob#path-parameters)).

Fixes #421 🦕


src/auth/googleauth.ts
```diff
@@ -203,23 +203,32 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
-          (await this.getExternalAccountClientProjectId());
-        this._cachedProjectId = projectId;
-        if (!projectId) {
-          throw new Error(
-            'Unable to detect a Project Id in the current environment. \n' +
-              'To learn more about authentication and Google APIs, visit: \n' +
-              'https://cloud.google.com/docs/authentication/getting-started'
-          );
+      // TODO: refactor the below code so that it doesn't mix and match
+      // promises and async/await.
```
ruyadorno, line 207: agreed 😄

src/auth/googleauth.ts
```diff
@@ -203,23 +203,32 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
+              throw new Error(
+                'Unable to detect a Project Id in the current environment. \n' +
+                  'To learn more about authentication and Google APIs, visit: \n' +
+                  'https://cloud.google.com/docs/authentication/getting-started'
+              );
+            }
+            resolve(projectId);
+          } catch (e) {
+            reject(e);
+          }
```
ruyadorno, line 229: From what I can understand from the context of this PR, it seems to me that the goal of the proposed changes here is for the `this._getDefaultProjectIdPromise` variable to hold a `Promise<projectId>` value instead of directly holding the `projectId` value. With that in mind I believe this change would translate that intent, while allowing to get rid of the unwanted mix and match of promises/async/await:

```suggestion
      this._getDefaultProjectIdPromise =
        Promise.resolve(this.getProductionProjectId())
          .then(projectId => projectId ?? this.getFileProjectId())
          .then(projectId => projectId ?? this.getDefaultServiceProjectId())
          .then(projectId => projectId ?? this.getGCEProjectId())
          .then(projectId => projectId ?? this.getExternalAccountClientProjectId())
          .then(projectId => {
            if (projectId) {
              this._cachedProjectId = projectId;
              return projectId
            } else {
              throw new Error(
                'Unable to detect a Project Id in the current environment. \n' +
                  'To learn more about authentication and Google APIs, visit: \n' +
                  'https://cloud.google.com/docs/authentication/getting-started'
              );
            }
          })

```

ruyadorno (commented):
Unrelated to the changes here but I also believe the signature of the `getProjectIdAsync` method can be changed from `Promise<string | null>` to `Promise<string>` since it should never be able to hold a `null` value.

src/auth/googleauth.ts
danielbankhead: Interesting; it looks like this is from a bad merge 😅

The existing code should be correct (via this PR: https://github.com/googleapis/google-auth-library-nodejs/pull/1436)

ruyadorno (approved):

danielbankhead:
Hey @ruyadorno,

Looks like your personal address will need to sign the CLA 😅

Also, I did a further refactor based on your refactor, let me know what you think.

ruyadorno (approved):
🤩 that's a nice way to handle that! 👍

ruyadorno:
> Looks like your personal address will need to sign the CLA 😅

ooohhh, sorry about that! maybe that happened because I still had that email assigned as "primary" on GitHub, I changed it to be sure it won't happen again in the future, for now feel free to either squash that commit or maybe edit that commit message to fixup the email if you think it's worth the effort 😄

PR #1454, opened 2022-08-31
fix: do not use #private

alexander-fenster:
Adding a `#` private method is technically a breaking change. Our packing tests across all client libraries are now failing with

```
node_modules/google-auth-library/build/src/auth/googleauth.d.ts:67:5 - error TS18028: Private identifiers are only available when targeting ECMAScript 2015 and higher.

67     #private;
       ~~~~~~~~
```

(in CI, this looks like a `pack-n-play` TypeScript test failure as a part of system tests)

Let's not be that modern and use the regular private method instead.

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -206,7 +206,7 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
    *
    * @returns Promise that resolves with project id (or `null`)
    */
-  async #getProjectIdOptional(): Promise<string | null> {
+  private async getProjectIdOptional(): Promise<string | null> {
     try {
       return await this.getProjectId();
     } catch (e) {
@@ -232,7 +232,7 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
    *
    * @returns projectId
    */
-  async #findAndCacheProjectId(): Promise<string> {
+  private async findAndCacheProjectId(): Promise<string> {
     let projectId: string | null | undefined = null;
 
     projectId ||= await this.getProductionProjectId();
@@ -255,7 +255,7 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
     }
 
     if (!this._findProjectIdPromise) {
-      this._findProjectIdPromise = this.#findAndCacheProjectId();
+      this._findProjectIdPromise = this.findAndCacheProjectId();
     }
     return this._findProjectIdPromise;
   }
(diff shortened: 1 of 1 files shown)


danielbankhead (approved):

ruyadorno:
Oh that's interesting, [private class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) have been on Node.js since v12. I believe this is more of a matter of reconfiguring our TypeScript compiler to a little more modern level of ECMAScript support rather than being an unsupported feature.

That said, there's maybe an argument to be made on what is the more JS-idiommatic way of expressing the intent of a private method/property vs TS-idiommatic and I'm not sure where the collective consensus within the larger JavaScript ecosystem sits on that.

Anyways just wanted to highlight that it might be possible to support `#private` notations and that it might be more familiar to a lot of JS devs and hear your thoughts on it @alexander-fenster (I might be missing more context here, feel free to correct me if I'm making any wrong assumption) 😊

danielbankhead:
@ruyadorno no problem, we just need to update gts & TypeScript in https://github.com/google/pack-n-play (the repo needs some love).

DesAWSume:
a year after, this issue bothers me. 
`node_modules/google-auth-library/build/src/util.d.ts:134:5 - error TS18028: Private identifiers are only available when targeting ECMAScript 2015 and higher.

134     #private;
        ~~~~~~~~


Found 1 error.`

danielbankhead:
@DesAWSume the error message explains what to do; simply target es2015+ in TypeScript:
- https://www.typescriptlang.org/tsconfig#target

DesAWSume:
@danielbankhead Thanks Daniel, I would have been fixed it, if it works for me. i tried. all the target versions. it keep tell me the same error. 

here is my tsconfig.json file
`{
  "compilerOptions": {
    "rootDir": ".",
    "outDir": "dist",
    "target": "ES2015",
    "resolveJsonModule": true,
    "composite": true,  // required on the dependency project for references to work
    "esModuleInterop": true,
    "noImplicitAny": false,
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true,
    "moduleResolution": "node"
  },
  "files": [         // by whitelisting the files to include, TS won't automatically
    "package.json"   // include all source below root, which is the default.
  ],
  "include": [
    "test/**/*.ts"
  ],
}
`

DesAWSume:
Oh~. 2nd look. I have a tsconfig under src folder, which my yarn build command will target to it. 
My bad. It fixed for me~

PR #1478, opened 2022-10-21
feat: Introduce environment variable for quota project

sai-sunder-s:

### DIFF
--- a/src/auth/authclient.ts
+++ b/src/auth/authclient.ts
@@ -147,4 +147,31 @@ export abstract class AuthClient
     }
     return headers;
   }
+
+  /**
+   * Gets the quota project id on the credential
+   */
+
+  getQuotaProjectId() {
+    return this.quotaProjectId;
+  }
+
+  /**
+   * Sets the quota project id on the credential
+   * @param quotaProjectId The quota project id to set
+   */
+  setQuotaProjectId(quotaProjectId?: string): void {
+    this.quotaProjectId = quotaProjectId;
+  }
+
+  /**
+   * Sets the quota project id from the GOOGLE_CLOUD_QUOTA_PROJECT
+   * environment variable on the credential
+   */
+  setQuotaProjectIdFromEnvironment(): void {
+    const quotaProjectIdFromEnv = process.env['GOOGLE_CLOUD_QUOTA_PROJECT'];
+    if (quotaProjectIdFromEnv && quotaProjectIdFromEnv.length !== 0) {
+      this.setQuotaProjectId(quotaProjectIdFromEnv);
+    }
+  }
 }
--- a/src/auth/authclient.ts
+++ b/src/auth/authclient.ts
@@ -151,15 +151,4 @@ export abstract class AuthClient
     }
     return headers;
   }
-
-  /**
-   * Sets the quota project id from the GOOGLE_CLOUD_QUOTA_PROJECT
-   * environment variable on the credential
-   */
-  setQuotaProjectIdFromEnvironment(): void {
-    const quotaProjectIdFromEnv = process.env['GOOGLE_CLOUD_QUOTA_PROJECT'];
-    if (quotaProjectIdFromEnv) {
-      this.quotaProjectId = quotaProjectIdFromEnv;
-    }
-  }
 }
--- a/src/auth/authclient.ts
+++ b/src/auth/authclient.ts
@@ -169,7 +169,7 @@ export abstract class AuthClient
    */
   setQuotaProjectIdFromEnvironment(): void {
     const quotaProjectIdFromEnv = process.env['GOOGLE_CLOUD_QUOTA_PROJECT'];
-    if (quotaProjectIdFromEnv && quotaProjectIdFromEnv.length !== 0) {
+    if (quotaProjectIdFromEnv) {
       this.setQuotaProjectId(quotaProjectIdFromEnv);
     }
   }
--- a/src/auth/authclient.ts
+++ b/src/auth/authclient.ts
@@ -151,7 +151,6 @@ export abstract class AuthClient
   /**
    * Gets the quota project id on the credential
    */
-
   getQuotaProjectId() {
     return this.quotaProjectId;
   }
--- a/src/auth/authclient.ts
+++ b/src/auth/authclient.ts
@@ -88,7 +88,11 @@ export abstract class AuthClient
   extends EventEmitter
   implements CredentialsClient
 {
-  protected quotaProjectId?: string;
+  /**
+   * The quota project ID. The quota project can be used by client libraries for the billing purpose.
+   * See {@link https://cloud.google.com/docs/quota| Working with quotas}
+   */
+  quotaProjectId?: string;
   transporter = new DefaultTransporter();
   credentials: Credentials = {};
   projectId?: string | null;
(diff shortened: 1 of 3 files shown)


src/auth/googleauth.ts
```diff
@@ -303,14 +303,14 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
   ): Promise<ADCResponse> {
     // If we've already got a cached credential, just return it.
     if (this.cachedCredential) {
+      this.cachedCredential.setQuotaProjectIdFromEnvironment();
```
alexander-fenster, line 306: Is this needed? If it's already cached, it should be returned as is (at least this is my understanding).

sai-sunder-s: I was on the fence for this. If we are not updating the cached credential, then the user will be confused why setting the environment variable didn't work (if they set the env var in an existing workload).
But I am ok to remove the change to cached cred as well.

src/auth/googleauth.ts
```diff
@@ -322,10 +322,8 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
       } else if (credential instanceof BaseExternalAccountClient) {
         credential.scopes = this.getAnyScopes();
       }
-      this.cachedCredential = credential;
-      projectId = await this.getProjectIdOptional();
 
-      return {credential, projectId};
+      return this.prepareADCResponse(credential);
```
alexander-fenster, line 326: Please do `return await` since `prepareADCResponse` is async. It does not change the behavior in the current code, but the code above does several `await` s so it will be in line with the remaining code. Also, if we ever add `try ... catch` here, `return` of the promise without awaiting will backfire.

src/auth/googleauth.ts
```diff
@@ -338,9 +336,7 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
       } else if (credential instanceof BaseExternalAccountClient) {
         credential.scopes = this.getAnyScopes();
       }
-      this.cachedCredential = credential;
-      projectId = await this.getProjectIdOptional();
-      return {credential, projectId};
+      return this.prepareADCResponse(credential);
```
alexander-fenster, line 339: Same as above

src/auth/googleauth.ts
```diff
@@ -365,9 +361,17 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
     // For GCE, just return a default ComputeClient. It will take care of
     // the rest.
     (options as ComputeOptions).scopes = this.getAnyScopes();
-    this.cachedCredential = new Compute(options);
-    projectId = await this.getProjectIdOptional();
-    return {projectId, credential: this.cachedCredential};
+    return this.prepareADCResponse(new Compute(options));
```
alexander-fenster, line 364: Same as above

src/auth/googleauth.ts
```diff
@@ -365,9 +361,17 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
+  }
+
+  private async prepareADCResponse(
+    credential: JSONClient | Impersonated | Compute | T
+  ): Promise<ADCResponse> {
+    this.cachedCredential = credential;
+    const projectId = await this.getProjectIdOptional();
+    credential.setQuotaProjectIdFromEnvironment();
```
alexander-fenster, line 372: For clarify, let's move this before `this.cachedCredential` assignment (even if it does not change the behavior).

alexander-fenster (approved):
LGTM with some comments. Please wait for Daniel's review before submitting.

src/auth/googleauth.ts
sai-sunder-s: Sure can add for consistency and future proofing. I skipped adding await because caller will be doing an await anyway.

src/auth/googleauth.ts
danielbankhead: `return await this.prepareADCResponse(cachedCredential);` would look pretty clear here.

It would unify the logic and make things consistent in this method - I think in the case where the quota has been set or changed after caching customers would want it more likely than not.

src/auth/googleauth.ts
```diff
@@ -365,9 +357,17 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
-    projectId = await this.getProjectIdOptional();
-    return {projectId, credential: this.cachedCredential};
+    return await this.prepareADCResponse(new Compute(options));
+  }
+
+  private async prepareADCResponse(
+    credential: JSONClient | Impersonated | Compute | T
+  ): Promise<ADCResponse> {
+    const projectId = await this.getProjectIdOptional();
+    credential.setQuotaProjectIdFromEnvironment();
```
danielbankhead, line 367: On second thought, this might be a bit problematic; say you explicitly set `quotaProjectId` in code, but this call constantly unsets your preference.

src/auth/authclient.ts
```diff
@@ -147,4 +147,30 @@ export abstract class AuthClient
...
+  /**
+   * Sets the quota project id from the GOOGLE_CLOUD_QUOTA_PROJECT
+   * environment variable on the credential
+   */
+  setQuotaProjectIdFromEnvironment(): void {
+    const quotaProjectIdFromEnv = process.env['GOOGLE_CLOUD_QUOTA_PROJECT'];
+    if (quotaProjectIdFromEnv) {
+      this.setQuotaProjectId(quotaProjectIdFromEnv);
+    }
+  }
```
danielbankhead, line 175: Going along with the other two comments (removing the get/set and resetting `quotaProjectId` to the env on every call), perhaps this method should be removed and the `env` variable be the default for `quotaProjectId`:
```
quotaProjectId = process.env['GOOGLE_CLOUD_QUOTA_PROJECT'] || '';
```

danielbankhead (commented):
Looks good, a few more things to consider after further review

src/auth/googleauth.ts
sai-sunder-s: For that case, how about we introduce a parameter to the adc method that provides a quota project id explicitly and we will prefer that over env var. This is the way python does.

src/auth/authclient.ts
sai-sunder-s: If we do this way instead of setting from ADC method, then in some creds, this value could be overriden by a value from the json file.

danielbankhead: Since our preference is for the environment variable over the JSON file when they're both available, then we should check to see if it exists before using the JSON input. Fortunately, there are just a few references to make this check (as seen on the left side bar):

Screen Shot 2022-10-26 at 5 58 05 PM

src/auth/googleauth.ts
danielbankhead: That seems like a user-friendly flow: if a quota id is passed explicitly, prefer it (for this one-time operation), otherwise use the credential's current quota id.

src/auth/authclient.ts
sai-sunder-s: Currently we are using environment variable only when the credential is returned through ADC. The priority order might be different when the user is explicitly creating the credential (including loading from a json file). So it will be good to restrict this change within scope of ADC. For more details, checkout [go/quota-env](https://docs.google.com/document/d/1_ZwNyfzhp4AkYm3KCfuPpAtACejVQgaGG9mY7daXmYU/edit?resourcekey=0-uhZZSFONLMSXvAQtXF2BnQ#bookmark=id.tu1rns2ty0fa)

src/auth/googleauth.ts
```diff
@@ -365,9 +363,25 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
+    );
+  }
+
+  private async prepareAndCacheADC(
+    credential: JSONClient | Impersonated | Compute | T,
+    quotaProjectIdOverride?: string
+  ): Promise<ADCResponse> {
+    const projectId = await this.getProjectIdOptional();
+
+    if (quotaProjectIdOverride && quotaProjectIdOverride !== '') {
```
danielbankhead, line 378: No need for the second part of this `if` statement.

test/test.googleauth.ts
```diff
@@ -1043,6 +1044,35 @@ describe('googleauth', () => {
...
+      mockLinuxWellKnownFile(
+        './test/fixtures/config-with-quota/.config/gcloud/application_default_credentials.json'
+      );
+      mockEnvVar('GOOGLE_CLOUD_QUOTA_PROJECT', '');
+      const result = await auth.getApplicationDefault();
+      const client = result.credential as JWT;
+      assert.strictEqual('my-quota-project', client.quotaProjectId);
+    });
+
+    it('getApplicationDefault should use quota project id from file if environment variable is not set', async () => {
```
danielbankhead, line 1067: We’ll need 1 more test where we have an env variable, get the ADC, set the quota project, then receive the ADC again

danielbankhead (commented):
Looks good; just need 1 more test.

test/test.googleauth.ts
sai-sunder-s: Updated existing test to cover this case

danielbankhead (approved):
Looks good to me!

PR #1486, opened 2022-11-08
refactor: Validate AWS Metadata URLs

danielbankhead:
Notes:
- Validates a URL's `hostname` rather than `host`
  - Important for URLs with ports
- Added test for URL with port
- Uses static variables for AWS metadata IP Addresses
- `RangeError` rather than `Error`
- Separated validators from setters in `constructor`

### DIFF
--- a/test/test.awsclient.ts
+++ b/test/test.awsclient.ts
@@ -197,14 +197,31 @@ describe('AwsClient', () => {
           credential_source: invalidCredentialSource,
         };
 
-        assert.throws(() => {
-          return new AwsClient(invalidOptions);
-        }, expectedError);
+        assert.throws(() => new AwsClient(invalidOptions), expectedError);
       });
     });
 
-    it('should throw when an unsupported url is provided', () => {
-      const expectedError = new Error('Invalid host "baddomain.com" for "url"');
+    it('should support credential_source with a port number', () => {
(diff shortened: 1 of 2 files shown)


test/test.awsclient.ts
```diff
@@ -197,14 +197,31 @@ describe('AwsClient', () => {
...
+        assert.throws(() => new AwsClient(invalidOptions), expectedError);
       });
     });
 
-    it('should throw when an unsupported url is provided', () => {
-      const expectedError = new Error('Invalid host "baddomain.com" for "url"');
+    it('should support credential_source with a port number', () => {
+      const validCredentialSource = {...awsCredentialSource};
+      const validURLWithPort = new URL(validCredentialSource.url);
+      validURLWithPort.port = '8888';
```
sai-sunder-s, line 207: I am not able to find any official aws documentatiton that says what ports are allowed. @lsirac any idea?

danielbankhead: I didn't see anything either; perhaps we can verify the host for now and if customers set a port we can allow it for now?

lsirac: I think that works 👍

lsirac (approved):

sai-sunder-s (approved):

PR #1487, opened 2022-11-10
chore(issue-1485): clarify audience + url example usage

### LINKED ISSUE #1485: Improve docs around fetching ID Tokens (Example for `idtokens-serverless.js` is misleading)
It's been mentioned in a couple [previous](https://github.com/googleapis/google-auth-library-nodejs/issues/1208) [issues](https://github.com/googleapis/google-auth-library-nodejs/issues/1339): the documentation and example code included in [samples/idtokens-serverless.js](https://github.com/googleapis/google-auth-library-nodejs/blob/89daaac8e1230935ed80733ac9c2cd89079e2f04/samples/idtokens-serverless.js) can be misleading, but especially so if you view the example from the Cloud Functions docs at https://cloud.google.com/functions/docs/securing/authenticating#generating_tokens_programmatically (and also for [Cloud Run](https://cloud.google.com/run/docs/authenticating/service-to-service)):

Screen Shot 2022-11-01 at 17 35 37

As you can see in the screenshot above, this code example is presented as something to use to actually generate an ID token itself:
> You can use the following code to generate an ID token.

But this is not true. The example actually shows how to create a client that will _make a request to your Cloud Function for you_, with an ID token that it handles generating.

The same impression could be given when reading the [README](https://github.com/googleapis/google-auth-library-nodejs/tree/89daaac8e1230935ed80733ac9c2cd89079e2f04#fetching-id-tokens) for this library, where it mentions this example code under the section `Fetching ID Tokens`. Again, it's not so much fetching an ID token as it is 'Invoking Resources with ID Tokens'.

There's at least one other issue with this exact interpretation: https://github.com/googleapis/google-auth-library-nodejs/issues/1432

If you want to obtain the ID token yourself (as I did), it seems the most direct way is to create the client and call `client.getRequestHeaders()`, which will return the headers that would be used for the client's requests:
```js
{
  Authorization: "Bearer THE_ID_TOKEN"
}
```

---

I'd also point out that the way the `url` and `targetAudience` parameters are explained is still a bit confusing.

The example does start out stating accurately that "For Cloud Functions, `endpoint` and `audience` should be equal.", but then it gets confusing and shows the same URL for a Cloud Function twice, with different values for the `url` and `targetAudience` variables with confusing definitions (`https://TARGET_AUDIENCE/`?).

I think this section might read a little more clearly as something like:

```js
// For Cloud Functions, the `url` and `audience` should be equal.

// Example (Cloud Functions):
// const url = 'https://project-region-projectid.cloudfunctions.net/myFunction';
// const targetAudience = 'https://project-region-projectid.cloudfunctions.net/myFunction';
```

or some variation of that (`const targetAudience = url;`, etc).

pattishin:

Fixes #1485 🦕

### DIFF
--- a/samples/idtokens-serverless.js
+++ b/samples/idtokens-serverless.js
@@ -45,11 +45,11 @@ function main(
   // [END cloudrun_service_to_service_auth]
 
   // [START functions_bearer_token]
-  
+
   // Cloud Functions uses your function's url as the `targetAudience` value
   // const targetAudience = 'https://project-region-projectid.cloudfunctions.net/myFunction';
   // For Cloud Functions, endpoint (`url`) and `targetAudience` should be equal
-  
+
   // [END functions_bearer_token]
 
   // [START functions_bearer_token]
--- a/samples/idtokens-serverless.js
+++ b/samples/idtokens-serverless.js
@@ -45,13 +45,11 @@ function main(
   // [END cloudrun_service_to_service_auth]
 
   // [START functions_bearer_token]
-  // For Cloud Functions, `endpoint` and `audience` should be equal.
-
-  // Example: https://project-region-projectid.cloudfunctions.net/myFunction
-  // const url = 'https://TARGET_HOSTNAME/TARGET_URL';
-
-  // Example (Cloud Functions): https://project-region-projectid.cloudfunctions.net/myFunction
-  // const targetAudience = 'https://TARGET_AUDIENCE/';
+  
+  // Cloud Functions uses your function's url as the `targetAudience` value
+  // const targetAudience = 'https://project-region-projectid.cloudfunctions.net/myFunction';
+  // For Cloud Functions, endpoint (`url`) and `targetAudience` should be equal
+  
   // [END functions_bearer_token]
 
   // [START functions_bearer_token]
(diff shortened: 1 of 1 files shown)


samples/idtokens-serverless.js
```diff
@@ -45,13 +45,11 @@ function main(
   // [END cloudrun_service_to_service_auth]
 
   // [START functions_bearer_token]
-  // For Cloud Functions, `endpoint` and `audience` should be equal.
 
-  // Example: https://project-region-projectid.cloudfunctions.net/myFunction
-  // const url = 'https://TARGET_HOSTNAME/TARGET_URL';
+  // Cloud Functions uses your function's url as the `targetAudience` value
+  // const targetAudience = 'https://project-region-projectid.cloudfunctions.net/myFunction';
+  // For Cloud Functions, endpoint (`url`) and `targetAudience` should be equal
```
danielbankhead, line 51: I think we still want the `const url` property, otherwise the variable wouldn't be declared for sample:

```suggestion
  // For Cloud Functions, endpoint (`url`) and `targetAudience` should be equal
  // const url = targetAudience;
```

danielbankhead (changes requested):
Looks good; one change request

pattishin:
Appreciate it @danielbankhead! 👍🏼

danielbankhead (approved):

PR #1491, opened 2022-11-18
feat: add api key support

arithmetic1728:
Add api key support. The feature has been tested with the following sample: 

https://github.com/arithmetic1728/nodejs-api-key-sample/blob/main/index.js

```
async function main() { 
    const language = require('@google-cloud/language'); 

    const client = new language.LanguageServiceClient({'apiKey': '<FILL IN THE API KEY>'});
    
    const text = 'Hello, World'; 
    const document = { type: 'PLAIN_TEXT', content: text, }; 
    const analyzeSentimentRequest = { document: document, encodingType: 'UTF8', };
    try { 
        const [response] = await client.analyzeSentiment(analyzeSentimentRequest); 
        const sentiment = response.documentSentiment; 
        console.log(`Text: ${text}`); 
        console.log(`Sentiment score: ${sentiment.score}`); 
        console.log(`Sentiment magnitude: ${sentiment.magnitude}`); 
    } catch (err) { 
        console.log(err) 
    } 
} 

main()
```

### DIFF
--- a/samples/api-key.js
+++ b/samples/api-key.js
@@ -1,0 +1,38 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+/**
+ * Import the GoogleAuth library, and create a new GoogleAuth client.
+ */
+const {GoogleAuth} = require('google-auth-library');
+
+/**
+ * Acquire a client, and make a request to an API that's enabled by default.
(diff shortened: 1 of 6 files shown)


arithmetic1728:
The docs failure doesn't seem to be related to my changes.

samples/api-key.js
```diff
@@ -0,0 +1,38 @@
...
+// limitations under the License.
+
+'use strict';
+
+/**
+ * Import the GoogleAuth library, and create a new GoogleAuth client.
+ */
+const {GoogleAuth} = require('google-auth-library');
+
+/**
```
bcoe, line 22: This does not follow sample guidelines, so could not be embedded on cloud.google.com:

https://googlecloudplatform.github.io/samples-style-guide/#region-tags

bcoe (changes requested):
At a glance, things look good to me, but the sample doesn't follow guidelines. Specifically, there's not region tag, which would mean it can't be embedded on cloud.google.com.

I would use this as an example:

https://github.com/googleapis/google-auth-library-nodejs/blob/main/samples/authenticateImplicitWithAdc.js

danielbankhead (commented):
Looks good - however I'm not seeing any other langs using this region tag. Do you need assistance in this area?

arithmetic1728:
we decided not to implement api key support as there is a workaround for the users, closing the PR.

PR #1493, opened 2022-11-23
fix: do not call metadata server if security creds and region are retrievable through environment vars

lsirac:
Fixes an issue introduced by the AWS IMDSv2 changes where the metadata server is being called even if everything is retrievable through the defined environment variables.

### DIFF
--- a/src/auth/awsclient.ts
+++ b/src/auth/awsclient.ts
@@ -346,10 +346,13 @@ export class AwsClient extends BaseExternalAccountClient {
   }
 
   private canRetrieveRegionFromEnvironment(): boolean {
+    // The AWS region can be provided through AWS_REGION or AWS_DEFAULT_REGION.
+    // Only one is required.
     return !!(process.env['AWS_REGION'] || process.env['AWS_DEFAULT_REGION']);
   }
 
   private canRetrieveSecurityCredentialsFromEnvironment(): boolean {
+    // Check if both AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are available.
     return !!(
       process.env['AWS_ACCESS_KEY_ID'] && process.env['AWS_SECRET_ACCESS_KEY']
     );
--- a/src/auth/awsclient.ts
+++ b/src/auth/awsclient.ts
@@ -343,7 +343,9 @@ export class AwsClient extends BaseExternalAccountClient {
   private get regionFromEnv(): string | null {
     // The AWS region can be provided through AWS_REGION or AWS_DEFAULT_REGION.
     // Only one is required.
-    return process.env['AWS_REGION'] || process.env['AWS_DEFAULT_REGION'] || null;
+    return (
+      process.env['AWS_REGION'] || process.env['AWS_DEFAULT_REGION'] || null
+    );
   }
 
   private get securityCredentialsFromEnv(): AwsSecurityCredentials | null {
--- a/src/auth/awsclient.ts
+++ b/src/auth/awsclient.ts
@@ -59,6 +59,15 @@ interface AwsSecurityCredentials {
 }
 
 /**
+ * Interface defining the AWS security-credentials retrieved from environment variables.
+ */
+interface AwsSecurityCredentialsFromEnvironment {
+  accessKeyId: string;
+  secretAccessKey: string;
+  token?: string;
+}
+
+/**
  * AWS external account client. This is used for AWS workloads, where
  * AWS STS GetCallerIdentity serialized signed requests are exchanged for
  * GCP access token.
@@ -101,7 +110,7 @@ export class AwsClient extends BaseExternalAccountClient {
     this.awsRequestSigner = null;
     this.region = '';
 
-    // data validators
+    // Data validators.
     this.validateEnvironmentId();
     this.validateMetadataServerURLs();
   }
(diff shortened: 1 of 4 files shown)


src/auth/awsclient.ts
```diff
@@ -335,4 +337,21 @@ export class AwsClient extends BaseExternalAccountClient {
...
+
+  private shouldUseMetadataServer(): boolean {
+    return (
+      !this.canRetrieveRegionFromEnvironment() ||
+      !this.canRetrieveSecurityCredentialsFromEnvironment()
+    );
+  }
+
+  private canRetrieveRegionFromEnvironment(): boolean {
+    return !!(process.env['AWS_REGION'] || process.env['AWS_DEFAULT_REGION']);
```
aeitzman, line 349: nit: add same comment from java PR - "The AWS region can be provided through AWS_REGION or AWS_DEFAULT_REGION. Only one is required."

src/auth/awsclient.ts
```diff
@@ -335,4 +337,21 @@ export class AwsClient extends BaseExternalAccountClient {
...
+  }
+
+  private canRetrieveSecurityCredentialsFromEnvironment(): boolean {
```
aeitzman, line 352: nit: same here for comment from java PR

aeitzman (approved):

src/auth/awsclient.ts
```diff
@@ -335,4 +337,21 @@ export class AwsClient extends BaseExternalAccountClient {
...
+  private shouldUseMetadataServer(): boolean {
+    return (
+      !this.canRetrieveRegionFromEnvironment() ||
+      !this.canRetrieveSecurityCredentialsFromEnvironment()
+    );
+  }
+
+  private canRetrieveRegionFromEnvironment(): boolean {
+    return !!(process.env['AWS_REGION'] || process.env['AWS_DEFAULT_REGION']);
+  }
```
danielbankhead, line 352: This method would be more useful as a getter than a boolean, and we should use it wherever we currently use this check.

```suggestion
  private get regionFromEnvironment(): string | undefined {
    return process.env['AWS_REGION'] || process.env['AWS_DEFAULT_REGION'];
  }
```

src/auth/awsclient.ts
```diff
@@ -335,4 +337,21 @@ export class AwsClient extends BaseExternalAccountClient {
...
+
+  private canRetrieveSecurityCredentialsFromEnvironment(): boolean {
+    return !!(
+      process.env['AWS_ACCESS_KEY_ID'] && process.env['AWS_SECRET_ACCESS_KEY']
+    );
```
danielbankhead, line 358: Same here, would be more useful as a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get).

src/auth/awsclient.ts
```diff
@@ -168,7 +168,9 @@ export class AwsClient extends BaseExternalAccountClient {
     // Initialize AWS request signer if not already initialized.
     if (!this.awsRequestSigner) {
       const metadataHeaders: Headers = {};
-      if (this.imdsV2SessionTokenUrl) {
+      // Only retrieve the IMDSv2 session token if both the security credentials and region are
+      // not retrievable through the environment.
```
danielbankhead, line 172: Perhaps another sentence as to why we have this here for future team to reference. Some questions we may ask:
- does this break something if we call it?
- does this improve performance?

danielbankhead (changes requested):
Direction and implementation looks good, a requesting a few refactors for description and method reusability.

src/auth/awsclient.ts
```diff
@@ -335,4 +337,21 @@ export class AwsClient extends BaseExternalAccountClient {
     });
     return response.data;
   }
+
+  private shouldUseMetadataServer(): boolean {
```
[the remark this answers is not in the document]
lsirac: The metadata server must be called if either the AWS region or AWS security creds can't be retrieved through the env variables. I don't think this is particularly confusing but open to suggestions. Added this as a comment.

src/auth/awsclient.ts
```diff
@@ -58,6 +58,15 @@ interface AwsSecurityCredentials {
...
 }
 
+/**
+ * Interface defining the AWS security-credentials retrieved from environment variables.
+ */
+interface AwsSecurityCredentialsFromEnvironment {
+  accessKeyId: string;
+  secretAccessKey: string;
+  token?: string;
+}
```
danielbankhead, line 68: Instead of duplicating interfaces, use the imported `AwsSecurityCredentials` interface

src/auth/awsclient.ts
```diff
@@ -335,4 +341,36 @@ export class AwsClient extends BaseExternalAccountClient {
...
+      process.env['AWS_SECRET_ACCESS_KEY']
+    ) {
+      awsSecurityCredentials = {
+        accessKeyId: process.env['AWS_ACCESS_KEY_ID'],
+        secretAccessKey: process.env['AWS_SECRET_ACCESS_KEY'],
+        token: process.env['AWS_SESSION_TOKEN'],
+      };
+    }
+    return awsSecurityCredentials;
+  }
```
danielbankhead, line 375: This could lead to a use before reference error if an environment variable isn't available. Here's an alternative:

```suggestion
  private get securityCredentialsFromEnv(): AwsSecurityCredentials | null {
    // Both AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are required.
    if (
      process.env['AWS_ACCESS_KEY_ID'] &&
      process.env['AWS_SECRET_ACCESS_KEY']
    ) {
      return {
        accessKeyId: process.env['AWS_ACCESS_KEY_ID'],
        secretAccessKey: process.env['AWS_SECRET_ACCESS_KEY'],
        token: process.env['AWS_SESSION_TOKEN'],
      };
    }

    return null;
  }
```

danielbankhead (changes requested):
Super-clean 🧼 2 things

danielbankhead (approved):
Thanks, lgtm

PR #1534, opened 2023-04-04
build: add typings for additional query params for auth

sofisl:
See doc: `New Query Parameters for Auth Requests`

@sai-sunder-s @danielbankhead, please provide any initial reviews/comments. Thanks!

### DIFF
--- a/src/auth/oauth2client.ts
+++ b/src/auth/oauth2client.ts
@@ -301,7 +301,9 @@ export interface GenerateAuthUrlOpts {
    * A way for developers and/or the auth team to provide a set of key value
    * pairs to be added as query parameters to the authorization url.
    */
-  [key: string]: querystring.ParsedUrlQueryInput;
+  [
+    key: string
+  ]: querystring.ParsedUrlQueryInput[keyof querystring.ParsedUrlQueryInput]
 }
 
 export interface AccessTokenResponse {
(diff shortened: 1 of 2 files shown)


sai-sunder-s (approved):

src/auth/oauth2client.ts
```diff
@@ -296,6 +296,12 @@ export interface GenerateAuthUrlOpts {
    * must be used with the 'code_challenge' parameter described above.
    */
   code_challenge?: string;
+
+  /**
+   * A way for developers and/or the auth team to provide a set of key value
+   * pairs to be added as query parameters to the authorization url.
+   */
+  [key: string]: any;
```
danielbankhead, line 304: I think we want more specific types here for the index:

```suggestion
  [key: string]: querystring.ParsedUrlQueryInput;
```

src/auth/oauth2client.ts
```diff
@@ -296,6 +296,12 @@ export interface GenerateAuthUrlOpts {
    * must be used with the 'code_challenge' parameter described above.
    */
   code_challenge?: string;
+
+  /**
+   * A way for developers and/or the auth team to provide a set of key value
+   * pairs to be added as query parameters to the authorization url.
+   */
+  [key: string]: querystring.ParsedUrlQueryInput;
```
danielbankhead, line 304: Apologies, it looks like it's a bit difficult to get to the root types. I tried this locally and it works:

```suggestion
  [
    key: string
  ]: querystring.ParsedUrlQueryInput[keyof querystring.ParsedUrlQueryInput]
```

sofisl: Thank you for checking locally! I didn't even see that it had failed!

danielbankhead (approved):

PR #1547, opened 2023-05-16
deps!: Remove fast-text-encoding as a dependency

### LINKED ISSUE #1546: remove dep on fast-text-encoder
You only load this when needed.
https://github.com/googleapis/google-auth-library-nodejs/blob/85f2c9393f64b97290a0698d3ac6754fa9bc951a/package.json#L23

But i also see that you have a requirement on NodeJS v12+

https://github.com/googleapis/google-auth-library-nodejs/blob/85f2c9393f64b97290a0698d3ac6754fa9bc951a/package.json#L6-L8

therefore this npm package could be removed, cuz text encoder/decoder is available globally since NodeJS v11

Also supporting NodeJS v12 is a bit outdated, think you should target LTS and update to something newer such as v16 that is the current LTS right now

So also think you should do:
```json
 "engines": { 
   "node": ">=16" 
 }, 
```

ddelgrosso1:

Fixes #1546  🦕


ddelgrosso1:
Most modern browsers [now support this natively](https://caniuse.com/textencoder). The browsers that do not are generally 6 or more years old.

danielbankhead (approved):

danielbankhead:
> Most modern browsers [now support this natively](https://caniuse.com/textencoder). The browsers that do not are generally 6 or more years old.

[Agreed](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder).

PR #1582, opened 2023-07-05
chore!: migrate to Node 14

sofisl:


.github/workflows/ci.yaml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [12, 14, 16]
+        node: [14, 16, 18]
```
danielbankhead, line 12: I think we should also include 20 since it’s the current/stable release

sofisl: We'd need to exclude this file from automation, which would just mean we have to manually update it. Are you OK with that?

danielbankhead (approved):

.github/workflows/ci.yaml
danielbankhead: https://github.com/googleapis/google-auth-library-nodejs/pull/1586

PR #1587, opened 2023-07-10
fix: return error code as a string

### LINKED ISSUE #393: bug: update GaxiosError code to include a string or a number
We need to allow either string or number representation for [GaxiosError code](https://github.com/googleapis/gaxios/blob/18a48fc298c8d0e707b7f6e5cd892cb7c239f1db/src/common.ts#L21).

See this PR describing the problem: https://github.com/googleapis/google-auth-library-nodejs/pull/1165

This is the original bug: https://github.com/googleapis/google-auth-library-nodejs/issues/1092

This will be a breaking change, so we need to wait until the next major version to make this change.

sofisl:
Fixes https://github.com/googleapis/gaxios/issues/393


src/transporters.ts
```diff
@@ -102,11 +102,11 @@ export class DefaultTransporter implements Transporter {
         err.message = body.error.errors
           .map((err2: Error) => err2.message)
           .join('\n');
-        err.code = body.error.code;
+        err.code = body.error.code.toString();
         err.errors = body.error.errors;
       } else {
         err.message = body.error.message;
-        err.code = body.error.code || res.status;
+        err.code = body.error.code.toString() || res.status.toString();
```
danielbankhead, line 109: I think keeping the code & status as numbers is more conventional, e.g.:
- https://github.com/googleapis/google-auth-library-nodejs/issues/1092
- https://developer.mozilla.org/en-US/docs/Web/API/Response/status

sofisl: Fair, but it seems like Gaxios expects a string: https://github.com/googleapis/gaxios/blob/18a48fc298c8d0e707b7f6e5cd892cb7c239f1db/src/common.ts#L21

SHould we close this as WAI? https://github.com/googleapis/google-auth-library-nodejs/issues/1092

PR #1653, opened 2023-09-29
docs(GoogleAuth): Document Credential Precedence

### LINKED ISSUE #1085: fix documentation related to authentication strategy
Related to #1084 and #983 

How, and in what order of operations, are we authenticating in this library? We have a document describing how environment variables, constructors, etc., operate together, but what about order of operations within the constructor? For example, where do `keyfile`, `credentials`, and`clientOptions` fit in against the wider order of operations for authentication? When should we be trying to get a projectId?

danielbankhead:
Fixes #1085 🦕

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -173,6 +173,17 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
    */
   static DefaultTransporter = DefaultTransporter;
 
+  /**
+   * Configuration is resolved in the following order of precedence:
+   * - {@link GoogleAuthOptions.credentials `credentials`}
+   * - {@link GoogleAuthOptions.keyFilename `keyFilename`}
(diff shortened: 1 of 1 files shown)


src/auth/googleauth.ts
```diff
@@ -195,8 +206,13 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
 
   /**
    * Obtains the default project ID for the application.
-   * @param callback Optional callback
```
aeitzman, line 198: nit: did you mean to delete the @param + @returns fields or just add after?

aeitzman (approved):

src/auth/googleauth.ts
danielbankhead: Good question - I meant to delete as it could cause confusion (e.g. if using a promise, no need for callback param).

PR #1656, opened 2023-09-29
feat: Expose Compute's Service Account Email

### LINKED ISSUE #1386: Compute should expose serviceAccountEmail
Currently, in class `Compute`: https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/computeclient.ts

The field `serviceAccountEmail` is private and there is no getter. 

This prevents finding the service account for the default credentials when executing on CE:
```tsx
const googleCredentials = await auth.getApplicationDefault()
if (googleCredentials.credential instanceof Compute) {
	// Doesn't compile
	return googleCredentials.credential.serviceAccountEmail
}
```

This is needed to authenticate to PostgreSQL using the service account.

danielbankhead:
This aligns with the Java library:

https://github.com/googleapis/google-auth-library-java/blob/64eed74479944dd596b4228c8324ee3dd3431789/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L473-L483

Fixes #1386 🦕

### DIFF
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -33,7 +33,7 @@ export interface ComputeOptions extends RefreshOptions {
 }
 
 export class Compute extends OAuth2Client {
-  private serviceAccountEmail: string;
+  serviceAccountEmail: string;
   scopes: string[];
 
   /**
--- a/src/auth/computeclient.ts
+++ b/src/auth/computeclient.ts
@@ -33,7 +33,7 @@ export interface ComputeOptions extends RefreshOptions {
 }
 
 export class Compute extends OAuth2Client {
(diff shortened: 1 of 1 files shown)


src/auth/computeclient.ts
```diff
@@ -33,7 +33,7 @@ export interface ComputeOptions extends RefreshOptions {
 }
 
 export class Compute extends OAuth2Client {
-  private serviceAccountEmail: string;
```
aeitzman, line 36: nit: readonly maybe? Looks like the java implementation has a getter but no setter

aeitzman (approved):

PR #1659, opened 2023-10-03
refactor: Replace LRU Dependency With Custom Implementation

### LINKED ISSUE #1368: Retire LRU dependency
The LRU dependency API surface has changed signficnatly between v6 and v7, is it worth migrating to a new version or should we try to implement similar behavior ourselves?

https://github.com/googleapis/google-auth-library-nodejs/pull/1359

danielbankhead:
Fixes #1368 🦕

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -1,0 +1,83 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export interface LRUOptions {
+  /**
+   * The maximum number of items to cache.
+   */
+  capacity: number;
+  /**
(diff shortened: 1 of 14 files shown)


danielbankhead:
`do not merge`: Want to add some tests before moving forward.

src/util.ts
```diff
@@ -0,0 +1,110 @@
...
+ */
+export class LRUCache<T> {
+  readonly capacity: number;
+
+  /**
+   * Maps are in order. Thus, the older item is the first item.
+   *
+   * {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map}
+   */
+  #cache = new Map<string, {lastAccessed: number; value: T}>();
```
aeitzman, line 41: nit: private instead of the # notation, unless I'm not understanding what you are trying to do here.

danielbankhead: `#` is JS-private (true private) vs TS’s type-only private (that can be ignored).

aeitzman: Is there a reason to use them here in particular? Doesn't look like we use the JS-private anywhere else in the code base, also ts style guide says to stick to the typescript private: go/tsjs-style#private-fields (even though that seems to at least partly be for google3 specific reasons)

danielbankhead: In the open source orgs (`google`, `googleapis`, etc.) we mostly have our own rules specified in the [`gts`](https://github.com/google/gts) library (historically the tools to enforce the linting rules weren’t available publicly, iirc). And here I think it has a more accurate intention and allows mangling by bundlers to make smaller packages, while `private` cannot.

aeitzman (approved):

PR #1663, opened 2023-10-06
feat: Unify Base `AuthClient` Options

### LINKED ISSUE #1621: Unable to configure Gaxios options for common AuthClients
**Is your feature request related to a problem? Please describe.**
For OAuth2Client for example, there's no way to configure the agent to make use of keep alive.

 **Describe the solution you'd like**
Ideally, the constructor for all AuthClient sub-classes accept options that would be passed to gaxios, which would allow setting the agent

 **Describe alternatives you've considered**
I think one could maybe overwrite the transporter ivar with one that is sub-classes to set an agent, but that feels bad.

 **Additional context**
Nearly all of the API wrappers in https://github.com/googleapis/google-api-nodejs-client use a similar constructor pattern, wherein all of the constructor options are a subset of gaxios options.

danielbankhead:
Also, expose `Transporter` options

Fixes #1621 🦕


src/auth/baseexternalclient.ts
```diff
@@ -169,9 +160,10 @@ export abstract class BaseExternalAccountClient extends AuthClient {
    */
   constructor(
     options: BaseExternalAccountClientOptions,
-    additionalOptions?: RefreshOptions
+    additionalOptions?: AuthClientOptions
```
aeitzman, line 163: This seems very confusing to me now that the BaseExternalAccountClientOptions object is a child of AuthClientOptions, doesn't seem like it will be clear to users what should be where. It would make sense to me to create a new constructor that takes in just 1 options objects (BaseExternalAccountClientOptions) and does the existing constructor logic, and then keep this constructor for backwards compatibility but just have it pass the merged options objects to the new constructor like you are calling the super here. And then have comments explaining that the old constructor is deprecated/only for compatability

danielbankhead: I wish I was around for the naming 😅 `BaseExternalAccountClient` is the base of external, non-Google Cloud accounts (AWS, Azure, etc.). Since all `AuthClient`s are required to have `transporter`, `projectId`, `quotaProjectId`, and now `universeDomain` properties it would make since for this to be handled in the root class.

src/auth/authclient.ts
```diff
@@ -13,31 +13,91 @@
...
    */
-  forceRefreshOnFailure: boolean;
+  universe_domain: string;
+}
+
+/**
+ * Base Auth Client configuration
+ */
+export type AuthClientOptions = Partial<
+  OriginalAndCamel<AuthJSONOptions> & {
```
aeitzman, line 48: So this options object will end up having each of the AuthJSONOptions twice? Once with key in camelCase and one with the key in snake_case?

src/auth/authclient.ts
```diff
@@ -88,16 +148,40 @@ export abstract class AuthClient
...
-  eagerRefreshThresholdMillis = 5 * 60 * 1000;
+  eagerRefreshThresholdMillis = DEFAULT_EAGER_REFRESH_THRESHOLD_MILLIS;
   forceRefreshOnFailure = false;
+  universeDomain = DEFAULT_UNIVERSE;
+
+  constructor(opts: AuthClientOptions = {}) {
+    super();
+
+    // Shared auth options
+    this.projectId = getOpt(opts, 'project_id') ?? null;
```
aeitzman, line 167: Why alias the getOriginalOrCamel here?  Made it more confusing for me to figure out what was actually going on here.

src/auth/idtokenclient.ts
```diff
@@ -41,7 +46,7 @@ export class IdTokenClient extends OAuth2Client {
    * See: https://developers.google.com/compute/docs/authentication
    */
   constructor(options: IdTokenOptions) {
-    super();
+    super(options);
```
aeitzman, line 49: This is new functionality right? Looks like the IdTokenClient didn't previously support RefreshClient options. Are we sure it should be used here?

test/test.util.ts
```diff
@@ -50,18 +61,23 @@ describe('util', () => {
...
       lru.set('first', 1);
       lru.set('second', 2);
 
-      await new Promise(res => setTimeout(res, maxAge + 1));
+      // back to the future 🏎️
+      sandbox.clock.tick(maxAge + 1);
 
+      // just set, so should be fine
       lru.set('third', 3);
+      assert.equal(lru.get('third'), 3);
```
aeitzman, line 76: Is this intended to be part of this PR?

samples/keepalive.js
```diff
@@ -30,21 +30,26 @@ const https = require('https');
  * Acquire a client, and make a request to an API that's enabled by default.
  */
 async function main() {
+  // create a new agent with keepAlive enabled
```
aeitzman, line 33: nit: Add period on the comment

src/auth/baseexternalclient.ts
aeitzman: Just realized we can't have multiple constructors in typescript so what I was suggesting doesn't make sense anyway. I agree that it make sense to handle it all in the base class. My concern is that having both options objects will be confusing to users since they are both AuthClientOptions. At the very least can you add to the constructor documentation that the additional options param is deprecated and the options can all be passed via the BaseExternalAccountClientOptions parameter? Other option is if we just wait to release this until the breaking change suggested by  #1672 then we won't be in this weird in between state at all.

danielbankhead: Docs-deprecated `additionalOptions`, as the [@deprecated](https://tsdoc.org/pages/tags/deprecated/) tag is considered breaking as it would produce build warnings.

src/auth/authclient.ts
```diff
@@ -88,16 +148,40 @@ export abstract class AuthClient
...
+    this.quotaProjectId = getOpt(opts, 'quota_project_id');
+    this.credentials = getOpt(opts, 'credentials') ?? {};
+    this.universeDomain = getOpt(opts, 'universe_domain') ?? DEFAULT_UNIVERSE;
+
+    // Shared client options
+    this.transporter = opts.transporter ?? new DefaultTransporter();
```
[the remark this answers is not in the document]
danielbankhead: A `Transporter` is an interface, which `gaxios` satisfies by having a `request` method with a matching signature. I think this makes it a lot easier for customers that simply want to pass the default `gaxios` or custom `Gaxios` instance w/o building/maintaining a transporter.

src/auth/authclient.ts
danielbankhead: Yep, with type hints and documentation. This way we can pass in the raw JSON properties from a file as-is or provide the options in a conventional JS pattern.

src/auth/authclient.ts
danielbankhead: That's fair, unaliased it

test/test.util.ts
danielbankhead: Accident; somehow the target wasn't `main`: #1667

src/auth/idtokenclient.ts
danielbankhead: I thought it would be cleaner to pass the parameters directly to it's parent; however we mostly care about setting `AuthClientOptions`. I figured it wouldn't hurt to expose underlying `OAuth2ClientOptions`, as it wouldn't harm `IdTokenClient`'s functionality. Thoughts?

Also, I noticed the `OAuth2ClientOptions` interface was accidentally duplicated from a merge; removed the duplicated interface to avoid confusion.

aeitzman (approved):

PR #1692, opened 2023-11-10
feat: Retrieve `universe_domain` for Compute clients

danielbankhead:
Dependencies:
- https://github.com/googleapis/gcp-metadata/pull/596
- https://github.com/googleapis/gcp-metadata/pull/598
- https://github.com/googleapis/gcp-metadata/pull/601

🦕

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -931,17 +931,19 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
     if (client instanceof BaseExternalAccountClient) {
       const serviceAccountEmail = client.getServiceAccountEmail();
       if (serviceAccountEmail) {
-        return {client_email: serviceAccountEmail};
+        return {
+          client_email: serviceAccountEmail,
+          universe_domain: client.universeDomain,
+        };
       }
     }
 
     if (this.jsonContent) {
-      const credential: CredentialBody = {
+      return {
         client_email: (this.jsonContent as JWTInput).client_email,
         private_key: (this.jsonContent as JWTInput).private_key,
         universe_domain: this.jsonContent.universe_domain,
       };
-      return credential;
     }
 
     if (await this._checkIsGCE()) {
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -53,6 +53,7 @@ import {
   ExternalAccountAuthorizedUserClient,
   ExternalAccountAuthorizedUserClientOptions,
 } from './externalAccountAuthorizedUserClient';
+import {originalOrCamelOptions} from '../util';
 
 /**
  * Defines all types of explicit clients that are determined via ADC JSON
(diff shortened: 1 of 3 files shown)


src/auth/googleauth.ts
```diff
@@ -131,6 +132,14 @@ const GoogleAuthExceptionMessages = {
     'Unable to detect a Project Id in the current environment. \n' +
     'To learn more about authentication and Google APIs, visit: \n' +
     'https://cloud.google.com/docs/authentication/getting-started',
+  NO_CREDENTIALS_FOUND:
+    'Unable to find credentials in current environment. \n' +
+    'To learn more about authentication and Google APIs, visit: \n' +
+    'https://cloud.google.com/docs/authentication/getting-started',
+  NO_UNIVERSE_DOMAIN_FOUND:
+    'Unable to detect a Universe Domain in the current environment.\n' +
```
aeitzman, line 140: nit: I think this message could be a little clearer - the link it points to doesn't mention universe domains anywhere, not sure if that is something that will be updated in the near future. Otherwise, maybe just make it clear that the universe domains is meant to be retrieved via the metadata keys assuming that's what the user would need to set up.

src/auth/googleauth.ts
```diff
@@ -286,6 +302,42 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
+  async #getUniverseFromMetadataServer() {
+    if (!(await this._checkIsGCE())) return;
+
+    let universeDomain: string;
+
+    try {
+      universeDomain = await gcpMetadata.universe('universe_domain');
+      universeDomain ||= DEFAULT_UNIVERSE;
+    } catch (e) {
+      if (e instanceof GaxiosError && e.status === 404) {
```
aeitzman, line 314: What is the situation that we would expect a 404 here? Are there any other error codes we would want to ignore and just use default universe?

src/auth/googleauth.ts
```diff
@@ -286,6 +302,42 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
...
+   * - The universe domain in {@link GoogleAuth.clientOptions}
+   * - {@link gcpMetadata.universe}
+   *
+   * @returns The universe domain
+   */
+  async getUniverseDomain(): Promise<string> {
+    this.#universeDomain ??= originalOrCamelOptions(this.clientOptions).get(
+      'universe_domain'
+    );
+    this.#universeDomain ??= await this.#getUniverseFromMetadataServer();
```
aeitzman, line 336: Would we not want to override value retrieved from the metadata server if someone explicitly sets universe domain via the options object? If not, I think that should be documented

src/auth/googleauth.ts
danielbankhead: Once the `universe` predefined key has been made available it will be linked there

src/auth/googleauth.ts
danielbankhead: (Provided background offline)

src/auth/googleauth.ts
danielbankhead: The documentation for this method mentions this precedence

aeitzman (approved):

matt-oakes:
This seems to have caused an issue in my Firebase functions which use this library. Details here https://github.com/googleapis/google-auth-library-nodejs/issues/1706

PR #1745, opened 2024-01-31
fix: Universe Domain Resolution

danielbankhead:
Additionally:
- Expose `getUniverseDomainFromMetadataServer`

🦕

### DIFF
--- a/src/auth/googleauth.ts
+++ b/src/auth/googleauth.ts
@@ -620,7 +620,8 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
     }
 
     // Create source client for impersonation
-    const sourceClient = new UserRefreshClient(json.source_credentials);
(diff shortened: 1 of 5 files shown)


src/auth/googleauth.ts
```diff
@@ -672,6 +667,10 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
   ): JSONClient {
     let client: JSONClient;
 
+    // user's preferred universe domain
+    const preferredUniverseDomain =
```
danielbankhead, line 671: Not a huge fan of this pattern, however the existing `.fromJSON` pattern makes it difficult to add preferred options. We should consider a refactor to `[AuthClient].fromJSON(json, overrrides)` later

alexander-fenster (approved):

PR #1762, opened 2024-02-27
fix!: Update OAuth2Client Endpoints to JWT Versions

danielbankhead:
The PEM endpoints have been long-deprecated and are causing issues for customers.

As a dependency, added `JWK` support to `NodeCrypto`.

Related:
- googleapis/google-cloud-node-core#514
- https://github.com/googleapis/node-gtoken/pull/466

Fixes: googleapis/google-cloud-node-core#514 🦕


danielbankhead:
Requires Node 16+.

PR #1765, opened 2024-03-05
fix: making aws request signer get a new session token each time 

### LINKED ISSUE #1755: Workload Identity with AWS & IMDSv2 use expired token
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

#### Environment details

  - OS: Linux
  - Node.js version: 16.20.2 (I know it's EOL..)
  - npm version: 8.19.4
  - `google-auth-library` version: 9.4.0

#### Steps to reproduce

  1. Setup federation with AWS according to https://cloud.google.com/iam/docs/workload-identity-federation-with-other-clouds . The JSON file should have `imdsv2_session_token_url`.
  2. Observe that API calls to Google (we're testing with [service account signJwt](https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signJwt)) are working normally
  3. Wait about an hour
  4. Observe that API calls are now failing

The error object we receive is:

```
GaxiosError
    at Gaxios._request (/app/node_modules/gaxios/build/src/gaxios.js:142:23)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async AwsClient.getAwsRoleName (/app/node_modules/google-auth-library/build/src/auth/awsclient.js:217:26)
    at async AwsRequestSigner.getCredentials (/app/node_modules/google-auth-library/build/src/auth/awsclient.js:110:34)
    at async AwsRequestSigner.getRequestOptions (/app/node_modules/google-auth-library/build/src/auth/awsrequestsigner.js:66:40)
    at async AwsClient.retrieveSubjectToken (/app/node_modules/google-auth-library/build/src/auth/awsclient.js:125:25)
    at async AwsClient.refreshAccessTokenAsync (/app/node_modules/google-auth-library/build/src/auth/baseexternalclient.js:288:30)
    at async AwsClient.getAccessToken (/app/node_modules/google-auth-library/build/src/auth/baseexternalclient.js:165:13)
    [scrubbed] {
  config: {
    url: 'http://169.254.169.254/latest/meta-data/iam/security-credentials',
    method: 'GET',
    responseType: 'text',
    headers: {
      'x-aws-ec2-metadata-token': '<private>',
      'User-Agent': 'google-api-nodejs-client/9.6.1',
      'x-goog-api-client': 'gl-node/16.20.2'
    },
    paramsSerializer: [Function: paramsSerializer],
    validateStatus: [Function: validateStatus],
    errorRedactor: [Function: defaultErrorRedactor]
  },
  response: {
    config: {
      url: 'http://169.254.169.254/latest/meta-data/iam/security-credentials',
      method: 'GET',
      responseType: 'text',
      headers: [Object],
      paramsSerializer: [Function: paramsSerializer],
      validateStatus: [Function: validateStatus],
      errorRedactor: [Function: defaultErrorRedactor]
    },
    data: '',
    headers: {
      connection: 'close',
      'content-length': '0',
      'content-type': 'text/plain',
      date: 'Wed, 07 Feb 2024 05:38:23 GMT',
      server: 'EC2ws'
    },
    status: 401,
    statusText: 'Unauthorized',
    request: {
      responseURL: 'http://169.254.169.254/latest/meta-data/iam/security-credentials'
    }
  },
  error: undefined,
  status: 401,
  [Symbol(gaxios-gaxios-error)]: '6.2.0'
}
```

I think this is because in https://github.com/googleapis/google-auth-library-nodejs/blob/3b19e9cfa0e7ca4ffd97fa0ebd96f065286573dc/src/auth/awsclient.ts#L151-L161 the IMDSv2 API is only called once when `this.awsRequestSigner` is null, then is reused for all subsequent requests.

However in `getImdsV2SessionToken` the `x-aws-ec2-metadata-token-ttl-seconds` is only configured to be 300s:

https://github.com/googleapis/google-auth-library-nodejs/blob/3b19e9cfa0e7ca4ffd97fa0ebd96f065286573dc/src/auth/awsclient.ts#L242

so after the application run for some times the GCP token expires and refreshing of the token fails because it use expired IMDSv2 token.

Semi-related but I checked how the Go SDK has implemented and it seems to be 1:1 match to this behavior and may also have similar problem https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.17.0:google/internal/externalaccount/aws.go;l=304

aeitzman:
fixes #1755

### DIFF
--- a/src/auth/awsclient.ts
+++ b/src/auth/awsclient.ts
@@ -155,7 +155,7 @@ export class AwsClient extends BaseExternalAccountClient {
       // The credential config contains all the URLs by default but clients may be running this
       // where the metadata server is not available and returning the credentials through the environment.
       // Removing this check may break them.
-      if (this.shouldUseMetadataServer() && this.imdsV2SessionTokenUrl) {
+      if (!this.regionFromEnv && this.imdsV2SessionTokenUrl) {
         metadataHeaders['x-aws-ec2-metadata-token'] =
           await this.getImdsV2SessionToken();
       }
@@ -166,6 +166,10 @@ export class AwsClient extends BaseExternalAccountClient {
         // https://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html
         if (this.securityCredentialsFromEnv) {
           return this.securityCredentialsFromEnv;
+        }
+        if (this.imdsV2SessionTokenUrl) {
+          metadataHeaders['x-aws-ec2-metadata-token'] =
+            await this.getImdsV2SessionToken();
         }
         // Since the role on a VM can change, we don't need to cache it.
         const roleName = await this.getAwsRoleName(metadataHeaders);
@@ -314,12 +318,6 @@ export class AwsClient extends BaseExternalAccountClient {
         headers: headers,
       });
     return response.data;
-  }
-
-  private shouldUseMetadataServer(): boolean {
-    // The metadata server must be used when either the AWS region or AWS security
-    // credentials cannot be retrieved through their defined environment variables.
-    return !this.regionFromEnv || !this.securityCredentialsFromEnv;
   }
 
   private get regionFromEnv(): string | null {
--- a/test/test.awsclient.ts
+++ b/test/test.awsclient.ts
@@ -330,6 +330,7 @@ describe('AwsClient', () => {
             reqheaders: {'x-aws-ec2-metadata-token-ttl-seconds': '300'},
           })
             .put('/latest/api/token')
+            .twice()
             .reply(200, awsSessionToken)
         );
 


danielbankhead (approved):

src/auth/awsclient.ts
```diff
@@ -155,7 +155,7 @@ export class AwsClient extends BaseExternalAccountClient {
       // The credential config contains all the URLs by default but clients may be running this
       // where the metadata server is not available and returning the credentials through the environment.
       // Removing this check may break them.
-      if (this.shouldUseMetadataServer() && this.imdsV2SessionTokenUrl) {
+      if (!this.regionFromEnv && this.imdsV2SessionTokenUrl) {
```
lsirac, line 158: If region is provided by the environment, but security credentials are not, we still need to get a session token and call MDS. I see that it still works because you've added a new check below, but this makes this function more confusing to me.

Can we keep the original function (shouldUseMetadataServer()) and make other changes? e.g. not sure if there is any benefit of not creating a new request signer on each call - doing that would simply things.

aeitzman: This is all going to get refactored in the next month or so anyway (when the suppliers get added as part of the programmatic auth change), I was just making this small change as a temporary fix since it seems to be blocking some people. I would rather worry about it then vs making this change larger, let me know what you think

lsirac (approved):

Issue #1766, opened 2024-03-05
Allow passing in AWS credentials instead of only using environment variables


aliciazavala:
**Is your feature request related to a problem? Please describe.**
Yes, the current implementation of the library only allows for AWS credentials to be passed via environment variables or metadata endpoint. This can be limiting in scenarios where the user wants to pass credentials dynamically or when they want to use different credentials for different services.

 **Describe the solution you'd like**
I would like the library to be updated to allow AWS credentials to be passed in as parameters to the relevant functions or methods. This would provide more flexibility and control to the user over which credentials to use.

 **Describe alternatives you've considered**
An alternative could be to allow the user to set the credentials in a configuration file that the library can read from. However, this might not be as flexible as passing the credentials directly to the functions or methods.

 **Additional context**
This feature would be particularly useful in multi-account AWS setups, where different services might be running under different accounts. Being able to specify the credentials at the function or method level would make it easier to manage such setups.

Here's a rough idea of what I'm suggesting:
https://github.com/googleapis/google-auth-library-nodejs/blob/6a6e49634863f61487688724d0d20632e03f0299/src/auth/awsclient.ts#L90-L108

```
 constructor(
    options: AwsClientOptions,
    additionalOptions?: AuthClientOptions,
    awsCredentials?: AwsSecurityCredentials,
  ) {
    super(options, additionalOptions);
    this.environmentId = options.credential_source.environment_id;
    // This is only required if the AWS region is not available in the
    // AWS_REGION or AWS_DEFAULT_REGION environment variables.
    this.regionUrl = options.credential_source.region_url;
    // This is only required if AWS security credentials are not available in
    // environment variables.
    this.securityCredentialsUrl = options.credential_source.url;
    this.regionalCredVerificationUrl =
      options.credential_source.regional_cred_verification_url;
    this.imdsV2SessionTokenUrl =
      options.credential_source.imdsv2_session_token_url;
    this.awsRequestSigner = null;
    this.region = '';
    this.credentialSourceType = 'aws';
    this.awsCredentials = awsCredentials || null;
```

and the `retrieveSubjectToken()` function 
https://github.com/googleapis/google-auth-library-nodejs/blob/6a6e49634863f61487688724d0d20632e03f0299/src/auth/awsclient.ts#L164-#L169

```
      this.awsRequestSigner = new AwsRequestSigner(async () => {
        // Check provided credentials first
        if (this.awsCredentials) {
          return this.awsCredentials;
        }
[...]

PR #1770, opened 2024-03-11
feat: Allow passing AWS creds

aliciazavala:
This feature would be particularly useful in multi-account AWS setups, where different services might be running under different accounts. Being able to specify the credentials at the function or method level would make it easier to manage such setups.

Fixes #[1766](https://github.com/googleapis/google-auth-library-nodejs/issues/1766) 🦕


lsirac:
Hi @aliciazavala, we are going to add the ability to provide AWS credentials soon. It will be in a way that enables refresh. Please keep an eye out for it in the coming weeks!

PR #1771, opened 2024-03-11
feat: `PassThrough` AuthClient

danielbankhead:
An `AuthClient` without any Authentication information. Useful for:
- Anonymous access
- Local Emulators
- Testing Environments

In other languages:
- https://github.com/googleapis/google-auth-library-php/pull/208
- https://github.com/googleapis/google-auth-library-python/pull/206

Unblocks:
- https://github.com/googleapis/nodejs-storage/issues/2400

🦕

### DIFF
--- a/src/auth/passthrough.ts
+++ b/src/auth/passthrough.ts
@@ -1,0 +1,60 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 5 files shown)


src/auth/passthrough.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+   * A required method of the base class.
+   * Always will return an empty object.
+   *
+   * @returns {}
+   */
+  async getAccessToken() {
+    return {};
```
alexander-fenster, line 48: should we write it as `{} as GetAccessTokenResponse` to make sure all implementations of `AuthClient` return same types? Otherwise it might be harder to switch implementations in the end user code.

src/auth/passthrough.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+   * A required method of the base class.
+   * Always will return an empty object.
+   *
+   * @returns {}
+   */
+  async getRequestHeaders() {
```
alexander-fenster, line 57: Same comment re: return type.

alexander-fenster (approved):

aeitzman (approved):

src/auth/passthrough.ts
danielbankhead: Updated

TimurSadykov (commented):
passing by LGTM :)

PR #1804, opened 2024-04-30
build: remove `compodoc`

danielbankhead:
It's constantly introducing breaking changes in patch releases and its unclear who's using it. Today its blocking all other PRs in this repo.

Let's drop it.

🦕

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "jws": "^4.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


sofisl (changes requested):
Seems like docs-test and predocs-test use it. Can you update those as well?

danielbankhead:
Opting to pin instead, as it is used by cloud rad

PR #1814, opened 2024-05-18
feat: Adding support of client authentication method.

BigTailWolf:


src/auth/oauth2client.ts
```diff
@@ -68,6 +68,12 @@ export enum CertificateFormat {
   JWK = 'JWK',
 }
 
+export enum ClientAuthentication {
```
aeitzman, line 71: Can you add a comment explaining what this enum is for?

src/auth/oauth2client.ts
```diff
@@ -475,6 +481,11 @@ export interface OAuth2ClientOptions extends AuthClientOptions {
    * The allowed OAuth2 token issuers.
    */
   issuers?: string[];
+  /**
+   * The client_authentication choice from baisc/post/none
```
aeitzman, line 485:
```suggestion
   * The client authentication type. Support values are basic, post, and none.
```

src/auth/oauth2client.ts
```diff
@@ -475,6 +481,11 @@ export interface OAuth2ClientOptions extends AuthClientOptions {
...
+   * https://docs.authlib.org/en/latest/client/oauth2.html#client-authentication
```
aeitzman, line 486: Why is this linking to a specific python implementation? I think we should link to the rfc directly instead ([RFC7591](https://datatracker.ietf.org/doc/html/rfc7591#section-2))

src/auth/oauth2client.ts
```diff
@@ -542,6 +554,8 @@ export class OAuth2Client extends AuthClient {
       oauth2IapPublicKeyUrl: 'https://www.gstatic.com/iap/verify/public_key',
       ...opts.endpoints,
     };
+    this.client_authentication =
+      opts.client_authentication || ClientAuthentication.ClientSecretPost;
```
aeitzman, line 558: Is this supposed to be post? The RFC says the default should be basic.

test/test.oauth2.ts
```diff
@@ -1421,6 +1426,23 @@ describe('oauth2', () => {
       assert.strictEqual(params.client_id, 'overridden');
     });
 
+    it('getToken should use basic header auth if provided in options', async () => {
```
aeitzman, line 1429: can you add tests for the none and post type and a default case?

src/auth/oauth2client.ts
```diff
@@ -660,20 +674,28 @@ export class OAuth2Client extends AuthClient {
     options: GetTokenOptions
   ): Promise<GetTokenResponse> {
     const url = this.endpoints.oauth2TokenUrl.toString();
-    const values = {
+    const headers: any = {'Content-Type': 'application/x-www-form-urlencoded'};
```
aeitzman, line 677: Don't need to use any here, can do something like this instead, same with below

https://github.com/googleapis/google-auth-library-nodejs/blob/7dbedff829be093dad0a2da32f0ab12862d473bc/src/auth/awsrequestsigner.ts#L142C5-L155C6

src/auth/oauth2client.ts
BigTailWolf: The legacy behavior is post for oauth2 endpoint.

src/auth/oauth2client.ts
BigTailWolf: Yeah, let me change into https://datatracker.ietf.org/doc/html/rfc7591#section-2

src/auth/oauth2client.ts
BigTailWolf: Added

src/auth/oauth2client.ts
BigTailWolf: Thanks for the example. When I made this change, I tried something else and none of them pass the generic type check then I left an `any`. Changed upon it.

src/auth/oauth2client.ts
```diff
@@ -475,6 +481,11 @@ export interface OAuth2ClientOptions extends AuthClientOptions {
    * The allowed OAuth2 token issuers.
    */
   issuers?: string[];
+  /**
+   * The client authentication type. Support values are basic, post, and none.
```
aeitzman, line 485:
```suggestion
   * The client authentication type. Supported values are basic, post, and none.
```

test/test.oauth2.ts
BigTailWolf: Yeah, no problem.

src/auth/oauth2client.ts
```diff
@@ -475,6 +485,12 @@ export interface OAuth2ClientOptions extends AuthClientOptions {
    * The allowed OAuth2 token issuers.
    */
   issuers?: string[];
+  /**
+   * The client authentication type. Supported values are basic, post, and none.
+   * If no type is provided, will default to post.
```
lsirac, line 490:
```suggestion
   * Defaults to post if not provided.
```

src/auth/oauth2client.ts
```diff
@@ -491,6 +507,7 @@ export class OAuth2Client extends AuthClient {
   protected refreshTokenPromises = new Map<string, Promise<GetTokenResponse>>();
   readonly endpoints: Readonly<OAuth2ClientEndpoints>;
   readonly issuers: string[];
+  private client_authentication: ClientAuthentication;
```
lsirac, line 510: Looks like this should be readonly if we follow the existing convention

src/auth/oauth2client.ts
```diff
@@ -660,20 +679,30 @@ export class OAuth2Client extends AuthClient {
...
   ): Promise<GetTokenResponse> {
     const url = this.endpoints.oauth2TokenUrl.toString();
-    const values = {
+    const headers: {[key: string]: string} = {
+      'Content-Type': 'application/x-www-form-urlencoded',
+    };
+    if (this.client_authentication === ClientAuthentication.ClientSecretBasic) {
+      const basic_auth =
+        'basic ' + btoa(`${this._clientId}:${this._clientSecret}`);
+      headers.Authorization = basic_auth;
```
lsirac, line 688:
```suggestion
      headers['Authorization'] = basic_auth;
```

src/auth/oauth2client.ts
```diff
@@ -660,20 +679,30 @@ export class OAuth2Client extends AuthClient {
     options: GetTokenOptions
   ): Promise<GetTokenResponse> {
     const url = this.endpoints.oauth2TokenUrl.toString();
-    const values = {
+    const headers: {[key: string]: string} = {
```
lsirac, line 682:
```suggestion
    const headers: Headers = {
```

src/auth/oauth2client.ts
```diff
@@ -660,20 +679,30 @@ export class OAuth2Client extends AuthClient {
...
+      'Content-Type': 'application/x-www-form-urlencoded',
+    };
+    if (this.client_authentication === ClientAuthentication.ClientSecretBasic) {
+      const basic_auth =
+        'basic ' + btoa(`${this._clientId}:${this._clientSecret}`);
+      headers.Authorization = basic_auth;
+    }
+    const values: {[key: string]: string | undefined} = {
```
lsirac, line 690:
```suggestion
    const values = {
```

src/auth/oauth2client.ts
```diff
@@ -660,20 +679,30 @@ export class OAuth2Client extends AuthClient {
...
       code: options.code,
       client_id: options.client_id || this._clientId,
-      client_secret: this._clientSecret,
       redirect_uri: options.redirect_uri || this.redirectUri,
       grant_type: 'authorization_code',
       code_verifier: options.codeVerifier,
     };
+    if (this.client_authentication === ClientAuthentication.ClientSecretPost) {
+      values.client_secret = this._clientSecret;
```
lsirac, line 698:
```suggestion
   Object.assign(values, {client_secret: this._clientSecret});
```

test/test.oauth2.ts
```diff
@@ -1421,6 +1426,68 @@ describe('oauth2', () => {
...
+        client_authentication: ClientAuthentication.ClientSecretBasic,
+      };
+      const oauth2client = new OAuth2Client(opts);
+      const _ = oauth2client.getToken({
+        code: 'code here',
+        client_id: CLIENT_ID,
+      });
+    });
+
+    it('getToken should not use basic header auth if provided none in options', async () => {
```
lsirac, line 1458: Where does the validation happen?

src/auth/oauth2client.ts
```diff
@@ -68,6 +68,16 @@ export enum CertificateFormat {
   JWK = 'JWK',
 }
 
+/**
+ * The client authentication type. Support values are basic, post, and none.
```
aeitzman, line 72:
```suggestion
 * The client authentication type. Supported values are basic, post, and none.
```

src/auth/oauth2client.ts
BigTailWolf: Done

test/test.oauth2.ts
BigTailWolf: Missed validation. Added.

test/test.oauth2.ts
```diff
@@ -1421,6 +1427,70 @@ describe('oauth2', () => {
...
+      const opts = {
+        clientId: CLIENT_ID,
+        clientSecret: CLIENT_SECRET,
+        redirectUri: REDIRECT_URI,
+        endpoints: {
+          oauth2TokenUrl: 'mytokenurl',
+        },
+        client_authentication: ClientAuthentication.None,
+      };
+      const oauth2client = new OAuth2Client(opts);
```
lsirac, line 1474: You're not validating that its not using basic auth if client auth is set to None. This just validates the type, not the actual behavior

test/test.oauth2.ts
```diff
@@ -1421,6 +1427,70 @@ describe('oauth2', () => {
...
+      const opts = {
+        clientId: CLIENT_ID,
+        clientSecret: CLIENT_SECRET,
+        redirectUri: REDIRECT_URI,
+        endpoints: {
+          oauth2TokenUrl: 'mytokenurl',
+        },
+      };
+      const oauth2client = new OAuth2Client(opts);
+      assert(
```
lsirac, line 1488: Same here

lsirac (changes requested):

test/test.oauth2.ts
BigTailWolf: This part is validating the constructor behavior.
Technically we shouldn't use `None` auth either against OAuth2 portal or STS portal.

So shall here be a dummy token url and then throwing an error and validate that error as expected: an Auth should be used?

lsirac: I think validating that there is no client auth in the request in the ways that are supported is enough

src/auth/oauth2client.ts
```diff
@@ -68,6 +68,16 @@ export enum CertificateFormat {
...
 
+/**
+ * The client authentication type. Supported values are basic, post, and none.
+ * https://datatracker.ietf.org/doc/html/rfc7591#section-2
+ */
+export enum ClientAuthentication {
+  ClientSecretPost,
+  ClientSecretBasic,
+  None,
+}
```
danielbankhead, line 79:
```suggestion
export enum ClientAuthentication {
  ClientSecretPost = 'ClientSecretPost',
  ClientSecretBasic = 'ClientSecretBasic',
  None = 'None',
}
```

src/auth/oauth2client.ts
```diff
@@ -491,6 +507,7 @@ export class OAuth2Client extends AuthClient {
   protected refreshTokenPromises = new Map<string, Promise<GetTokenResponse>>();
   readonly endpoints: Readonly<OAuth2ClientEndpoints>;
   readonly issuers: string[];
+  readonly client_authentication: ClientAuthentication;
```
danielbankhead, line 510:
```suggestion
  readonly clientAuthentication: ClientAuthentication;
```

src/auth/oauth2client.ts
```diff
@@ -660,20 +679,34 @@ export class OAuth2Client extends AuthClient {
...
+      const basic_auth =
+        'Basic ' +
+        Buffer.from(`${this._clientId}:${this._clientSecret}`).toString(
+          'base64'
+        );
+      headers['Authorization'] = basic_auth;
+    } else if (
+      this.client_authentication === ClientAuthentication.ClientSecretPost
+    ) {
+      Object.assign(values, {client_secret: this._clientSecret});
```
danielbankhead, line 702:
```suggestion
      values.client_secret = this._clientSecret;
```

src/auth/oauth2client.ts
```diff
@@ -660,20 +679,34 @@ export class OAuth2Client extends AuthClient {
...
+    }
     const res = await this.transporter.request<CredentialRequest>({
       ...OAuth2Client.RETRY_CONFIG,
       method: 'POST',
       url,
       data: querystring.stringify(values),
-      headers: {'Content-Type': 'application/x-www-form-urlencoded'},
+      headers: headers,
```
danielbankhead, line 709:
```suggestion
      headers,
```

src/auth/oauth2client.ts
```diff
@@ -660,20 +679,34 @@ export class OAuth2Client extends AuthClient {
     options: GetTokenOptions
   ): Promise<GetTokenResponse> {
     const url = this.endpoints.oauth2TokenUrl.toString();
+    const headers: Headers = {
+      'Content-Type': 'application/x-www-form-urlencoded',
+    };
     const values = {
```
danielbankhead, line 685: You can add `grant_type` to the `GetTokenOptions` type.

```suggestion
    const values: Omit<GetTokenOptions, 'codeVerifier'> = {
```

src/auth/oauth2client.ts
```diff
@@ -660,20 +679,34 @@ export class OAuth2Client extends AuthClient {
...
       code: options.code,
       client_id: options.client_id || this._clientId,
-      client_secret: this._clientSecret,
       redirect_uri: options.redirect_uri || this.redirectUri,
       grant_type: 'authorization_code',
       code_verifier: options.codeVerifier,
```
danielbankhead, line 690:
```suggestion
      grant_type: options.grant_type || 'authorization_code',
      code_verifier: options.codeVerifier || options.code_verifier,
```

src/auth/oauth2client.ts
```diff
@@ -68,6 +68,16 @@ export enum CertificateFormat {
...
+ */
+export enum ClientAuthentication {
+  ClientSecretPost,
+  ClientSecretBasic,
+  None,
+}
+
 export interface GetTokenOptions {
   code: string;
   codeVerifier?: string;
```
danielbankhead, line 83:
```suggestion
  codeVerifier?: string;
  /**
  * @alias GetTokenOptions['codeVerifier']
  **/
  code_verifier?: string;
  grant_type?: 'authorization_code';
```

test/test.oauth2.ts
```diff
@@ -1421,6 +1427,88 @@ describe('oauth2', () => {
...
+        client_authentication: ClientAuthentication.ClientSecretBasic,
+      };
+      const oauth2client = new OAuth2Client(opts);
+      const res = await oauth2client.getToken({
+        code: 'code here',
+        client_id: CLIENT_ID,
+      });
+      scope.done();
+      assert(res.res);
+      assert(res.res.data.access_token === 'abc');
```
danielbankhead, line 1461: Improves the error message in the event of a failed assertion (prints expected/actual rather than `false`).

```suggestion
      assert.equal(res.res.data.access_token, 'abc');
```

test/test.oauth2.ts
```diff
@@ -1421,6 +1427,88 @@ describe('oauth2', () => {
...
+        await oauth2client.getToken({
+          code: 'code here',
+          client_id: CLIENT_ID,
+        });
+      } catch (err) {
+        scope.done();
+        const e = err as GaxiosError;
+        assert(e);
+        assert.strictEqual(e.response!.status, 401);
+      }
```
danielbankhead, line 1493:
```suggestion
      try {
        await oauth2client.getToken({
          code: 'code here',
          client_id: CLIENT_ID,
        });
        throw new Error('Expected an error');
      } catch (err) {
        assert(err instanceof GaxiosError);
        assert.equal(err.response?.status, 401);
      } finally {
        scope.done();
      }
```

test/test.oauth2.ts
```diff
@@ -1421,6 +1427,88 @@ describe('oauth2', () => {
...
+        redirectUri: REDIRECT_URI,
+        endpoints: {
+          oauth2TokenUrl: 'mytokenurl',
+        },
+      };
+      const oauth2client = new OAuth2Client(opts);
+      assert(
+        oauth2client.client_authentication ===
+          ClientAuthentication.ClientSecretPost
+      );
```
danielbankhead, line 1509:
```suggestion
      assert.equal(oauth2client.client_authentication, ClientAuthentication.ClientSecretPost);
```

danielbankhead (changes requested):

src/auth/oauth2client.ts
BigTailWolf: >  The compiler gives me error of `Property 'client_secret' does not exist on type 'Omit<GetTokenOptions, "codeVerifier">'.ts(2339)`

danielbankhead: We can add the type to `values`:

```
values: Omit<GetTokenOptions, 'codeVerifier'> & {client_secret: string}
```

Essentially, we don't want properties that are difficult to track down in the event of an error or general debugging.

danielbankhead (approved):

lsirac (approved):

PR #1835, opened 2024-07-11
feat: Extend API Key Support

danielbankhead:
🦕


src/auth/googleauth.ts
```diff
@@ -451,6 +451,13 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
       );
     }
 
+    // No ADC, but perhaps there's an API key available...
+    if ('apiKey' in options && options.apiKey) {
```
sofisl, line 455: I actually think API Key should override ADC:

```Additionally, if application default credentials (ADC) are enabled and an API key is specified, ADC should be ignored.```

Also, we should add an error situation if an API Key and credentials are used together:

```API Keys and Credentials are mutually exclusive authentication methods and cannot be used together.```

**Review by sofisl — CHANGES_REQUESTED** [review]

**Comment by sofisl** [comment]
Thank you for taking on this work!!

File: .readme-partials.yaml [hunk 2]
```diff
@@ -220,6 +220,26 @@ body: |-
...
+
+  Additionally, an API key it can be used in `GoogleAuth` via the `clientOptions` parameter and will be passed to any generated `OAuth2Client` instances:
+  ```js
+  const auth = new GoogleAuth({
+    clientOptions: {
+      apiKey: 'my-api-key'
+    }
+  })
+  ```
+
```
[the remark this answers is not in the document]
danielbankhead: Added

samples/authenticateAPIKey.js
```diff
@@ -0,0 +1,63 @@
+// Copyright 2024 Google LLC
```
danielbankhead, line 1: Designed to match:
- https://cloud.google.com/docs/authentication/api-keys#using-with-client-libs

README.md
```diff
@@ -264,6 +264,28 @@ console.log(tokenInfo.scopes);
...
+#### Using an API Key
+
+An API key can be provided to the constructor:
+```js
+const client = new OAuth2Client({
+  apiKey: 'my-api-key'
+});
+```
+
+Note, classes that extend from this can utilize this parameter as well, such as `JWT` and `UserRefreshClient`.
```
sofisl, line 276:
```suggestion
Note, classes that extend from `OAuth2Client` can utilize this parameter as well, such as `JWT` and `UserRefreshClient`.
```

sofisl (approved):

PR #1837, opened 2024-07-25
fix: add missing return value to `fromAPIKey`

hansemannn:
According to [the docs](https://cloud.google.com/nodejs/docs/reference/google-auth-library/latest/google-auth-library/googleauth#google_auth_library_GoogleAuth_fromAPIKey_member_1_), the `fromAPIKey` function should return an instance of the class again, but returns `void` in the actual code.

To properly use it in chaining scenarios, e.g. `const client = new JWT().fromAPIKey('<api-key>')`, this issue should be addressed.

### DIFF
--- a/src/auth/jwtclient.ts
+++ b/src/auth/jwtclient.ts
@@ -377,12 +377,15 @@ export class JWT extends OAuth2Client implements IdTokenProvider {
   /**
    * Creates a JWT credentials instance using an API Key for authentication.
    * @param apiKey The API Key in string form.
+   * @return The current instance.
    */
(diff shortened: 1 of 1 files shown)


danielbankhead:
The docs are referencing `GoogleAuth`, which generates a `JWT`:

https://github.com/googleapis/google-auth-library-nodejs/blob/0b78d91534d97b37859a2303b332f8ccd52dbf69/src/auth/googleauth.ts#L806-L817

Sample usage:

```ts
import {GoogleAuth} from 'google-auth-library';

const key = '';

const client = auth.fromAPIKey(key);
```

PR #1843, opened 2024-08-06
feat: Add `AnyAuthClient` type

danielbankhead:
🦕

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -92,9 +92,7 @@ export {DefaultTransporter} from './transporters';
 type ALL_EXPORTS = (typeof import('./'))[keyof typeof import('./')];
 
 /**
- * A union type for all AuthClients.
- *
- * @experimental
+ * A union type for all {@link AuthClient `AuthClient`}s.
  */
 export type AnyAuthClient = InstanceType<
   // Extract All `AuthClient`s from exports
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -88,5 +89,14 @@ export {
 export {PassThroughClient} from './auth/passthrough';
 export {DefaultTransporter} from './transporters';
 
+/**
+ * A union type for all AuthClients.
+ *
+ * @experimental
+ */
+export type AuthClients = InstanceType<
```
danielbankhead, line 97: Not too sure about the name; `AuthClients` could imply `AuthClient[]`. Would something like `AnyAuthClient` be clearer?

leahecole: I see what you mean about the naming. `AnyAuthClient` works for me, or `UnionAuthClients`? this is a tough naming situation.

My other feedback is that there is a lot going on in the `InstanceType` and the `Extract` - is it possible to split this up into a couple of lines and to comment what's going on so it's easier to read? It's hard to parse just by looking at it

leahecole (changes requested):

src/index.ts
danielbankhead: Thanks, updated

leahecole (approved):

PR #1846, opened 2024-08-15
chore: pin compodoc

sofisl:


danielbankhead:
Currently pinned for Node 14+ support

PR #1861, opened 2024-09-10
docs: Update README to Point to Newer Instructions

piaxc:
Remove mention of service account key download and point to correct place for ADC setup.


danielbankhead (approved):

danielbankhead (changes requested):
Actually, could you update `.readme-partials.yaml` instead?

PR #1875, opened 2024-10-01
feat: Impersonated Universe Domain Support

danielbankhead:
🦕


src/auth/impersonated.ts
```diff
@@ -224,7 +241,9 @@ export class Impersonated extends OAuth2Client implements IdTokenProvider {
       delegates: this.delegates,
       audience: targetAudience,
       includeEmail: options?.includeEmail ?? true,
+      useEmailAzp: options?.includeEmail ?? true,
```
aeitzman, line 244: Is this field related to the universe domain change?

src/auth/googleauth.ts
```diff
@@ -659,18 +658,18 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
 
     const targetScopes = this.getAnyScopes() ?? [];
 
-    const client = new Impersonated({
+    return new Impersonated({
       ...json,
-      delegates: json.delegates ?? [],
```
aeitzman, line 664: Why was the delegates field deleted here?

src/auth/googleauth.ts
```diff
@@ -632,9 +632,7 @@ export class GoogleAuth<T extends AuthClient = JSONClient> {
       );
     }
 
-    // Create source client for impersonation
-    const sourceClient = new UserRefreshClient();
-    sourceClient.fromJSON(json.source_credentials);
+    const sourceClient = this.fromJSON(json.source_credentials);
```
aeitzman, line 635: This change is intentional to allow for impersonating any arbitrary type of source clients instead of just UserRefreshClient() right? Just for my own understanding.

src/auth/impersonated.ts
danielbankhead: It's recommended for impersonation, and now required for TPC. Other langs have this implemented:
- https://github.com/googleapis/google-auth-library-python/pull/1270
- https://github.com/googleapis/google-auth-library-php/pull/581/files#diff-9caed8dbd080d92079e2315774fe52b2e394a8f34d9e2883031ae9530bb1bbb5R138

src/auth/googleauth.ts
danielbankhead: Yep! Other clients are able to use impersonation and its required in TPC (e.g. for external accounts)

src/auth/googleauth.ts
danielbankhead: It's handled in the constructor itself:
- https://github.com/googleapis/google-auth-library-nodejs/blob/dc043cea4f87eeb642c165b9781586996f09854d/src/auth/impersonated.ts#L123-L127

We can de-dupe here

src/auth/googleauth.ts
aeitzman: Gotcha, only note I have is that the impersonation is handled directly by the external account client itself in this library (this already works for TPC, the inmpersonation URL gets generated with the universe domain when the create-credential-config is run in gCloud): https://github.com/googleapis/google-auth-library-nodejs/blob/a65d8a11450fdc0f69ea228def462e5a77beecd5/src/auth/baseexternalclient.ts#L602C4-L605C9 

This PR adds support for a different type of credential configuration file which we should support, but just wanted to point out I don't think this is a path we would expect people to normally use for BYOID creds.

aeitzman (approved):

src/auth/googleauth.ts
danielbankhead: Thanks! Important call-out.

PR #1878, opened 2024-10-03
fix: Disable Universe Domain Check

danielbankhead:
Temporarily Disable Universe Domain Check for MDS by default

🦕

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "ncp": "^2.0.0",
     "nock": "^13.0.0",
     "null-loader": "^4.0.0",
+    "pdfmake": "0.2.12",
     "puppeteer": "^21.0.0",
     "sinon": "^18.0.0",
     "ts-loader": "^8.0.0",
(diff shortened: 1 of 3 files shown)


sofisl (approved):

bd-nweinstein:
Hi there, I'm curious why this was disabled? Is there an associated issue? I think I'm encountering a bug caused by this (I'm on google-auth-library 9.4.1) where my app running in GKE randomly fails to connect to bigquery or firestore with the error

`Error: Invalid response from metadata service: incorrect Metadata-Flavor header.`

It seems as though my metadata server does not return this header when `universe_domain` is queried.

ernestoalejo:
@bd-nweinstein I think we have the same issue: https://github.com/googleapis/google-auth-library-nodejs/issues/1946

mssodhi:
@bd-nweinstein @ernestoalejo we're also experiencing the same issue, but using pubsub. Randomly started getting `Error: Invalid response from metadata service: incorrect Metadata-Flavor header.`

PR #1898, opened 2024-11-22
feat: add debug logging support

feywind:
Adds debug logging support for auth. This needs a bit more work/testing still.

It's also waiting on this to finish:
https://github.com/googleapis/gax-nodejs/pull/1669

### DIFF
--- a/src/auth/baseexternalclient.ts
+++ b/src/auth/baseexternalclient.ts
@@ -20,8 +20,6 @@ import {
   GaxiosResponse,
 } from 'gaxios';
 import * as stream from 'stream';
-import {log as makeLog} from 'google-logging-utils';
-const log = makeLog('auth');
 
 import {Credentials} from './credentials';
 import {AuthClient, AuthClientOptions} from './authclient';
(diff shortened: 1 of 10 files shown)


feywind:
I need to update this for when the debug logger piece is officially reviewed and released, and there's a system test thing to look at, but the code itself is largely ready, I think.

src/auth/baseexternalclient.ts
```diff
@@ -20,6 +20,8 @@ import {
   GaxiosResponse,
 } from 'gaxios';
 import * as stream from 'stream';
+import {log as makeLog} from 'google-logging-utils';
+const log = makeLog('auth');
```
d-goog, line 24: I think it may be cleaner and flexible to add this to the base `AuthClient` class and have it as a member or static method.

This will allow:
- Customers to change/modify the log functionality (if they want)
- Dedupe the import statements across the files

Additionally, in the next major we're planning to unify the `Transporter` instance into `AuthClient` via this PR:
- https://github.com/googleapis/google-auth-library-nodejs/pull/1860/files#diff-1be86cf1a764f43a299d2d037e8db93a74904846592cfe5f31c0981079c08992R256

It would be an even smaller migration after that has landed as we could extend the `GaxiosOptions` interface with a `Symbol` (which would note the type of log) and use `log.info` on every request (as all requests would use that `_request` method).

feywind: Thanks for the quick review. I'll make that change tomorrow.

feywind: A lot of the logging can be done through AuthClient now, good call. There are a few objects that don't seem to derive ultimately from `AuthClient`, so I moved makeLog() into the class itself to make it cleaner. (`makeLog` caches the loggers by name, so there's little harm in calling it multiple times.)

d-goog (approved):

d-goog:
Is this version of the PR safe to close?

feywind:
Yep, sorry, I seem to have forgotten to close it.

PR #1902, opened 2024-12-19
docs(readme): add example for service_account_impersonation_url in clientOptions

varun27896:
Thank you for implementing support for passing custom AWS credentials in your library. Your efforts are much appreciated.

While I was working on implementing the custom supplier, I noticed that the `service_account_impersonation_url` field is missing from the `clientOptions` object in your documentation. Without including this field, a permissions error is thrown during execution. This PR adds that field to the README


d-goog (approved):
Thanks!

PR #1903, opened 2024-12-19
feat: add debug logging support

feywind:
Adds debug logging support for auth.

It's also waiting on this to finish:
https://github.com/googleapis/gax-nodejs/pull/1669

This is a duplicate of [another PR](https://github.com/googleapis/google-auth-library-nodejs/pull/1898) to make the GitHub bot macguffins go.


d-goog:
Hello! With recent enhancements in the library we may be able to greatly streamline this PR by adding a listener here (thanks to the removal of Transporter):
- https://github.com/googleapis/google-auth-library-nodejs/blob/a511e562783402468093d70e77e73762fa95ae79/src/auth/authclient.ts#L325-L345

Every auth request goes through this function. Let me know if you’d like to sync on this or have any questions.

feywind:
@d-goog Ah, lovely. Thanks!

d-goog (approved):

src/auth/refreshclient.ts
```diff
@@ -107,8 +109,10 @@ export class UserRefreshClient extends OAuth2Client {
         refresh_token: this._refreshToken,
         target_audience: targetAudience,
       } as {}),
-    });
+    };
+    AuthClient.setMethodName(opts, 'fetchIdToken');
```
d-goog, line 113: Completely optional: Since these classes extend `AuthClient` we could use the class's `setMethodName` and avoid importing `AuthClient`.

```suggestion
    UserRefreshClient.setMethodName(opts, 'fetchIdToken');
```

Here and in other places

feywind: Actually, after thinking about it, I'm not a super fan of static inheritance like that. I think it makes it more obscure what's actually being called. (Also, several of the other usages were not deriving from anything that had `setMethodName` so the usage was mixed.) I'm okay with it either way, but maybe leaning toward not changing it.

PR #1916, opened 2025-01-22
docs: `README.md` grammar

dandv:
See https://github.com/googleapis/google-auth-library-nodejs/pull/1911#discussion_r1924775940

Fixes #1911 🦕

PR #1917, opened 2025-01-22
Advise which file to edit instead of the README.md

dandv:
See https://github.com/googleapis/google-auth-library-nodejs/pull/1916


d-goog:
@dandv as mentioned by the [generated-files-bot](https://github.com/apps/generated-files-bot), the README file itself is a generated from a template. Adding the comment directly to the README would not help in this case. See:
- https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/node_library/README.md

PR #1923, opened 2025-01-23
docs: JSON credential best practices

d-goog:
🦕

### DIFF
--- a/src/auth/jwtclient.ts
+++ b/src/auth/jwtclient.ts
@@ -303,6 +303,10 @@ export class JWT extends OAuth2Client implements IdTokenProvider {
   /**
    * Create a JWT credentials instance using the given input options.
    * @param json The input object.
+   *
+   * @remarks
+   *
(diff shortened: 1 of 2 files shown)


src/auth/jwtclient.ts
```diff
@@ -303,6 +303,10 @@ export class JWT extends OAuth2Client implements IdTokenProvider {
   /**
    * Create a JWT credentials instance using the given input options.
    * @param json The input object.
+   *
+   * @remarks
+   *
+   * **Important**: If you accept a credential configuration (credential JSON/File/Stream) from an external source for authentication to Google Cloud, you must validate it before providing it to any Google API or library. Providing an unvalidated credential configuration to Google APIs can compromise the security of your systems and data. For more information, refer to {@link https://cloud.google.com/docs/authentication/external/externally-sourced-credentials Validate credential configurations from external sources}.
```
sai-sunder-s, line 309: this method is going to allow only Service accounts right?

If so we do not need to warn here. Having it doesn't have any downside though

sai-sunder-s (approved):

sai-sunder-s:
Is there going to be a separate change for the client libs or is this it?

sai-sunder-s:
> Is there going to be a separate change for the client libs or is this it?

I think I got it. GoogleAuthOptions is used in the client libs?

d-goog:
> Is there going to be a separate change for the client libs or is this it?

This is it; it will automatically propagate down.



> > Is there going to be a separate change for the client libs or is this it?
> 
> I think I got it. GoogleAuthOptions is used in the client libs?

Correct.

Issue #1489, opened 2022-11-16
How to set user agent in the auth request ?


amrsa1:
our domain is protected with IAP authentication, and its blocking our e2e test.

so we have added API authentication using the code snippet below 

```
import { GoogleAuth } from 'google-auth-library';
const auth = new GoogleAuth();
const url = `${configList.baseURL}/dashboard`;
const targetAudience = "51297171365dbt6s3oc4d09cf.apps.googleusercontent.com";
async function IapAuth() {
    try {
            console.info(`request IAP ${url} with target audience ${targetAudience}`);
            const client = await auth.getIdTokenClient(targetAudience);
            const res = await client.request({
                url, headers: {
                    user-Agent: 'e2e.test'
                }
            });
            IAP_token = res.config.headers.Authorization
            process.env.IAP_token = IAP_token
            console.log(`√ Iap token has been obtained ----> ${JSON.stringify(IAP_token)}`)
    } catch (err: any) {
        console.error(JSON.stringify(`X ---> ${JSON.stringify(err.message)}`));
        process.exitCode = 1;
    }
}

But since our domain is protected by DataDome, we need to pass a certain user agent in the request in order to not be blocked by Recaptcha, but the solution above didn't work out  since my user agent is not overnight the existing one but it appends with it 

and I'm getting 
```

the response below

```
response: {
    config: {
      url: 'https://test.net/ch/test/en/accounts/mpre/dashboard',
      headers: [Object],
      paramsSerializer: [Function: paramsSerializer],
      validateStatus: [Function: validateStatus],
      responseType: 'json',
      method: 'GET'
    },
    data: {
[...]

Issue #1418, opened 2022-05-28
No pem found for envelope (wrong 'kid' property match)


jpike88:
https://github.com/googleapis/google-auth-library-nodejs/blob/d7893c1dc70b3aa45c12bb9b6c0e5346a293b130/src/auth/oauth2client.ts#L1304

My code looks like this:

```
const client = new OAuth2Client(
		'CLIENT_ID_1'
	);
	
const ticket = await client.verifyIdToken({
			idToken: token,
			audience: [
				'CLIENT_ID_1',
				'CLIENT_ID_2',
			],
		});
		const payload = ticket.getPayload();
		return payload.email;
```

It works well for my iOS and web Google Sign in implementations.
For the android one, it's failing, with error:

```No pem found for envelope: {"alg":"RS256","kid":"6f8e1cb15641463c6df0f33394b03c92fcc889ac","typ":"JWT"}```

The payload and envelope are separated correctly, it seems there is just the wrong 'kid' matching going on with the certs fetched from `getFederatedSignonCertsAsync`. How can I fix this?

And this is unlikely related to caching, as I had the same identical problem on local emulator, as well as a physical device, also have the same problem testing in the cloud. All freshly signed into a Google account without ever being signed into it.

jpike88:
Found the root of the problem

The kid is matching a certificate on this url.
`https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com`

However, this library doesn't include those certificates when figuring out which certificate to work with.

@danielbankhead can you explain this? There's this bizarre crossover with google auth and firebase... the library is called googleSignIn in android, this is incredibly confusing and has blown away a lot of my time

jpike88:
My workaround looks like this now... nasty but it works.

```ts

import { OAuth2Client } from 'google-auth-library';
import firebase from 'firebase-admin';

const firebaseClient = firebase.initializeApp({
				credential: {
// firebase credential key
});

// resolve a token to the user email's
export async function deriveEmailFromGoogleToken(
	token: string
): Promise<string> {
	const client = new OAuth2Client(
		'CLIENT_ID_1'
	);
	
	try {
		const ticket = await client.verifyIdToken({
			idToken: token,
			audience: [
				'281074435194-iacdh6vqefvlkg5d39612ovu8qerhj9i.apps.googleusercontent.com',
[...]

PR #1937, opened 2025-02-05
refactor!: Remove `Transporter`

d-goog:
## Description

The `Transporter` interface has been a pain-point for customers for years. To highlight a few:
- https://github.com/googleapis/google-auth-library-nodejs/pull/1406
- https://github.com/googleapis/google-auth-library-nodejs/issues/1405
- https://github.com/googleapis/google-auth-library-nodejs/issues/1489
- https://github.com/googleapis/google-auth-library-nodejs/issues/1621

Additionally, this was the primary blocker for Firebase-Admin's adoption for this library:
- https://github.com/googleapis/google-auth-library-nodejs/issues/1418#issuecomment-1908458291

As `Transporter` is essentially a wrapper around `Gaxios` with a few auth-centric methods we can remove this layer of complexity to greatly improve UX.

## Testing

As we're removing `Transporter` unit test coverage has improved with this change.
Additionally, the current integration tests in pass with minimal change.

## Impact

With this change we're optimizing the configuration and request flow to reduce complexity:
- From: Client Libraries -> Auth -> Transporter -> Gaxios
- To: Client Libraries -> Auth -> Gaxios

Additionally, this will unblock:
- Auth Metrics (which would benefit from centralized request handling)
  - https://github.com/googleapis/google-auth-library-nodejs/pull/1860
- Debug logging (which would benefit from centralized request handling)
  - https://github.com/googleapis/google-auth-library-nodejs/pull/1903
- `gaxios` v6 (as it would streamline the PR and make it significantly smaller given we do not have)
- `fetch` API Support in Auth (via `gaxios` v6)

## Additional Information

In a separate PR I will draft a migration guide for customers migrating from v9 to v10.

As a TL;DR for migrating:
- if using `Transporter` use `Gaxios` instead, which is conveniently re-exported by this library
- if passing `transporter` config no changes are necessary

🦕


test/test.authclient.ts
```diff
@@ -38,4 +39,31 @@ describe('AuthClient', () => {
       assert.equal(authClient[camelCased], value);
     }
   });
+
+  it('should allow disabling of the default interceptor', () => {
```
sofisl, line 43: Ultimately, I'm a bit confused about this change: are we replacing functionality or removing it entirely? If replacing, it would be nice to see more test coverage (something at parity with confirming transporters functionality). If the latter, then I think maybe it would be best to add some type of warning for users attempting to use the Transporter property.

sofisl (commented):
In general, a few notes:

1. The [issues](https://github.com/googleapis/google-auth-library-nodejs/issues?q=is%3Aissue%20transporter%20) linked here don't make it immediately clear to me that we need to remove `transporter`. Can you help explain how you came to this solution?
2. It seems like this is just removing `transporter`. Since that is the case, can we add a warning to the `transporter` property, and confirm it with a test? If it really is refactoring, can we have test parity with the previous `transporters` test file?
3. Can you provide instructions for how downstream libraries will need to handle this change?
4. Can you link any relevant PRs that explain how gaxios can provide this functionality? It seems like we're now just relying on gaxios for `transporter` and I couldn't immediately tell which commits where relevant [here](https://github.com/googleapis/gaxios/commits/main/)

sofisl:
Ok, sgtm. Thanks for the explanation. Would it be possible to [keep this test file](https://github.com/googleapis/google-auth-library-nodejs/pull/1937/files#diff-78c30591fddee985625f2aa9bf36440e9d687155e8dcbb22ec5ab6c019bc961b), and replace with the Gaxios `transporter` functionality?

d-goog:
> Ok, sgtm. Thanks for the explanation. Would it be possible to [keep this test file](https://github.com/googleapis/google-auth-library-nodejs/pull/1937/files#diff-78c30591fddee985625f2aa9bf36440e9d687155e8dcbb22ec5ab6c019bc961b), and replace with the Gaxios `transporter` functionality?

Yep, modernized and migrated the tests over into AuthClient's test file

sofisl (approved):

PR #1938, opened 2025-02-12
feat!: `Request` Revamp

### LINKED ISSUE #1829: DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead
when  import { OAuth2Client } from "google-auth-library"; this warning appears and when removed the import, it disappears 

```
(node:32561) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

OS - MacOs sonoama v.14.5
node -v = 20.14.0
npm -v = 10.8.1
"google-auth-library": "^9.11.0",

d-goog:
## Description

This is a large refactor of the auth library's internals with sweeping improvements in UX, correctness, performance, security, and scalability.

Examples:
- UX: Adopts [`fetch` API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) conventions
- Correctness: Allows users to add charset to content-types and removes `responseType` parameter from `GaxiosOptions` (which was problematic in error cases)
- Performance: There is far less object-copying when things can be safely reused, especially for `Headers`

## Testing

All existing tests pass and enhanced numerous throughout the codebase. Notably, the sample tests required minimal changes.

## Impact

This revamp brings number of improvements to library:
- It returns users predictable, consistent [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers), while allowing the open, flexible [`HeadersInit`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#setting_headers) for requests.
- Bumps our dependencies to the Node 18+ versions
- Simplifies requests:
  - With native support for the [`RequestInit`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#setting_a_body) body types, users can omit certain headers, such as `content-type`
- Improves cross-platform compatibility through JS standards adoption (`Request`, `Headers`, `URLSearchParams`, etc.)/

## Additional Information

While a large PR, most of the changes are internal and doesn't affect external API consumption much outside from returning `Headers`.

Note: a large portion of this PR are non-functional changes related to Node 18's `Headers` - which are not case-insensitive when compared in tests (which required a lot of lower-casing header names).

A follow-up PR will add `fetch` support.

Fixes: #1829

🦕


src/auth/awsrequestsigner.ts
```diff
@@ -60,15 +49,15 @@ interface GenerateAuthHeaderMapOptions {
   // The AWS service URL query string.
   canonicalQuerystring: string;
   // The HTTP method used to call this API.
-  method: HttpMethod;
+  method: string;
```
sofisl, line 52: why is this type changing? seems more flexible? also not clear if this change has to do with the gaxios upgrade.

src/auth/awsrequestsigner.ts
```diff
@@ -127,29 +116,36 @@ export class AwsRequestSigner {
     const additionalAmzHeaders = amzOptions.headers;
     const awsSecurityCredentials = await this.getCredentials();
     const uri = new URL(url);
+
+    if (typeof requestPayload !== 'string' && requestPayload !== undefined) {
+      throw new TypeError(
+        `'requestPayload' is expected to be a string if provided. Got: ${requestPayload}`
+      );
```
sofisl, line 123: do we know what requestPayload type could be? This could be an ugly error `Got [object Object]`

src/auth/oauth2client.ts
```diff
@@ -1112,20 +1109,17 @@ export class OAuth2Client extends AuthClient {
...
-      opts.headers = opts.headers || {};
-      if (r.headers && r.headers['x-goog-user-project']) {
-        opts.headers['x-goog-user-project'] = r.headers['x-goog-user-project'];
-      }
-      if (r.headers && r.headers.Authorization) {
-        opts.headers.Authorization = r.headers.Authorization;
-      }
+      opts.headers = Gaxios.mergeHeaders(opts.headers);
+
+      this.addUserProjectAndAuthHeaders(opts.headers, r.headers);
```
sofisl, line 1116: nice DRY!

src/auth/oauth2common.ts
```diff
@@ -160,45 +160,44 @@ export abstract class OAuthClientAuthHandler {
...
-        } else {
-          throw new Error(
-            `${contentType} content-types are not supported with ` +
-              `${this.#clientAuthentication!.confidentialClientType} ` +
-              'client authentication'
-          );
-        }
-      } else {
+
+      if (!METHODS_SUPPORTING_REQUEST_BODY.includes(method)) {
```
sofisl, line 164: I see why you made this change - definitely more readable! - but ideally could be another PR (given this is a big PR already, and this is more of a refactor, and we're already doing so many other gaxios changes). NB.

src/auth/stscredentials.ts
```diff
@@ -203,30 +202,21 @@ export class StsCredentials extends OAuthClientAuthHandler {
...
       options: options && JSON.stringify(options),
     };
-    // Remove undefined fields.
-    Object.keys(values).forEach(key => {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      if (typeof (values as {[index: string]: any})[key] === 'undefined') {
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        delete (values as {[index: string]: any})[key];
+
+    // Keep defined fields.
```
sofisl, line 206: Again, another small refactor that in general seems harmless but in the context of a feature, might be better as a F/U.

src/auth/urlsubjecttokensupplier.ts
```diff
@@ -87,7 +87,6 @@ export class UrlSubjectTokenSupplier implements SubjectTokenSupplier {
       url: this.url,
       method: 'GET',
       headers: this.headers,
-      responseType: this.formatType,
```
sofisl, line 90: intended?

src/auth/awsrequestsigner.ts
d-goog: We want it to be a string, however we marshal objects to strings a few lines above it. This is preserving the existing functionality and makes it clearer, we used to throw later in the call stack.

src/auth/stscredentials.ts
d-goog: Understood, however this required the refactor now given the more precise type requirements of 
`URLSearchParams`.

sofisl (approved):

PR #1939, opened 2025-02-20
feat: `fetch`-Compatible API

d-goog:
## Description

This will greatly improve ergonomics for customers and our downstream libraries. Notably:
- `GoogleAuth`/`AuthClient` can be a drop-in replacement for libraries requiring `fetch`-API compliance

Here's an example for customers wanting this functionality:
- https://www.youtube.com/watch?v=IY-z00bfnOc&t=291s (from  4:51-5:10)

## Impact

Customers can now streamline their request management to:
```ts
const fetchWithAuth: typeof fetch = (...args) => auth.fetch(...args);
await fetchWithAuth('https://example.com');
```

## Testing

Added tests and updated some samples to use `#fetch`.

## Additional Information

To avoid parsing/marshaling the returned data from `GoogleAuth#fetch`/`AuthClient#fetch` to `GaxiosResponse#data`, set `{responseType: 'stream'}` for the [`GaxiosOptions`](https://github.com/googleapis/gaxios).
This can be useful when using this API as a drop-in replacement for `fetch` and the receiving API expects [`Response#bodyUsed`](https://developer.mozilla.org/en-US/docs/Web/API/Response/bodyUsed) to be `false`.
We could expose a `noMarshalFetchData` option here for `AuthClientOptions` or upstream in `GaxiosOptions` if this becomes a widely-desired feature.

Builds on:
- https://github.com/googleapis/gaxios/pull/680

🦕


sofisl (commented):
Exciting! Just so we don't confuse people, can we keep the `request` documentation in the README as well (you can use fetch, or request)? Also, can we create a sample for request as well?

d-goog:
> Exciting! Just so we don't confuse people, can we keep the `request` documentation in the README as well (you can use fetch, or request)? Also, can we create a sample for request as well?

I'll update the README to mention both are available and will be supported long-term - I think we'll want to encourage folks use the `fetch` method going forward rather than `request` given flexibility of the API

sofisl:
We are still testing request so no need to keep the other samples.

sofisl (approved):

PR #1940, opened 2025-02-21
feat: Normalize `GoogleAuth<T>` from `JSONClient` to `AuthClient`

### LINKED ISSUE #1677: feat!: Normalize `GoogleAuth<T>` to `AuthClient`
We should normalize `GoogleAuth<T extends AuthClient = JSONClient>` to `GoogleAuth<T extends AuthClient = AuthClient>` (and `GoogleAuthOptions` as well).

Additionally, we should:

- change `clientOptions` from:
https://github.com/googleapis/google-auth-library-nodejs/blob/4f94ffe474ee41b560813b81e8d621be848d38d0/src/auth/googleauth.ts#L106-L114
to:
  ```ts
  /**
   * Options object passed to the constructor of the client
   */
  clientOptions?: ConstructorParameters<typeof T>[0];
  ```
  In both `GoogleAuth` and `GoogleAuthOptions` to accurately match the expected client options.

- change `cachedCredential` from:
https://github.com/googleapis/google-auth-library-nodejs/blob/f950465967e6140d061d44ab0d12f08e0117c0ec/src/auth/googleauth.ts#L160-L161
to:
  ```ts
  cachedCredential: T | null = null; 
  ```
  in order to normalize the expected `AuthClient`

Dependencies:
- https://github.com/googleapis/google-auth-library-nodejs/pull/1624
- https://github.com/googleapis/google-auth-library-nodejs/pull/1663
- https://github.com/googleapis/google-auth-library-nodejs/issues/1672

Implementation Notes:
- Look into overload narrowing (which is a pain at the time of writing), which is required for accurate `ConstructorParameters` results:
  - https://github.com/microsoft/TypeScript/issues/37079
  - https://stackoverflow.com/questions/59745527/is-it-possible-to-narrow-the-types-of-overloaded-parameters-without-exhaustively
  - Note: We may just have to reorganize overloads in each `AuthClient` to the desired overload signature as the first overload ('desired' = whichever one exposes the extended `AuthClientOptions` as the first option).

d-goog:
## Description

Improve a long-open pain-point for customers by setting the default AuthClient generic in GoogleAuth to `AuthClient`:
- https://github.com/googleapis/google-auth-library-nodejs/issues/1677

This improves the default compatibly between `GoogleAuth` instances as they would all be compatible by default (given `AuthClient` is the base type of all `AuthClients`)

## Impact

A few things:
- Improved ergonomics for sharing instances of `GoogleAuth`
- Provides a type hints for `clientOptions`, making it easier for customers to see options for all auth clients:
  - Screenshot 2025-02-20 at 9 54 08 PM

## Testing

This is mostly a TypeScript-level change; non-functional. However, I've added a simple test to ensure compatibility.

## Additional Information

The `clientOptions` field could be improved by the following:
- https://github.com/microsoft/TypeScript/issues/3841
As it would be nice to if we could use the constructor of `T` to narrow the options to just `T`'s constructor options.

Fixes #1677

🦕


sofisl (commented):
Since it's breaking, should this wait until Node 20? We can close this PR and create a next major bug for it.

sofisl:
If you can confirm it will be a no-op for already released packages (gapics, handwrittens, and apiary), then I'm ok with it. Otherwise since we've already released these as majors I don't think it'll be wise to add another breaking change.

d-goog:
Given that the downstream projects were released out-of-band of our agreed upon plan I do not think that should gate this particular enhancement: go/cloudsdk-js-2024

sofisl (approved):

PR #1966, opened 2025-05-14
feat!: check for trust boundaries before sending requests google apis.

vverman:
This change includes trust boundaries for Service Accounts in the google auth nodejs library. 

## Impact

This will allow clients who use the google auth library to also enable trust boundaries ensuring service accounts can access only those resources which are within the trust boundary. We use this [design doc](https://docs.google.com/document/d/1IkRoxRelBF-wUHktFLxKqfcQ2tqAwUg-XcYz_IrVHsQ/edit?resourcekey=0-KdUdpwuLu6eJtcdDb0lISw&tab=t.0#heading=h.17wg41voij6q) as reference for the code changes.

## Testing
Testing will be added in post initial design review.

### DIFF
--- a/src/auth/authclient.ts
+++ b/src/auth/authclient.ts
@@ -198,6 +198,9 @@ export declare interface AuthClient {
   on(event: 'tokens', listener: (tokens: Credentials) => void): this;
 }
 
+/**
+ * The base of all Auth Clients.
+ */
 export abstract class AuthClient
   extends EventEmitter
   implements CredentialsClient
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         node: [18, 20, 22]
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node }}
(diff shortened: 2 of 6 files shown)


lsirac:
The current approach has each client duplicating trust boundary logic. Can we centralize and simplify this?

src/auth/trustboundary.ts
```diff
@@ -0,0 +1,266 @@
...
+ *
+ * @param authenticatedClient An authenticated AuthClient instance.
+ * @param url The specific URL to fetch data from.
+ * @returns A Promise resolving to a new TrustBoundary instance.
+ * @throws {Error} If the request fails or the response is invalid.
+ * @internal
+ */
+async function _fetchTrustBoundaryData(
+  authenticatedClient: AuthClient,
+  authHeader: string,
```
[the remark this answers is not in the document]
vverman: Since the lookup endpoint is called in the refreshTokenNoCache(), and in the refreshToken function, the new authClient credentials aren't set. 

They are set in the getRequestMetadataAsync() in the oauth2client. We cannot make the trust boundary calls in the oauth2client as it doesn't exclusively pertain to service accounts.

Another option is to set the access_token of the client credentials in the refreshTokenNoCache function. That would allow us to get the header directly from the authClient without the header param.

src/auth/authclient.ts
```diff
@@ -317,11 +325,22 @@ export abstract class AuthClient
     ) {
       headers.set('x-goog-user-project', this.quotaProjectId);
     }
+    if (
+      !headers.has('x-goog-allowed-locations') &&
```
nbayati, line 329: I think we need to send the most up to date TB data. The `!headers.has('x-goog-allowed-locations')` check will skip setting and updating the value it if was previously set.

src/auth/authclient.ts
```diff
@@ -317,11 +325,22 @@ export abstract class AuthClient
...
+      this.trustBoundary &&
+      this.trustBoundary.encodedLocations
```
nbayati, line 331: Checking that `encodedLocations` is not an empty string is not enough.
We only want to set the header if the value is not no-op (0x0)

src/auth/authclient.ts
```diff
@@ -343,6 +363,10 @@ export abstract class AuthClient
       target.set('authorization', authorizationHeader);
     }
 
+    if (xGoogAllowedLocs) {
```
nbayati, line 366: The name of the method still implies that it only sets `x-goog-user-project` and `authorization` headers. I'd recommend either creating a separate method to set `x-goog-allowed-locations`, or updating the method name.

PR #2020, opened 2025-05-27
feat: Implement TrustBoundaries

vverman:
This change includes trust boundaries for Service Accounts, Workload and Workforce Federation in the google auth nodejs library.

Impact
This will allow clients who use the google auth library to also enable trust boundaries ensuring service accounts can access only those resources which are within the trust boundary. We use this [design doc](https://docs.google.com/document/d/1IkRoxRelBF-wUHktFLxKqfcQ2tqAwUg-XcYz_IrVHsQ/edit?resourcekey=0-KdUdpwuLu6eJtcdDb0lISw&tab=t.0#heading=h.17wg41voij6q) as reference for the code changes.

Testing
Unit testing included for all clients who use trust boundaries.

### DIFF
--- a/src/auth/authclient.ts
+++ b/src/auth/authclient.ts
@@ -339,7 +339,7 @@ export abstract class AuthClient
   }
 
   /**
-   * Adds the `x-goog-user-project` and `authorization` and 'x-goog-allowed-locations'
+   * Adds the `x-goog-user-project`, `authorization`, and 'x-goog-allowed-locations'
    * headers to the target Headers
    * object, if they exist on the source.
    *
@@ -347,10 +347,7 @@ export abstract class AuthClient
    * @param source the headers to source from
    * @returns the target headers
    */
-  protected addUserProjectAndAuthAndTBHeaders<T extends Headers>(
-    target: T,
-    source: Headers,
-  ): T {
+  protected addCommonHeaders<T extends Headers>(target: T, source: Headers): T {
     const xGoogUserProject = source.get('x-goog-user-project');
     const authorizationHeader = source.get('authorization');
     const xGoogAllowedLocs = source.get('x-goog-allowed-locations');
--- a/src/auth/authclient.ts
+++ b/src/auth/authclient.ts
@@ -20,7 +20,7 @@ import {OriginalAndCamel, originalOrCamelOptions} from '../util';
 import {log as makeLog} from 'google-logging-utils';
 
 import {PRODUCT_NAME, USER_AGENT} from '../shared.cjs';
-import { TrustBoundary, TrustBoundaryData } from './trustboundary';
+import { TrustBoundaryData } from './trustboundary';
 
 /**
  * Easy access to symbol-indexed strings on config objects.
@@ -218,6 +218,7 @@ export abstract class AuthClient
   eagerRefreshThresholdMillis = DEFAULT_EAGER_REFRESH_THRESHOLD_MILLIS;
   forceRefreshOnFailure = false;
   universeDomain = DEFAULT_UNIVERSE;
+  trustBoundaryEnabled: boolean
   trustBoundary?: TrustBoundaryData | null;
 
   /**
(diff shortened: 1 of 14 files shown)


d-goog (approved):
LGTM so far - just a few minor things

lsirac:
I think the current implementation is acceptable but can potentially be improved. @d-goog @nbayati would like to hear your thoughts:

1) Instead of each auth client subclass (like JWTClient, ComputeClient, BaseExternalAccountClient, etc.) implementing a full fetchTrustBoundary method as currently defined in the TrustBoundaryProvider interface, their primary role would be to provide the specific configuration data required for the trust boundary lookup. This mainly involves supplying the unique API endpoint URL relevant to that client type (e.g., the service account lookup URL for JWTClient or the dynamically determined workload/workforce pool URL for BaseExternalAccountClient). The core logic for actually performing the lookup (including caching, enablement checks, and the HTTP request via the common lookupTrustBoundary utility in trustboundary.ts) would be common and invoked centrally.

2)  The invocation of the trust boundary data fetch should be managed at a higher, common level, likely within the base AuthClient class, as part of its token refresh lifecycle. Currently, each client subclass individually calls this.fetchTrustBoundary(...) after its specific access token refresh mechanism completes (e.g., in jwtclient.ts, computeclient.ts). My suggestion is to move this invocation into a shared method within AuthClient. After any subclass successfully refreshes its access token, this centralized method in AuthClient would then be responsible for initiating the trust boundary data fetch, using the endpoint information provided by the subclass (as per point 1) and the newly acquired access token.

This approach would mean individual auth clients are primarily responsible for "declaring" that they are "trust boundary supported" and give their trust boundary endpoint. The AuthClient base class, in conjunction with the existing lookupTrustBoundary utility, handles the "how and when" of fetching and managing the trust boundary data in a more centralized manner.

src/auth/baseexternalclient.ts
```diff
@@ -715,4 +731,90 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+        if (wfPoolId) {
+          lookupTrustBoundaryUrl = WORKFORCE_LOOKUP_ENDPOINT.replace(
+            '{pool_id}',
+            encodeURIComponent(wfPoolId),
+          );
+        } else {
+          throw new Error(
+            'TrustBoundaryLookup: Failed to fetch trust boundary data due to missing workforce pool id',
+          );
+        }
```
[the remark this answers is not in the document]
vverman: Done!

src/auth/impersonated.ts
```diff
@@ -253,4 +268,27 @@ export class Impersonated extends OAuth2Client implements IdTokenProvider {
...
   }
+
+  /**
+   * Fetches a trustBoundary for the given service account.
+   * @param authHeader the authheader for calling the lookup endpoint
+   */
+  async fetchTrustBoundary(
+    authHeader: string,
+  ): Promise<TrustBoundaryData | null> {
+    if (!this.targetPrincipal) {
```
nbayati, line 279: We should throw an error if target principal is missing.

src/auth/jwtclient.ts
```diff
@@ -407,4 +423,27 @@ export class JWT extends OAuth2Client implements IdTokenProvider {
...
   }
+
+  /**
+   * Fetches a trustBoundary.
+   * @param authHeader the authheader for calling the lookup endpoint
+   */
+  async fetchTrustBoundary(
+    authHeader: string,
+  ): Promise<TrustBoundaryData | null> {
+    if (!this.email) {
```
nbayati, line 434: If this.email is missing, we should throw an error instead of returning a cached value. The lookup URL requires the email to function, and falling back to a potentially stale cache is a security risk.

src/auth/authclient.ts
```diff
@@ -314,23 +325,32 @@ export abstract class AuthClient
     ) {
       headers.set('x-goog-user-project', this.quotaProjectId);
     }
+    if (
+      this.trustBoundary &&
+      this.trustBoundary.encodedLocations &&
+      this.trustBoundary.encodedLocations !== NoOpEncodedLocations
```
nbayati, line 331: Based on my recent conversations with the backend folks: We should explicitly set the x-allowed-locations header to an empty string when the trust boundary is a "no-op" (0x0), rather than omitting the header entirely. They mentioned that while the server can handle both cases, passing an empty header is the preferred behavior as it provides more explicit information to the server about the SDK's action.

PR #2036, opened 2025-06-12
feat: Implement X509 Cert Based Authentication

vverman:
## 

With this PR, we are now able to use X509 Certificates to authenticate via Google Auth Library.

## Testing
Added unit tests and all integration tests passing as per [this](https://docs.google.com/document/d/1uutxffv21zZfZb1FVIaReROrWh_jiZS7dtB9NFZh_NI/edit?tab=t.0#heading=h.jjf3ex8pq9ke) testing document

## Additional Information

Change also includes trust chain verification to be passed to STS endpoint


src/auth/certificatesubjecttokensupplier.ts
```diff
@@ -0,0 +1,299 @@
...
+      );
+    }
+  }
+
+  /**
+   * Constructs the subject token, which is the base64-encoded certificate chain.
+   * @returns A promise that resolves with the subject token.
+   */
+  public async getSubjectToken(): Promise<string> {
+    const opts: GaxiosOptions = {method: 'GET'};
```
nbayati, line 149: Are these two lines needed?

src/auth/certificatesubjecttokensupplier.ts
```diff
@@ -0,0 +1,299 @@
...
+    AuthClient.setMethodName(opts, 'getSubjectToken');
```
nbayati, line 150: Are these two lines needed?

src/auth/certificatesubjecttokensupplier.ts
```diff
@@ -0,0 +1,299 @@
...
+
+import {GaxiosOptions} from 'gaxios';
+import {SubjectTokenSupplier} from './identitypoolclient';
+import {AuthClient} from './authclient';
+import {getWellKnownCertificateConfigFileLocation, isValidFile} from '../util';
+import * as fs from 'fs';
+import {X509Certificate} from 'crypto';
+import * as https from 'https';
+
+// --- Constants ---
```
nbayati, line 23: I noticed the use of sectioning comments like `// --- Main Class ---`. While these can be useful for organization, it seems this pattern isn't common in the rest of our codebase. I think it's important to maintain consistency in the codebase.

src/util.ts
```diff
@@ -251,3 +258,43 @@ export function removeUndefinedValuesInObject(object: {[key: string]: any}): {
   });
   return object;
 }
+
+/**
+ * Helper to check if a path points to a valid file.
+ */
+export function isValidFile(filePath: string): boolean {
```
nbayati, line 265: Consider adding unit tests for the new helper methods.

src/auth/certificatesubjecttokensupplier.ts
```diff
@@ -0,0 +1,299 @@
...
+   */
+  useDefaultCertificateConfig?: boolean;
+  /**
+   * The file path to the certificate configuration JSON file.
+   * Required if `useDefaultCertificateConfig` is not true.
+   */
+  certificateConfigLocation?: string;
+  /**
+   * The file path to the trust chain (PEM format). If not provided,
+   * only the leaf certificate will be sent.
```
nbayati, line 67: I think it's best if we remove the "If not provided, only the leaf certificate will be sent." part and just note that this field is optional. How we are creating the subject token to send to STS is an implementation detail that I think should not be exposed here, as it'd make the reader more confused than informed.

src/auth/certificatesubjecttokensupplier.ts
```diff
@@ -0,0 +1,299 @@
...
+  public createMtlsHttpsAgent(): https.Agent {
+    try {
+      const cert = fs.readFileSync(this.certPath);
+      const key = fs.readFileSync(this.keyPath);
+      return new https.Agent({key, cert});
+    } catch (err) {
+      const message = err instanceof Error ? err.message : String(err);
+      throw new CertificateSourceUnavailableError(
+        `Failed to read certificate or key file: ${message}`,
+      );
```
nbayati, line 140: Suggestion: it might be helpful to include the file path in the error message. We could use two different try/catch blocks for reading the key and the certificate files, so we know which file caused the error.

src/auth/identitypoolclient.ts
```diff
@@ -87,6 +90,28 @@ export interface IdentityPoolClientOptions
...
       subject_token_field_name?: string;
     };
+    /**
+     * The certificate location to call to retrieve the subject token. Either this or a file
+     * location should be specified.
+     */
+    certificate?: {
+      /**
+       * Specify whether the certificate config should be used from the default location
+       * either this or the certificate_config_location must be provided
```
nbayati, line 100: nit: Please capitalize the first word and end each sentence with a period.

src/auth/identitypoolclient.ts
```diff
@@ -87,6 +90,28 @@ export interface IdentityPoolClientOptions
...
+       */
+      use_default_certificate_config?: boolean;
+      /**
+       * Location to fetch certificate config from in case default config is not to be used.
+       * either this or use_default_certificate_config=true should be provided
```
nbayati, line 105: nit: Please capitalize the first word and end each sentence with a period.

src/auth/identitypoolclient.ts
```diff
@@ -87,6 +90,28 @@ export interface IdentityPoolClientOptions
...
+       */
+      certificate_config_location?: string;
+      /**
+       * Location to fetch trust chain from to send to STS endpoint.
+       * in case no location is provided, we will just send the leaf certificate as the
+       * trust chain
```
nbayati, line 111: This is not really accurate. If a trust chain file is not provided, we just send the leaf certificate as the "subject token" and not the trust chain. I'd suggest focusing on what this field IS rather than the details of how we'll use it in the implementation. For instance, in the [go library](https://github.com/googleapis/google-cloud-go/blob/93c44c9e530005ba4e2bc83091625ed4b8c8604a/auth/credentials/externalaccount/externalaccount.go#L189-L197) we have the following comments for these three fields:

```
type CertificateConfig struct {
	// UseDefaultCertificateConfig, if true, attempts to load the default
	// certificate configuration. It checks the GOOGLE_API_CERTIFICATE_CONFIG
	// environment variable first, then a conventional default file location.
	// Cannot be true if CertificateConfigLocation is set.
	UseDefaultCertificateConfig bool `json:"use_default_certificate_config"`
	// CertificateConfigLocation specifies the path to the client certificate
	// and private key file. Must be set if UseDefaultCertificateConfig is
	// false or unset.
	CertificateConfigLocation string `json:"certificate_config_location"`
	// TrustChainPath specifies the path to a PEM-formatted file containing
	// the X.509 certificate trust chain. This file should contain any
	// intermediate certificates required to complete the trust chain between
	// the leaf certificate (used for mTLS) and the root certificate(s) in
	// your workload identity pool's trust store. The leaf certificate and
	// any certificates already present in the workload identity pool's trust
	// store are optional in this file. Certificates should be ordered with the
	// leaf certificate (or the certificate which signed the leaf) first.
	TrustChainPath string `json:"trust_chain_path"`
}
```

src/auth/identitypoolclient.ts
```diff
@@ -183,9 +209,30 @@ export class IdentityPoolClient extends BaseExternalAccountClient {
...
+        this.stsCredential = new StsCredentials({
+          tokenExchangeEndpoint: this.getTokenUrl(),
+          clientAuthentication: this.clientAuth,
+          transporter: new Gaxios(stsClientGaxiosOptions),
+        });
+        this.transporter = new Gaxios(stsClientGaxiosOptions);
       } else {
         throw new Error(
-          'No valid Identity Pool "credential_source" provided, must be either file or url.',
+          'No valid Identity Pool "credential_source" provided, must be either file, or url, or certificate.',
```
nbayati, line 235: nit: 
```suggestion
          'No valid Identity Pool "credential_source" provided, must be either file, url, or certificate.',
```

test/test.identitypoolclient.ts
lsirac, line 1: Add tests:
* Malformed cert/key files: Tests for scenarios where the files pointed to by cert_path or key_path exist but are not valid PEM or key formats. 
* Malformed trust chain certs. While the order is tested, a test where one of the certificates within the trust chain PEM file is malformed should be added (this is related to the suggestion of more granular parsing I mentioned)
* Empty cert_config.json: Test the behavior when the cert_config.json file is empty or not valid JSON.
* Missing cert_path: Similar to the missing key_path test, add one for a missing cert_path in the config.

src/auth/certificatesubjecttokensupplier.ts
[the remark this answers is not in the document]
vverman: That's a good question,

The synchronous file operations are needed because reading the certificate and key files are needed for creating the IdentityPoolClient. For e.g. to create the mtlsAgent. Since the constructor cannot be async, we need to do it synchronously.

I have changed the instances of file reads in processChainFromPaths() to async since that is called from the async getSubjectToken().

src/auth/identitypoolclient.ts
```diff
@@ -87,6 +90,28 @@ export interface IdentityPoolClientOptions
        */
       subject_token_field_name?: string;
     };
+    /**
```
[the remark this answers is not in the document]
vverman: Done

src/util.ts
vverman: Done!

src/auth/baseexternalclient.ts
```diff
@@ -715,4 +716,8 @@ export abstract class BaseExternalAccountClient extends AuthClient {
       : 'unknown';
     return `gl-node/${nodeVersion} auth/${pkg.version} google-byoid-sdk source/${credentialSourceType} sa-impersonation/${saImpersonation} config-lifetime/${this.configLifetimeRequested}`;
   }
+
+  getTokenUrl(): string {
```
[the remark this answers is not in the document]
lsirac: I believe in TypeScript members of a class (properties and methods) are public by default if no access modifier is explicitly specified. To make a method private, you need to use the private keyword. So let's mark this as protected.

src/auth/baseexternalclient.ts
```diff
@@ -234,12 +234,13 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
   private readonly serviceAccountImpersonationLifetime?: number;
-  private readonly stsCredential: sts.StsCredentials;
-  private readonly clientAuth?: ClientAuthentication;
+  protected stsCredential: sts.StsCredentials;
+  protected readonly clientAuth?: ClientAuthentication;
   private readonly workforcePoolUserProject?: string;
   public projectNumber: string | null;
   private readonly configLifetimeRequested: boolean;
   protected credentialSourceType?: string;
+  private readonly tokenUrl: string;
```
lsirac, line 243: nit: maybe group these based on visibility

src/auth/certificatesubjecttokensupplier.ts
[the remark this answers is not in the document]
lsirac: Can't we technically defer the creation of the mTLS `https.Agent` and the specific `Gaxios` transporter instance until they are about to be used with the token exchange methods, allowing the file reads to be done asynchronously?

nbayati (approved):
LGTM, Good job! :)

lsirac (approved):

PR #2038, opened 2025-06-18
feat: Implement Centralized TrustBoundaries for Service Account, Impersonated and GCE Credentials

vverman:
This change includes trust boundaries for Service Account, Impersonated and GCE Credentials in the google auth nodejs library.

Impact
This will allow clients who use the google auth library to also enable trust boundaries ensuring workflows using the aforementioned credentials can access only those resources which are within the trust boundary. We use this [design doc](https://docs.google.com/document/d/1IkRoxRelBF-wUHktFLxKqfcQ2tqAwUg-XcYz_IrVHsQ/edit?resourcekey=0-KdUdpwuLu6eJtcdDb0lISw&tab=t.0#heading=h.17wg41voij6q) as reference for the code changes.

Testing
Added


src/auth/baseexternalclient.ts
```diff
@@ -637,11 +642,10 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
     // //iam.googleapis.com/projects/$PROJECT_NUMBER/locations/...
-    const match = audience.match(/\/projects\/([^/]+)/);
-    if (!match) {
-      return null;
-    }
-    return match[1];
+    return (
+      audience.match(/\/projects\/(?<projectNumber>[^/]+)\/locations\//)?.groups
+        ?.projectNumber ?? null
+    );
```
lsirac, line 648: Why was this changed?

src/auth/authclient.ts
```diff
@@ -289,6 +300,10 @@ export abstract class AuthClient
     res?: GaxiosResponse | null;
   }>;
 
+  getTrustBoundaryUrl(): string | null {
```
lsirac, line 303: Can you add documentation here

src/auth/baseexternalclient.ts
```diff
@@ -715,4 +719,69 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+  }
+
+  getTrustBoundaryUrl(): string {
+    let lookupTrustBoundaryUrl = null;
+    if (this.audience.match(WORKFORCE_AUDIENCE_PATTERN)) {
+      //client configured for workforce authorization
+      const wfPoolId = this.#getWorkForcePoolId(this.audience);
+      if (!wfPoolId) {
+        throw new RangeError(
+          'TrustBoundaryLookup: Failed to fetch trust boundary data due to missing workforce pool id',
```
lsirac, line 765: These errors should just tell the user that the audience they provided is invalid. Same below.

src/auth/baseexternalclient.ts
```diff
@@ -715,4 +719,69 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+        );
+      }
+      lookupTrustBoundaryUrl = WORKFORCE_LOOKUP_ENDPOINT.replace(
+        '{pool_id}',
+        encodeURIComponent(wfPoolId),
+      );
+    } else {
+      //client configured for workload authorization
```
lsirac, line 773:
```suggestion
```

src/auth/downscopedclient.ts
```diff
@@ -395,4 +397,14 @@ export class DownscopedClient extends AuthClient {
           downscopedAccessToken.expiry_date - this.eagerRefreshThresholdMillis
       : false;
   }
+
+  getTrustBoundaryUrl(): string {
+    const trustBoundaryUrl = this.authClient.getTrustBoundaryUrl();
+    if (!trustBoundaryUrl) {
+      throw new Error(
+        'TrustBoundary: Error getting tbUrl because of missing trustBoundaryUrl in calling client of DownScopedClient',
```
lsirac, line 405: Any cases where the auth client being used is not throwing an error?

Also, this error message should be improved if it is needed

src/auth/externalAccountAuthorizedUserClient.ts
```diff
@@ -222,7 +222,7 @@ export class ExternalAccountAuthorizedUserClient extends AuthClient {
     const headers = new Headers({
       authorization: `Bearer ${accessTokenResponse.token}`,
     });
-    return this.addSharedMetadataHeaders(headers);
+    return await this.addSharedMetadataHeaders(headers);
```
lsirac, line 225: This was missing..? strange

src/auth/impersonated.ts
```diff
@@ -253,4 +255,18 @@ export class Impersonated extends OAuth2Client implements IdTokenProvider {
 
     return res.data.token;
   }
+
+  getTrustBoundaryUrl(): string {
+    const targetPrincipal = this.getTargetPrincipal();
+    if (!targetPrincipal) {
+      throw new Error(
```
lsirac, line 262: Is this possible?

lsirac (commented):
Did a first pass:

1. Error messages e.g. from getTrustBoundaryUrl() need to be improved. Please double check that the failures in those cases are actually possible. 
2. Feature gating currently uses the wrong env var and checks the wrong value.
3. I believe you are using the wrong header value, should be x-allowed-locations.

Please add tests for the following:

* Verify that Trust Boundary lookups are correctly skipped when a non-default `universeDomain` is configured for `ComputeClient`, `Impersonated`, and `BaseExternalAccountClient`. 

* Test `DownscopedClient` wrapping an `authClient` that has a non-default `universeDomain` to confirm the lookup is skipped.
*   Test `DownscopedClient` wrapping an `authClient` where the lookup fails, but a valid (non-noop) trust boundary is already cached on the inner client.

* Add tests using malformed `audience` strings for both workload and workforce pools to ensure the regex parsing logic is robust and handles unexpected formats without crashing.

* Add a test where the initial Trust Boundary lookup fails, but a valid (non-noop) trust boundary was already present in the cache from a previous successful call. The test should verify that the cached data is returned and the refresh error is suppressed.

src/auth/baseexternalclient.ts
vverman: The first one checks for the literal string  `/projects/` followed by any characters that aren't a forward slash.

The second one checks for the literal string  `/projects/`, followed by non forward slash chars, followed by the `/locations/`

The getProjectNumber in the previous case would not return null in case of a malformed string like: 

> iam.googleapis.com/projects/locations/... 

or it would return a value in case of something like:

> iam.googleapis.com/projects/12345/serviceAccounts

which doesn't match the pattern. So my change is just a stricter check for the pattern.

src/auth/downscopedclient.ts
vverman: Yes. Consider the scenario where I use a downscoped client whose source is an Oauth2Client (which has been instantiated for individual users).

This source client is not covered for TB scenarios and its TBUrl would be falsy. Thus, in my understanding, we should throw here and allow for a descriptive error.

src/auth/externalAccountAuthorizedUserClient.ts
vverman: The `addSharedMetadataHeaders` has been converted into an async method to allow for calling the lookup endpoint at the authClient level.

`addSharedMetadataHeaders` is only ever used in async scenarios so there isn't any significant impact to this change. 

The upside is it makes our code significantly cleaner, easier to read and reduces duplication.

src/auth/impersonated.ts
vverman: Yes, in the constructor we see:

`this.targetPrincipal = options.targetPrincipal ?? '';`

Which means targetPrincipal can have an empty string when the options don't have a corresponding value. This means that our tbUrl would be improperly configured

In which case, I think we should throw an error.

src/auth/trustboundary.ts
```diff
@@ -0,0 +1,119 @@
...
+    throw new Error(
+      'TrustBoundary: Error calling lookup endpoint without valid access token',
+    );
+  }
+  const headers = new Headers({
+    //we can directly pass the access_token as the trust boundaries are always fetched after token refresh
+    authorization: 'Bearer ' + client.credentials.access_token,
+  });
+
+  const opts: GaxiosOptions = {
```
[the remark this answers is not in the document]
vverman: This retries 3 times.

> With retry: true and no other specific count provided, gaxios will attempt the request a total of 4 times: the initial request plus 3 retries by default.

Yes I will change it to only say GET.

src/auth/trustboundary.ts
```diff
@@ -0,0 +1,119 @@
...
+  };
+
+  try {
+    const {data: trustBoundaryData} =
+      // preferred to client.request to avoid unnecessary retries
+      await client.transporter.request<TrustBoundaryData>(opts);
+
+    // Check for the specific No-Op case and return null.
+    if (trustBoundaryData.encodedLocations === NoOpEncodedLocations) {
+      return null;
```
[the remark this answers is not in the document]
vverman: I am returning null and checking the value inside the authClient for NoOp to decide whether to include an empty string or not

src/auth/authclient.ts
```diff
@@ -239,6 +245,11 @@ export abstract class AuthClient
     this.quotaProjectId = options.get('quota_project_id');
     this.credentials = options.get('credentials') ?? {};
     this.universeDomain = options.get('universe_domain') ?? DEFAULT_UNIVERSE;
+    this.trustBoundaryEnabled =
+      (
+        process.env['GOOGLE_AUTH_ENABLE_TRUST_BOUNDARIES'] || ''
+      ).toLowerCase() === 'true';
```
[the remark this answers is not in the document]
vverman: Yes it was indeed, thanks for the catch, resolved

test/test.compute.ts
```diff
@@ -261,4 +266,88 @@ describe('compute', () => {
...
+        .matchHeader('authorization', mockAuthHeader)
+        .reply(200, expectedTrustBoundaryData);
+
+      const trustBoundary = await getTrustBoundary(compute);
+
+      assert.deepStrictEqual(trustBoundary, expectedTrustBoundaryData);
+      scope.done();
+    });
+
+    it('getTrustBoundary should return null when default domain is not googleapis.com', async () => {
```
vverman, line 313: Trust Boundary lookups are correctly skipped when a non-default universeDomain is configured

test/test.downscopedclient.ts
```diff
@@ -1252,4 +1253,98 @@ describe('DownscopedClient', () => {
...
+        tbClient,
+        testClientAccessBoundary,
+      );
+      await assert.rejects(
+        getTrustBoundary(downscopedClient),
+        /Error: TrustBoundary: GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED env variable set for invalid client type/,
+      );
+    });
+
+    it('should return null if source client does not have valid universe', async () => {
```
vverman, line 1318: Test DownscopedClient wrapping an authClient that has a non-default universeDomain to confirm the lookup is skipped.

test/test.impersonated.ts
```diff
@@ -568,4 +573,76 @@ describe('impersonated', () => {
...
+        .matchHeader('authorization', mockAuthHeader)
+        .reply(200, expectedTrustBoundaryData);
+
+      const trustBoundary = await getTrustBoundary(impersonated);
+
+      assert.deepStrictEqual(trustBoundary, expectedTrustBoundaryData);
+      scope.done();
+    });
+
+    it('getTrustBoundary should return null when default domain is not googleapis.com', async () => {
```
vverman, line 617: Trust Boundary lookups are correctly skipped when a non-default universeDomain is configured

src/auth/trustboundary.ts
```diff
@@ -0,0 +1,137 @@
...
+    tbEnabled === undefined ||
+    falsyValues.has(tbEnabled)
+  ) {
+    return false;
+  }
+  if (truthyValues.has(tbEnabled)) {
+    return true;
+  }
+  throw new Error(
+    `Invalid syntax for the GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED env variable: "${tbEnabled}" is not a valid boolean representation`,
```
nbayati, line 135: Maybe "invalid value"?

lsirac (commented):
I don't see a good way of centralizing this in AuthClient without lots of changes  👎 

Let's do the following:
1. Move and rename `getTrustBoundary()` to `protected async refreshTrustBoundary` method in `authclient`. 

2. Modify `refreshToken` classes in the key base classes (`OAuth2Client`, `BaseExternalAccountClient`, etc) to call the new `refreshTrustBoundary` method we defined on token refresh. 

3. Make `addSharedMetadataHeaders` synchronous again.

Please also update the description of the PR and the title to reflect what the changes cover.

src/auth/trustboundary.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (!trustBoundaryUrl) {
+    return null;
+  }
+
+  if (!client.credentials.access_token) {
+    throw new Error(
+      'TrustBoundary: Error calling lookup endpoint without valid access token',
+    );
+  }
+  const headers = new Headers({
```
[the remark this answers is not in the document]
vverman: Thanks for that, I added it in!

src/auth/trustboundary.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/**
+ * Fetches trust boundary data for an authenticated client.
+ * Handles caching checks and potential fallbacks.
+ * @param authenticatedClient An authenticated AuthClient instance to make the request.
+ * @returns A Promise resolving to TrustBoundaryData or empty-string for no-op trust boundaries.
+ * @throws {Error} If the request fails and there is no cache available.
+ */
+export async function getTrustBoundary(
```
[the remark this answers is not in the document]
vverman: Made the changes!

src/auth/authclient.ts
```diff
@@ -198,6 +203,9 @@ export declare interface AuthClient {
   on(event: 'tokens', listener: (tokens: Credentials) => void): this;
 }
 
+/**
+ * The base of all Auth Clients.
```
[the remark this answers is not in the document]
vverman: Done!

src/auth/authclient.ts
```diff
@@ -289,6 +302,18 @@ export abstract class AuthClient
...
 
+  /**
+   * Constructs the trust boundary lookup URL for the client.
+   *
+   * @return The trust boundary URL string, or `null` if the client type
+   * does not support trust boundaries.
+   * @throws {Error} If the URL cannot be constructed for a compatible client,
+   * for instance, if a required property like a service account email is missing.
+   */
+  async getTrustBoundaryUrl(): Promise<string | null> {
```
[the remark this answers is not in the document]
vverman: Done!

src/auth/authclient.ts
```diff
@@ -217,6 +225,9 @@ export abstract class AuthClient
   eagerRefreshThresholdMillis = DEFAULT_EAGER_REFRESH_THRESHOLD_MILLIS;
   forceRefreshOnFailure = false;
   universeDomain = DEFAULT_UNIVERSE;
+  trustBoundaryEnabled: boolean;
+  trustBoundary?: TrustBoundaryData | null;
+  trustBoundaryUrl?: string | null;
```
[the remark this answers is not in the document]
vverman: Done!

src/auth/computeclient.ts
```diff
@@ -137,4 +139,26 @@ export class Compute extends OAuth2Client {
       }
     }
   }
+
+  async getTrustBoundaryUrl(): Promise<string> {
+    let email = this.serviceAccountEmail;
+
+    if (email === 'default') {
```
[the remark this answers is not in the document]
vverman: Created a new function and tests are added! Thank you!

src/auth/authclient.ts
```diff
@@ -304,7 +329,10 @@ export abstract class AuthClient
    *
    * @param headers object to append additional headers to.
    */
-  protected addSharedMetadataHeaders(headers: Headers): Headers {
+  protected async addSharedMetadataHeaders(
```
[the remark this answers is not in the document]
vverman: The changes are made! Thanks for the suggestion, we're now doing trust boundary refresh while refreshing the access token!

src/auth/authclient.ts
```diff
@@ -523,6 +511,106 @@ export abstract class AuthClient
...
+
+    const accessToken = tokens.access_token;
+
+    if (!accessToken || this.isExpired(tokens)) {
+      throw new Error(
+        'TrustBoundary: Error calling lookup endpoint without valid access token',
+      );
+    }
+
+    const headers = new Headers({
```
nbayati, line 552: Can we use addSharedMetadataHeaders here?
This way the subsequent calls to the allowed location endpoint will have trust boundary header too.

src/auth/authclient.ts
```diff
@@ -523,6 +511,106 @@ export abstract class AuthClient
...
+          httpMethodsToRetry: ['GET'],
+        },
+      },
+      headers,
+      url: trustBoundaryUrl,
+    };
+
+    try {
+      const {data: trustBoundaryData} =
+        // preferred to client.request to avoid unnecessary retries
```
nbayati, line 578: This comment seems a bit ambiguous.

My understanding is that we do want network retries on transient errors, which is handled by the `retry: true` option passed to the transporter. However, we are using this.transporter.request directly to bypass the AuthClient.request method's specific logic for refreshing the token on a 401 error, which would be unnecessary since we've just refreshed it.

Could we clarify the comment to make this distinction clear for future maintainers? Perhaps something like:
// Using the transporter directly to avoid the token-refresh logic in client.request.

test/test.compute.ts
```diff
@@ -261,4 +266,302 @@ describe('compute', () => {
...
+
+    const MOCK_ACCESS_TOKEN = 'abc123';
+    const MOCK_AUTH_HEADER = `Bearer ${MOCK_ACCESS_TOKEN}`;
+    const SERVICE_ACCOUNT_EMAIL = 'service-account@example.com';
+    const EXPECTED_TB_DATA: TrustBoundaryData = {
+      locations: ['sadad', 'asdad'],
+      encodedLocations: '000x9',
+    };
+    const NO_OP_TB_DATA: TrustBoundaryData = {
+      locations: ['sadad', 'asdad'],
```
nbayati, line 281: No-op response either would have an empty `locations`, or it might not have `locations` field at all.

test/test.compute.ts
```diff
@@ -261,4 +266,302 @@ describe('compute', () => {
...
+        mockExample(),
+      ];
+
+      await compute.request({url});
+
+      assert.deepStrictEqual(compute.trustBoundary, EXPECTED_TB_DATA);
+      scopes.forEach(s => s.done());
+    });
+
+    it('fetchTrustBoundary should fail gcpMetadata call', async () => {
```
nbayati, line 384: nit:
```suggestion
    it('refreshTrustBoundary should throw when gcpMetadata call fails', async () => {
```

src/auth/authclient.ts
vverman: Negar: Why don't we use the client.request here?

Good point. You see the call to refresh trustBoundaries is made inside the refreshToken()

However the updated token for the client are set inside the calling method of refreshToken() which is getRequestMetadataAsync().

So if we make a call via client.request it will constantly keep refreshing the token.

danieljbruce:
I added a "Do Not Merge" label because I think we should wait for nbayati approval on this one.

test/test.compute.ts
vverman: Done!

test/test.compute.ts
```diff
@@ -261,4 +266,302 @@ describe('compute', () => {
...
+      });
+      const scopes = [setupTokenNock(SERVICE_ACCOUNT_EMAIL), mockExample()];
+
+      await compute.request({url});
+
+      assert.deepStrictEqual(compute.trustBoundary, null);
+      scopes.forEach(s => s.done());
+    });
+
+    it('fetchTrustBoundary should use the email from metadataServer if no serviceAccountEmail passed', async () => {
```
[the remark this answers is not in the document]
vverman: done!

test/test.compute.ts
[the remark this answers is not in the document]
vverman: Sorry forgot to respond to this. The common test cases for AuthClient are now within the test.authclient.ts.

nbayati (approved):
Looks good! Just left one small comment about the method name.

src/auth/authclient.ts
```diff
@@ -357,7 +357,10 @@ export abstract class AuthClient
    * @param source the headers to source from
    * @returns the target headers
    */
-  protected addCommonHeaders<T extends Headers>(target: T, source: Headers): T {
+  protected copyCommonHeaders<T extends Headers>(
```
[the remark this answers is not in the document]
vverman: "copyCommonHeaders" could be interpreted as copying the headers that are common between the source and target objects, rather than copying a specific set of predefined headers."

I agree, though I doubt that the x-goog-user-project is for auth, I believe it is for billing.

How about applyHeadersFromSource?

Negar approves!

lsirac (approved):

sofisl (approved):
It would be awesome if there was also a sample and/or README addition, but code-wise LGTM 👍

PR #2050, opened 2025-07-08
chore(deps): upgrade sinon to 21

### LINKED ISSUE #2049: Upgrade sinon to 21
Need to reconfigure tests for the fake timers: https://github.com/googleapis/google-auth-library-nodejs/pull/2037

leahecole:
Fixes #2049 
Major credit to @feywind for already making a util for nodejs-pubsub that I could lift and use here!

### DIFF
--- a/.github/scripts/package.json
+++ b/.github/scripts/package.json
@@ -16,6 +16,6 @@
     "devDependencies": {
         "@octokit/rest": "^19.0.0",
         "mocha": "^10.0.0",
-        "sinon": "^18.0.0"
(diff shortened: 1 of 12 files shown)


feywind (dismissed):
Two overall comments:

- In pubsub, I pulled this out into a utility function so that it's consistent across the tests, but ymmv
- You might want to pull out setImmediate and clearImmediate too, since those are more or less just aliases for process.nextTick

feywind (approved):

PR #2106, opened 2025-08-01
fix: pin nock and typescript dependencies to fix window.crypto.subtle.verify signature argument type

miguelvelezsa:
Due to a new TS  version, we are getting an error in crypto file [1].
Also test are timing out due to a new version in nock.

Set specific versions for both dependencies as short term solution, until we implement the real fix.

[1] src/crypto/browser/crypto.ts:94:7 - error TS2345: Argument of type 'Uint8Array<ArrayBufferLike>' is not assignable to parameter of type 'BufferSource'.
  Type 'Uint8Array<ArrayBufferLike>' is not assignable to type 'ArrayBufferView<ArrayBuffer>'.
    Types of property 'buffer' are incompatible.
      Type 'ArrayBufferLike' is not assignable to type 'ArrayBuffer'.
        Type 'SharedArrayBuffer' is not assignable to type 'ArrayBuffer'.
          Types of property '[Symbol.toStringTag]' are incompatible.
            Type '"SharedArrayBuffer"' is not assignable to type '"ArrayBuffer"'.

94       signatureArray,
         ~~~~~~~~~~~~~~

src/crypto/node/crypto.ts:90:3 - error TS2322: Type 'ArrayBuffer | SharedArrayBuffer' is not assignable to type 'ArrayBuffer'.
  Type 'SharedArrayBuffer' is not assignable to type 'ArrayBuffer'.
    Types of property '[Symbol.toStringTag]' are incompatible.
      Type '"SharedArrayBuffer"' is not assignable to type '"ArrayBuffer"'.

90   return buffer.buffer.slice(


sofisl (approved):

feywind:
Linking for reference: https://github.com/microsoft/TypeScript/issues/14116

We talked about it on chat, but going forward it looks like we can lock to minver safely (`~`).

PR #2118, opened 2025-08-18
feat: TrustBoundaries for external accounts.

vverman:
This change includes trust boundaries for External Accounts (workforce and workload authorization) in the google auth nodejs library. 

Impact
This will allow clients who use the google auth library to also enable trust boundaries ensuring workload and workforce federation authorization flows can access only those resources which are within the trust boundary. We use this [design doc](https://docs.google.com/document/d/1IkRoxRelBF-wUHktFLxKqfcQ2tqAwUg-XcYz_IrVHsQ/edit?resourcekey=0-KdUdpwuLu6eJtcdDb0lISw&tab=t.0#heading=h.17wg41voij6q) as reference for the code changes.

Testing
All features are unit tested as per [this](https://docs.google.com/document/d/1TXbIuzsmulN6XaYoC7fChVxWHiv0D4LHguKxmfm2v6Y/edit?tab=t.cvojbvfjbyb) testing document.


src/auth/baseexternalclient.ts
```diff
@@ -708,4 +714,77 @@ export abstract class BaseExternalAccountClient extends AuthClient {
   protected getTokenUrl(): string {
     return this.tokenUrl;
   }
+
+  /**
+   * Returns the workforce identity pool-id if it is determinable
```
nbayati, line 719: nit: I think using "pool id" in the comments instead of "pool-id" would improve readability. It also makes searching the codebase easier, since we usually search for the term without a hyphen.

src/auth/baseexternalclient.ts
```diff
@@ -708,4 +714,77 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+   * @param audience The STS audience used to determine the pool-id.
+   * @return The pool-id associated with the workforce identity pool, if
+   *   this can be determined from the STS audience field. Otherwise, null is
+   *   returned.
+   */
+  #getWorkForcePoolId(audience: string): string | null {
+    // STS audience pattern:
+    // .../workforcePools/$WORKFORCE_POOL_ID/providers/...
+    return (
+      audience.match(/\/workforcePools\/(?<workforcePool>[^/]+)\/providers\//)
```
nbayati, line 730: To make the client more robust, consider adding a check for the audience field in the constructor. This ensures we "fail fast" and propagate a clear error message if a developer tries to instantiate an external client without a required audience. It would prevent a more cryptic TypeError from happening later when audience.match() is called.

src/auth/baseexternalclient.ts
```diff
@@ -708,4 +714,77 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+    const projectNumber = this.getProjectNumber(this.audience);
+    if (wlPoolId && projectNumber) {
+      return WORKLOAD_LOOKUP_ENDPOINT.replace(
+        '{project_id}',
+        projectNumber,
+      ).replace('{pool_id}', wlPoolId);
+    }
+
+    throw new RangeError(
+      `TrustBoundary: Invalid audience provided: "${this.audience}" does not correspond to workforce or workload`,
```
nbayati, line 787: nit:
```suggestion
    throw new RangeError(
      `TrustBoundary: Invalid audience provided: "${this.audience}" does not correspond to a workforce or workload pool.`,
```

nbayati (changes requested):
Consider adding unit tests for the following failure-mode scenarios:

   * The retry mechanism for the trust boundary lookup (e.g., the endpoint fails once
     with a 503, then succeeds).
   * The fallback mechanism where a failed trust boundary lookup correctly uses a
     previously cached value from a prior successful call.
   * Handling of a malformed response from the trust boundary endpoint (e.g., the
     endpoint returns a 200 OK status but with an invalid or empty JSON payload).

sofisl (approved):

nbayati (approved):
LGTM

src/auth/authclient.ts
```diff
@@ -668,6 +663,21 @@ export abstract class AuthClient
     }
   }
 
+  /**
+   * Gets the trust boundary header to be attached to the request.
+   */
+  protected getTrustBoundaryHeader(): string | null {
```
lsirac, line 669: Why is this being added as part of this PR (and not the earlier PRs)?

src/auth/baseexternalclient.ts
```diff
@@ -708,4 +714,76 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+    return (
+      audience.match(/\/workforcePools\/(?<workforcePool>[^/]+)\/providers\//)
+        ?.groups?.workforcePool ?? null
+    );
+  }
+
+  /**
+   * Returns the workload identity pool id if it is determinable
+   * from the audience resource name.
+   * @param audience The STS audience used to determine the pool id.
```
lsirac, line 738: I don't think we need to refer to this as "STS audience", just "audience"

src/auth/baseexternalclient.ts
```diff
@@ -708,4 +714,76 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+    // .../workloadIdentityPools/POOL_ID/providers/...
+    return (
+      audience.match(
+        /\/workloadIdentityPools\/(?<workloadPool>[^/]+)\/providers\//,
+      )?.groups?.workloadPool ?? null
+    );
+  }
+
+  protected async getTrustBoundaryUrl(): Promise<string> {
+    if (this.serviceAccountImpersonationUrl) {
```
lsirac, line 754: We should add a comment here explaining why we check this first

src/auth/baseexternalclient.ts
```diff
@@ -708,4 +714,76 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+    //check for workforce
+    const wfPoolId = this.#getWorkForcePoolId(this.audience);
+    if (wfPoolId) {
+      return WORKFORCE_LOOKUP_ENDPOINT.replace(
+        '{universe_domain}',
+        this.universeDomain,
+      ).replace('{pool_id}', encodeURIComponent(wfPoolId));
+    }
+
+    //check for workload
```
lsirac, line 776: Same here:

```suggestion
    // Check if the audience corresponds to a workload identity pool.
```

src/auth/baseexternalclient.ts
```diff
@@ -708,4 +714,76 @@ export abstract class BaseExternalAccountClient extends AuthClient {
...
+          `TrustBoundary: A service account email is required for trust boundary lookups but could not be determined from the serviceAccountImpersonationUrl ${this.serviceAccountImpersonationUrl}.`,
+        );
+      }
+      return SERVICE_ACCOUNT_LOOKUP_ENDPOINT.replace(
+        '{universe_domain}',
+        this.universeDomain,
+      ).replace('{service_account_email}', encodeURIComponent(email));
+    }
+
+    //check for workforce
```
lsirac, line 767: Lets keep comments consistent, e.g.:
```suggestion
    // Check if the audience corresponds to a workforce identity pool.
```

src/auth/computeclient.ts
```diff
@@ -144,9 +144,9 @@ export class Compute extends OAuth2Client {
   protected async getTrustBoundaryUrl(): Promise<string> {
     const email = await this.resolveServiceAccountEmail();
     const trustBoundaryUrl = SERVICE_ACCOUNT_LOOKUP_ENDPOINT.replace(
-      '{service_account_email}',
-      encodeURIComponent(email),
-    );
+      '{universe_domain}',
```
lsirac, line 147: I am leaving this comment here just to be sure that we are skipping trust boundary lookup if the universe is not googleapis.com. Is this the case?

src/auth/externalAccountAuthorizedUserClient.ts
```diff
@@ -309,6 +312,46 @@ export class ExternalAccountAuthorizedUserClient extends AuthClient {
       this.refreshToken = refreshResponse.refresh_token;
     }
 
+    // Set credentials and refresh trust boundary data.
+    this.credentials = {...this.cachedAccessToken};
+    delete (this.credentials as CredentialsWithResponse).res;
+    this.trustBoundary = await this.refreshTrustBoundary(this.credentials);
```
lsirac, line 318: Is the formatting off for this?

src/auth/externalAccountAuthorizedUserClient.ts
```diff
@@ -309,6 +312,46 @@ export class ExternalAccountAuthorizedUserClient extends AuthClient {
...
+
+  /**
+   * Retrieves the workforce pool ID from the audience.
+   *
+   * The audience has the format:
+   * `//iam.googleapis.com/locations/global/workforcePools/{pool_id}/providers/{provider_id}`
+   *
+   * @returns The workforce pool ID, or `null` if it cannot be determined.
+   */
+  getWorkforcePoolId(): string | null {
```
lsirac, line 331: Looks like this is duplicated, can we create a shared utility function so both BaseExternalAccountClient and ExternalAccountAuthorizedUserClient could use it?

src/auth/authclient.ts
lsirac, line 669: This could be simplified:

```
if (!this.trustBoundaryEnabled || !this.trustBoundary) {
  return null;
}
// Empty header sent in case trust-boundary has no-op encoded location.
return this.trustBoundary.encodedLocations === NoOpEncodedLocations
  ? ''
  : this.trustBoundary.encodedLocations;

File: src/util.ts [hunk 12]
```diff
@@ -300,3 +300,22 @@ export function getWellKnownCertificateConfigFileLocation(): string {
 function _isWindows(): boolean {
   return os.platform().startsWith('win');
 }
+
+/**
+ * Returns the workforce identity pool id if it is determinable
+ * from the audience resource name.
+ * @param audience The STS audience used to determine the pool id.
```
lsirac, line 307:
```suggestion
 * @param audience The audience used to determine the pool id.
```

src/util.ts
```diff
@@ -300,3 +300,22 @@ export function getWellKnownCertificateConfigFileLocation(): string {
 function _isWindows(): boolean {
   return os.platform().startsWith('win');
 }
+
+/**
+ * Returns the workforce identity pool id if it is determinable
```
lsirac, line 305: ID or identifier here and everywhere else

src/util.ts
```diff
@@ -300,3 +300,22 @@ export function getWellKnownCertificateConfigFileLocation(): string {
...
+ * from the audience resource name.
+ * @param audience The STS audience used to determine the pool id.
+ * @return The pool id associated with the workforce identity pool, if
+ *   this can be determined from the STS audience field. Otherwise, null is
+ *   returned.
+ */
+export function getWorkforcePoolIdFromAudience(
```
lsirac, line 312: to keep things consistent, lets also define the workload method in this file

src/auth/authclient.ts
```diff
@@ -407,9 +407,14 @@ export abstract class AuthClient
       headers.set('x-goog-user-project', this.quotaProjectId);
     }
 
-    const trustBoundaryHeader = this.getTrustBoundaryHeader();
-    if (trustBoundaryHeader !== null) {
-      headers.set('x-allowed-locations', trustBoundaryHeader);
+    if (this.trustBoundaryEnabled && this.trustBoundary) {
```
lsirac, line 410: I noticed you changed this to include `this.trustBoundaryEnabled` but I don't see an added unit test. Is this covered by a unit test?

src/auth/authclient.ts
```diff
@@ -407,9 +407,14 @@ export abstract class AuthClient
...
+      //Empty header sent in case trust-boundary has no-op encoded location.
```
lsirac, line 411:
```suggestion
      // Empty header sent in case trust-boundary has no-op encoded location.
```

lsirac (approved):

Issue #2139, opened 2025-09-30
Workload Identity Federation GCP <-> AWS failing


benkoska:
### Please make sure you have searched for information in the following guides.

- [x] Search the issues already opened: https://github.com/GoogleCloudPlatform/google-cloud-node/issues
- [x] Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js
- [x] Check our Troubleshooting guide: https://github.com/googleapis/google-cloud-node/blob/main/docs/troubleshooting.md
- [x] Check our FAQ: https://github.com/googleapis/google-cloud-node/blob/main/docs/faq.md
- [x] Check our libraries HOW-TO: https://github.com/googleapis/gax-nodejs/blob/main/client-libraries.md
- [x] Check out our authentication guide: https://github.com/googleapis/google-auth-library-nodejs
- [x] Check out handwritten samples for many of our APIs: https://github.com/GoogleCloudPlatform/nodejs-docs-samples

### A screenshot that you have tested with "Try this API".

The intended behavior does not directly use a Google API, but is a combination of Google Workload Federation and the AWS counterpart that is used by this package to support Workload Identity Federation.

If you want to test the API call manually, you can run the following commands on the configured EC2 instance (after following the tutorial linked in the gist)
```bash
TOKEN=$(curl -sS -m 2 -X PUT "http://169.254.169.254/latest/api/token" \
  -H "X-aws-ec2-metadata-token-ttl-seconds: 21600") || echo "token-failed:$?"
ROLE=$(curl -sS -m 2 -H "X-aws-ec2-metadata-token: $TOKEN" \
  "http://169.254.169.254/latest/meta-data/iam/security-credentials/") || echo "role-failed:$?"
curl --verbose -sS -m 2 -H "X-aws-ec2-metadata-token: $TOKEN"   "http://169.254.169.254/latest/meta-data/iam/security-credentials/$ROLE"
```

which will correctly return a token, accessKeyID and secretAccessKey as so
```json
> GET /latest/meta-data/iam/security-credentials/<ROLE> HTTP/1.1
> Host: 169.254.169.254
> User-Agent: curl/8.11.1
> Accept: */*
> X-aws-ec2-metadata-token: [....]
> 
* Request completely sent off
< HTTP/1.1 200 OK
< X-Aws-Ec2-Metadata-Token-Ttl-Seconds: 21391
< Content-Type: text/plain
< Accept-Ranges: none
[...]

PR #2140, opened 2025-09-30
fix: #2139

### LINKED ISSUE #500: Workload Identity Federation GCP <-> AWS failing
### Please make sure you have searched for information in the following guides.

- [x] Search the issues already opened: https://github.com/GoogleCloudPlatform/google-cloud-node/issues
- [x] Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js
- [x] Check our Troubleshooting guide: https://github.com/googleapis/google-cloud-node/blob/main/docs/troubleshooting.md
- [x] Check our FAQ: https://github.com/googleapis/google-cloud-node/blob/main/docs/faq.md
- [x] Check our libraries HOW-TO: https://github.com/googleapis/gax-nodejs/blob/main/client-libraries.md
- [x] Check out our authentication guide: https://github.com/googleapis/google-auth-library-nodejs
- [x] Check out handwritten samples for many of our APIs: https://github.com/GoogleCloudPlatform/nodejs-docs-samples

### A screenshot that you have tested with "Try this API".

The intended behavior does not directly use a Google API, but is a combination of Google Workload Federation and the AWS counterpart that is used by this package to support Workload Identity Federation.

If you want to test the API call manually, you can run the following commands on the configured EC2 instance (after following the tutorial linked in the gist)
```bash
TOKEN=$(curl -sS -m 2 -X PUT "http://169.254.169.254/latest/api/token" \
  -H "X-aws-ec2-metadata-token-ttl-seconds: 21600") || echo "token-failed:$?"
ROLE=$(curl -sS -m 2 -H "X-aws-ec2-metadata-token: $TOKEN" \
  "http://169.254.169.254/latest/meta-data/iam/security-credentials/") || echo "role-failed:$?"
curl --verbose -sS -m 2 -H "X-aws-ec2-metadata-token: $TOKEN"   "http://169.254.169.254/latest/meta-data/iam/security-credentials/$ROLE"
```

which will correctly return a token, accessKeyID and secretAccessKey as so
```json
> GET /latest/meta-data/iam/security-credentials/<ROLE> HTTP/1.1
> Host: 169.254.169.254
> User-Agent: curl/8.11.1
> Accept: */*
> X-aws-ec2-metadata-token: [....]
> 
* Request completely sent off
< HTTP/1.1 200 OK
< X-Aws-Ec2-Metadata-Token-Ttl-Seconds: 21391
< Content-Type: text/plain
< Accept-Ranges: none
< Last-Modified: Tue, 30 Sep 2025 00:02:53 GMT
< Content-Length: [...]
< Date: Tue, 30 Sep 2025 00:14:07 GMT
< Server: EC2ws
< Connection: close
< 
{
  "Code" : "Success",
  "LastUpdated" : "2025-09-30T00:02:01Z",
  "Type" : "AWS-HMAC",
  "AccessKeyId" : "[....]",
  "SecretAccessKey" : "[...]",
  "Token" : "[...]",
  "Expiration" : "2025-09-30T06:16:41Z"
}
```

Notably though, one notices that the response from the AWS server includes a `Content-Type: text/plain` header.

### Link to the code that reproduces this issue. A link to a **public** Github Repository or gist with a minimal reproduction.

https://gist.github.com/benkoska/ec03dae2f37eb7ae051215c188fbf4d8

### A step-by-step description of how to reproduce the issue, based on the linked reproduction.

1. Follow the linked tutorial to configure Workload Identity Federation
2. Run the given script from an AWS EC2 instance that is configured for Workload Identity Federation
3. The script will fail and throw an error

### A clear and concise description of what the bug is, and what you expected to happen.

The process fails to authenticate and receive the token necessary to run operations in the project (in the example, list all instances in a given zone).

If the same credentials file is used with the `gcloud` CLI, you are able to authenticate and utilize Workload Identity Federation as intended by executing commands in the project.

### A clear and concise description WHY you expect this behavior, i.e., was it a recent change, there is documentation that points to this behavior, etc. **

It is the intended behavior as described in the linked tutorial on Workload Identity Federation and if you run the equivalent `gcloud` CLI commands, it works as intended.

benkoska:
## Description
Given that the AWS API (called https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/defaultawssecuritycredentialssupplier.ts#L250) returns a JSON body with the correct tokens, but with a Content-Type of `text/plain` instead of `application/json` and given that the gaxios call no longer specifies a responseType (https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/defaultawssecuritycredentialssupplier.ts#L245), gaxios tries to infer the return type from the header (`text/plain`), therefore simply returning a string instead of a object.

The function calling this method to get the credential, assumes that the response is an object though, accessing fields such as `awsCreds.AccessKeyId` (https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/defaultawssecuritycredentialssupplier.ts#L178), which evaluate to undefined, as the properties do not exist on a String.

The rest of the code seems to assume that if an object is returned and no error is thrown, that the fields are correctly set and do not further validate anything. This leads to further API calls that rely on these values to simply use them, without validating if they are set, resulting in downstream errors.

This patch fixes this, by setting the `responseType` back to `json` (which it used to be set to) meaning that gaxios will try to parse it as a json object, even if the Content-Type header is `text/plain` and adding an explicit validation that the corresponding fields exist, throwing an error if they do not.

## Impact
If the response from the AWS Security Credentials API is invalid, it will throw an error.

## Testing
No tests were added or changed. A test would need to be run on an AWS EC2 instance that is configured for Workload Identity Federation with a Google Cloud Project to run and cannot be run using the current setup (to my knowledge).

## Additional Information
None

## Checklist

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-auth-library-nodejs/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Code coverage does not decrease
- [x] Appropriate docs were updated
- [x] Appropriate comments were added, particularly in complex areas or places that require background
- [x] No new warnings or issues will be generated from this change

Fixes googleapis/google-cloud-node-core#500 🦕

### DIFF
--- a/src/auth/defaultawssecuritycredentialssupplier.ts
+++ b/src/auth/defaultawssecuritycredentialssupplier.ts
@@ -175,6 +175,9 @@ export class DefaultAwsSecurityCredentialsSupplier
       metadataHeaders,
       context.transporter,
     );
+    if (awsCreds == null || awsCreds.AccessKeyId == null || awsCreds.SecretAccessKey == null || awsCreds.Token == null) {
+      throw new Error("Retrieved invalid aws credentials, expected fields `AccessKeyId`, `SecretAccessKey` and `Token` to be non-null but one or more are null.");
+    }
     return {
       accessKeyId: awsCreds.AccessKeyId,
       secretAccessKey: awsCreds.SecretAccessKey,
@@ -243,6 +246,9 @@ export class DefaultAwsSecurityCredentialsSupplier
       ...this.additionalGaxiosOptions,
       url: `${this.securityCredentialsUrl}/${roleName}`,
       headers: headers,
+      // NOTE: Do not remove this. AWS returns `Content-Type: text/plain` even though
+      // the response is a json, so it is necessary to have this.
+      responseType: 'json'
     } as GaxiosOptions;
     AuthClient.setMethodName(opts, '#retrieveAwsSecurityCredentials');
     const response =


leahecole (changes requested):
@benkoska apologies for the delay, we have been triaging issues while on rotation and I've finally made it to this one in the backlog. Could you please add a test for this?

sofisl:
I'm sorry @benkoska, would you mind potentially reopening this in googleapis/google-cloud-node-core? We migrated the library over there, which incidentally will have better integration tests and will be easier to merge features :) thank you!

PR #2174, opened 2025-11-16
feat: Agentic Identities in Cloudrun

vverman:
This feature implements the ability for agentic identities to authenticate themselves via X509 cert bound tokens. We are limiting the scope here to only cloud run based agentic workloads.


sofisl:
@vverman could you open this in google-cloud-node-core, where this library has moved to? cc @miguelvelezsa @GautamSharda we should deprecate/archive this repo so as to not confuse people.

sofisl:
I'm sorry @vverman , would you mind potentially reopening this in googleapis/google-cloud-node-core? We migrated the library over there, which incidentally will have better integration tests and will be easier to merge features :) thank you!

